PR #13, opened 2020-02-21
PoS extension initial commit for review

talvasconcelos:
Initial commit for @arcbtc and others review. Its lacking the invoice creation and checking. Also it need to get some data at start, like the rate and the fiat pair (USD, EUR, etc...). I think it would be best if they are set at creation time, to minimize end user messing around!


arcbtc:
Closing, as this badboy is live!

PR #19, opened 2020-04-25
Allow lightning scheme

### LINKED ISSUE #17: Accept payment_requests starting lightning:ln...
Often invoices are generate with a kind of "scheme" prefix: `lightning:` and the the payment requests follows. Currently LNbits gives this error: `Malformed request: checksum is incorrect.
400 BAD REQUEST`.

frennkie:
This adds the possibility to paste or scan an invoice (on "SEND") even if the payment request is prefixed by "lightning:" or "bitcoin:" (as per recommendation in https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md) which some wallets (e.g. eclair) do.

fix #17

I would suggest to squash on merge (or should I do it manually?)

### DIFF
--- a/lnbits/core/static/js/wallet.js
+++ b/lnbits/core/static/js/wallet.js
@@ -248,8 +248,13 @@ new Vue({
       this.sendCamera.show = false;
     },
     decodeInvoice: function () {
+      if (this.send.data.bolt11.startsWith('lightning:')) {
+        this.send.data.bolt11 = this.send.data.bolt11.slice(10);
+      }
+
+      let invoice;
       try {
-        var invoice = decode(this.send.data.bolt11);
+        invoice = decode(this.send.data.bolt11);
       } catch (error) {
         this.$q.notify({
           timeout: 3000,
@@ -261,7 +266,7 @@ new Vue({
         return;
       }
 
-      var cleanInvoice = {
+      let cleanInvoice = {
         msat: invoice.human_readable_part.amount,
         sat: invoice.human_readable_part.amount / 1000,
         fsat: LNbits.utils.formatSat(invoice.human_readable_part.amount / 1000)


frennkie:
Sorry for the unrelated changes; but my IDE (PyCharm) fixed a few of the wrong indents and spaces...

sgeisler:
Hi, the added functionality looks good but is obscured by the reformatting changes and the structure of the PR. I think review would be much easier if you could either change this PR or open a new one so that it avoids:

* commits that cancel each other out (don't fix errors from previous commits in newer ones in you PR, use interactive rebases instead)
* reformatting commits (you can probably deactivate that behavior in your IDE)
* merging master into your branch, rebase on master instead if it changed considerably since you forked

These improvements will make reviewing your PR much easier :)

(Please ignore the broken CI for now, it's not due to any of your changes since they only touched the JS part.)

frennkie:
Updated. Beside the additional formatting changes the points you listed should have been resolved by the squashing. But I agree that this is cleaner.

lnbits/core/static/js/wallet.js
```diff
@@ -248,8 +248,15 @@ new Vue({
       this.sendCamera.show = false;
     },
     decodeInvoice: function () {
+      if (this.send.data.bolt11.startsWith('lightning:')) {
+        this.send.data.bolt11 = this.send.data.bolt11.slice(10);
+      } else if (this.send.data.bolt11.startsWith('bitcoin:')) {
+        this.send.data.bolt11 = this.send.data.bolt11.slice(8);
```
sgeisler, line 254: According to [bip121](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) `bitcoin:<ln-invoice>` isn't a valid bitcoin URI. I couldn't find a formal definition of the bip21 LN extension, but https://github.com/LN-Zap/zap-desktop/pull/3131 looks like a good example how it's handled in practice.

frennkie: Correct me if I'm wrong... `LNbits` would not send to a Bitcoin address here, right?

In that case I would limit this PR to only add `'lightning:'` - agreed?

sgeisler: It wouldn't send on-chain tx, but there is a fallback scheme for bip21 URIs that looks like this: `bitcoin:bc1…?lightning=lnbc1…`. It's described in LN-Zap/zap-desktop#3131.

frennkie:
Let's focus this PR on the `lightning:` change.

For the fallback it appears that the BOLT is missing the exact details https://github.com/lightningnetwork/lightning-rfc/issues/323 (although as you linked there seems to be a general agreement).

frennkie:
@sgeisler I don't really get what zap-desktop is doing in their code (I admit that my JS skills are limited). I would parse this e.g. in this way:

```
      if (this.send.data.bolt11.startsWith('lightning:')) {
        // this invoice is prefixed - simply remove the 'lighting:' part
        this.send.data.bolt11 = this.send.data.bolt11.slice(10);
      } else if (this.send.data.bolt11.startsWith('bitcoin:')) {
        // a prefix of 'bitcoin:' indicates bip21 encoding with a bitcoin address 
        // for fallback and the bolt11 added as the value for the 'lightning' parameter
        let split_bolt = this.send.data.bolt11.split('?');
        if (split_bolt.length === 2 && split_bolt[1].startsWith('lightning=')) {
          this.send.data.bolt11 = split_bolt[1].slice(10);
        }
      }
```

What do you think?

sgeisler:
bip21 allows multiple parameters to be [chained using `&`](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki#abnf-grammar). So naively assuming everything behind the lightning key to be the bolt11 value doesn't work in general. Given the complexity I'd advise against just splitting the string a dozen times, instead writing a regex with a capture group seems to be the right approach.

frennkie:
Ok.. definitely too naive.. ;-) 

Will you merge the basic lightning PR?

sgeisler:
I'm not sure why the CI fails for your PR but not master. You haven't changed anything that's tested. Could you please rebase on master to test if CI works afterwards?

frennkie:
Hm.. maybe the `env` is different on other `branches` than on `master`..?!

It seems to fail after: `getenv("LNTXBOT_API_ENDPOINT")`

sgeisler:
Yeah, found this too, it seems they use some secret API endpoint for integration testing. Secret env vars [aren't available for PR builds](https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions). I informed people with the rights to fix it, no more I can do here :(

PR #22, opened 2020-04-26
Fix mypy errors

sgeisler:
This build on @eillarra's earlier commit to fix or ignore all type errors so that the CI will not fail on every PR anymore but only if it actually introduces potential bugs.

### DIFF
--- a/lnbits/__init__.py
+++ b/lnbits/__init__.py
@@ -1,9 +1,9 @@
 import importlib
 
 from flask import Flask
-from flask_assets import Environment, Bundle
-from flask_compress import Compress
-from flask_talisman import Talisman
+from flask_assets import Environment, Bundle  # type: ignore
+from flask_compress import Compress  # type: ignore
+from flask_talisman import Talisman  # type: ignore
 from os import getenv
 from werkzeug.middleware.proxy_fix import ProxyFix
 
@@ -15,7 +15,7 @@ from .settings import FORCE_HTTPS
 disabled_extensions = getenv("LNBITS_DISABLED_EXTENSIONS", "").split(",")
 
 app = Flask(__name__)
-app.wsgi_app = ProxyFix(app.wsgi_app, x_proto=1, x_host=1)
+app.wsgi_app = ProxyFix(app.wsgi_app, x_proto=1, x_host=1) # type: ignore
 valid_extensions = [ext for ext in ExtensionManager(disabled=disabled_extensions).extensions if ext.is_valid]
 
 
--- a/lnbits/bolt11.py
+++ b/lnbits/bolt11.py
@@ -1,3 +1,5 @@
+# type: ignore
+
 import bitstring
 import re
 from binascii import hexlify
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -16,7 +16,10 @@ def create_account() -> User:
         user_id = uuid4().hex
         db.execute("INSERT INTO accounts (id) VALUES (?)", (user_id,))
 
-    return get_account(user_id=user_id)
+    new_account = get_account(user_id=user_id)
+    assert new_account, "Newly created account couldn't be retrieved"
+
+    return new_account
 
 
 def get_account(user_id: str) -> Optional[User]:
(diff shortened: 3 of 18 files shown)


lnbits/__init__.py
```diff
@@ -15,7 +15,7 @@
 disabled_extensions = getenv("LNBITS_DISABLED_EXTENSIONS", "").split(",")
 
 app = Flask(__name__)
-app.wsgi_app = ProxyFix(app.wsgi_app, x_proto=1, x_host=1)
+app.wsgi_app = ProxyFix(app.wsgi_app, x_proto=1, x_host=1) # type: ignore
```
sgeisler, line 18: Replacing methods like this seems hacky and unsupported by mypy but apparently works

lnbits/bolt11.py
```diff
@@ -1,3 +1,5 @@
+# type: ignore
```
sgeisler, line 1: Ignoring `bolt11` since it will be replaced soonish anyway

lnbits/core/crud.py
```diff
@@ -16,7 +16,10 @@ def create_account() -> User:
         user_id = uuid4().hex
         db.execute("INSERT INTO accounts (id) VALUES (?)", (user_id,))
 
-    return get_account(user_id=user_id)
+    new_account = get_account(user_id=user_id)
+    assert new_account, "Newly created account couldn't be retrieved"
```
sgeisler, line 20: There are a few places where I used this pattern, I'm not sure if a function `expect(x: Option[T], assert_msg: str) -> T` would be a good abstraction in the python world (I'd totally go for it as a rust programmer :smile:).

lnbits/core/models.py
```diff
@@ -59,7 +59,7 @@ def msat(self) -> int:
 
     @property
     def sat(self) -> int:
-        return self.amount / 1000
+        return int(self.amount / 1000)
```
sgeisler, line 62: @eillarra, maybe this should rather be an integer division `//`? 
```suggestion
        return self.amount // 1000
```

lnbits/core/services.py
```diff
@@ -23,8 +23,10 @@ def create_invoice(*, wallet_id: str, amount: int, memo: str) -> Tuple[str, str]
 
 
 def pay_invoice(*, wallet_id: str, bolt11: str, max_sat: Optional[int] = None) -> str:
-    temp_id = f"temp_{urlsafe_short_hash()}"
+    wallet = get_wallet(wallet_id)
+    assert wallet, "invalid wallet id"
```
sgeisler, line 27: From the usage of this function I inferred that the wallet id is always valid at this point, so giving this function an invalid id should raise an error. I moved the check forward to avoid adding dummy payments to the wallet that wouldn't be removed due to an early exit of the function on failure.

lnbits/core/views/lnurl.py
```diff
@@ -1,8 +1,8 @@
 import requests
 
 from flask import abort, redirect, request, url_for
-from lnurl import LnurlWithdrawResponse, handle as handle_lnurl
-from lnurl.exceptions import LnurlException
+from lnurl import LnurlWithdrawResponse, handle as handle_lnurl  # type: ignore
```
sgeisler, line 4: Should be fixed in `lnurl` at some point

sgeisler:
Damn, wrong repo :see_no_evil:

sgeisler:
I opened the PR for the wrong branch unfortunately and it seems like I can't change it anymore. Maybe we can discuss it here anyway so I don't have to copy all my comments over and if we think it's ok you  can just cherry pick it on master?

lnbits/core/services.py
eillarra: The problem with this is that we need to call `get_wallet().balance` as late as possible to avoid parallel payments going through when balance is not enough. Two options here:

- call again `get_wallet()` when checking the balance, so the temp payment is also taken into account (we make 2 queries then)
- move the get_wallet and assert lower. the temp payment will be created in any case, but it would be deleted after the exception has been catched

I added the temp payments to avoid sending different payments while we wait for the response from the backend. If we have another solution to avoid this problem, and it ius better, we can move things around again.

lnbits/core/models.py
eillarra: Sure 👍

lnbits/core/services.py
sgeisler: Oh, I totally missed that, I kinda assumed something ORM-like to take care of fetching up to date values, but I remember seeing the raw queries. Will fix.

sgeisler:
Discussion of my inline comments seems finished, I'll rebase and reopen a PR to master for final review.

PR #26, opened 2020-04-28
refactor: remove wallet only dependencies from core

eillarra:
Some wallets have special requirements like `lnd-grpc` or `pylightning`.
These are now optional and only required if the user tries to use certain funding sources.

For example, the gRPC libraries require a decent time to be installed/compiled on a RaspiBlitz (see https://github.com/rootzoll/raspiblitz/pull/1135), so at the end of the day they will be using the `LndRestWallet`... I think it makes sense to remove this dependencies from the standard requirements, and make them optional.

wgsi dependencies have also been removed, as they can depend on the production installation.

All this shoud be well documented in the "funding sources" part of the documentation, that doesn't exist yet :)


arcbtc:
Whats the real reward?
A lot of users running nodes could will put off by the extra installs. Are we not trying to make running nodes easier?

requirements.txt
```diff
@@ -1,28 +1,20 @@
...
-flask-compress==1.4.0
+flask-compress==1.5.0
 flask-talisman==0.7.0
 flask==1.1.2
-gevent==20.4.0
-googleapis-common-protos==1.51.0
-greenlet==0.4.15 ; platform_python_implementation == 'CPython'
-grpcio-tools==1.28.1
-grpcio==1.28.1
-gunicorn==20.0.4
```
sgeisler, line 16: Were these ever needed as direct deps? I don't see them removed in the code/encapsulated with `try` anywhere.

sgeisler:
The CI failure seems to stem from you basing your work on master before #25 got merged. So a rebase should make it work imo.

requirements.txt
eillarra: `lnd-grpc` requires `grpcio`, `grpcio-tools`, `googleapis-common-protos`.

I removed the wgsi choice from the core library too: when deploying users should install `gunicorn` and `gevent` (`greenlet` is rewuired/installed by the latter). But I am not sure if we should keep them or remove them. You can run LNbits with `flask run` but in production is recommended to use a wgsi server line gunicorn.

In the end, it seems that the RaspiBlitz willl finally be using the gRPC libraries, so this PR can be ignored for now, and we can leave all dependencies there, even if some of them might never be used by the user...

eillarra: Regarding the extra CI: I am not used to GitHub Actions, I didn't realize I was adding another CI :) if we want to only work with GiHub Actions, is OK for me too, just moving the tests there.

arcbtc:
Having sat for a long time waiting for LNbits to install on a Raspiblitz, the install experience is not acceptable and will turn uses away.

Until LND has the complied gRPC built-in we should not include the lnd_grpc library, and the blitz should use REST.

PR #38, opened 2020-05-05
allow inkey endpoints to be accessed with adminkeys.

fiatjaf:


lnbits/core/crud.py
```diff
@@ -114,17 +114,22 @@ def get_wallet(wallet_id: str) -> Optional[Wallet]:
...
             """,
-            (key,),
+            (key, key),
         )
 
-    return Wallet(**row) if row else None
+        if not row:
+            return None
+
+        if key_type in {"admin", "adminkey"} and row["adminkey"] != key:
```
eillarra, line 129: Maybe we just have to use "adminkey" and "invoicekey" as key_type if "admin" and "invoice" are confusing, instead of allowing all options. We can easily search and replace the old values in the decorators.

fiatjaf: Ok. I was just confused about keys having one name and we having to specify a different name here as `key_type`.

PR #43, opened 2020-06-07
description_hash support, spark backend and lnurlp extension.

fiatjaf:
It takes the description hash as a hex-encoded string.

Raises an exception for c-lightning and OpenNode backends, since they don't support it.


lnbits/core/views/api.py
```diff
@@ -27,13 +28,21 @@ def api_payments():
 @api_validate_post_request(
     schema={
         "amount": {"type": "integer", "min": 1, "required": True},
-        "memo": {"type": "string", "empty": False, "required": True},
+        "memo": {"type": "string", "empty": False, "required": False},
+        "description_hash": {"type": "string", "empty": False, "required": False},
     }
 )
 def api_payments_create_invoice():
+    if "description_hash" in g.data:
```
eillarra, line 36: Shouldn't we check if at least one of the two parameters (`memo` or `description_hash`) has been included?

lnbits/wallets/lnbits.py
```diff
@@ -12,11 +13,16 @@ def __init__(self):
...
+    def create_invoice(self, amount: int, memo: str = "", description_hash: bytes = b"") -> InvoiceResponse:
         r = post(
             url=f"{self.endpoint}/api/v1/payments",
             headers=self.auth_invoice,
-            json={"out": False, "amount": amount, "memo": memo}
+            json={
+                "out": False,
+                "amount": amount,
+                "memo": memo,
+                "description_hash": hexlify(description_hash).decode("ascii"),
```
eillarra, line 24: `bytes.hex()` can be used here directly => `description_hash.hex()`
https://docs.python.org/3/library/stdtypes.html#bytes.hex

lnbits/core/views/api.py
fiatjaf: Can't we have `memo` as an empty string if nothing is provided?

fiatjaf: I forgot to add `"default": ""` to `memo`.

fiatjaf:
Because I'm stupid I've mixed multiple things in this same PR, so I renamed it.
And will merge without asking for anyone else's opinion.

PR #44, opened 2020-06-10
README improvements

dennisreimann:
Working in some things I noticed while setting up the project locally.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,7 +9,7 @@ LNBITS_DISABLED_EXTENSIONS="amilk,events"
 LNBITS_FORCE_HTTPS=1
 LNBITS_SERVICE_FEE="0.0"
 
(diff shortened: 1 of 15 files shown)


eillarra:
Maybe is a good moment to see what needs to be removed from the README. Ideally, dev docs (how to set up a dev environment, info about different wallets) should go in the `docs/` folder.

We can leave the README for a general overview and then point to the actual docs from there.

Docs are using GitHub pages and are hosted on <https://lnbits.org>

eillarra:
Looks good to me. Can you maybe squash the different commits?

dennisreimann:
Ok, here's a all-in-one version :)

eillarra:
Thanks, I hope everything went well, because I was just merging it :)

PR #54, opened 2020-07-21
Modify Dockerfile to use pipenv instead of pip

SteliosRammos:
Modified the Dockerfile which now does the following:

1. Runs the pipenv-install.sh script to Install all dependencies using pipenv instead of pip. It also check the FLASK_ENV value to see whether to install dev packages or not
2. Runs a boot.sh script to start a gunicorn server on port 5000 (single worker)

The LNBits docker image can be built and ran with the following commands:

$ docker build -t lnbits .
$ docker run  --name lnbits -v /lnbits/data:/home/lnbits/lnbits/data --env-file .env -p 5000:5000  lnbits

Docker-compose could further ease the setup to use a single 'docker-compose up' command but would require the installation of Docker Compose by the user, so I haven't included it in the PR. It's open to discussion though.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,10 +4,10 @@ FLASK_ENV=development
 LNBITS_SITE_TITLE=LNbits
 LNBITS_ALLOWED_USERS="all"
 LNBITS_DEFAULT_WALLET_NAME="LNbits wallet"
-LNBITS_DATA_FOLDER="/your_custom_data_folder"
+LNBITS_DATA_FOLDER=lnbits/data
 LNBITS_DISABLED_EXTENSIONS="amilk,events"
 LNBITS_FORCE_HTTPS=1
-LNBITS_SERVICE_FEE="0.0"
+LNBITS_SERVICE_FEE=0.0
 
 # Choose from LNPayWallet, OpenNodeWallet, LntxbotWallet, LndWallet (gRPC), LndRestWallet, CLightningWallet, LnbitsWallet
 LNBITS_BACKEND_WALLET_CLASS=LntxbotWallet
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,23 @@
 FROM python:3.7
 
-WORKDIR /app
-COPY requirements.txt /app/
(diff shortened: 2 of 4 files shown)


fiatjaf:
Is this good? Can we merge?

SteliosRammos:
It should be good to go. Perhaps there needs to be an update to the installation/running instructions as well since the they will look a little different (and simpler) with docker.

fiatjaf:
Do you have any idea why `Pipenv` has so many less dependencies than `requirements.txt`? I know I added `ecdsa` recently to `requirements.txt` and not to `Pipenv`, but even without this the difference would still be big.

We shouldn't have two different ways of managing dependencies.

SteliosRammos:
I am not entirely sure. Maybe the requirements.txt was generated with the command ```pip freeze > requirements.txt``` which lists all installed packages (and their dependencies). I believe pipenv manages the dependences in the Pipfile.lock instead, only keeping the explicit packages in the Pipfile. 

Would missing packages be picked up by the code cov checks?

eillarra:
My comments before @fiatjaf merges this :)

The Dockerfile in the root is used to deploy to lnbits.com so I would not change that file. But I do think we need to have a **deployment** section in the docs so I suggest we move this Dockerfile proposal there, as one of the deployment options (Docker).

Now, that being said, I have 2 doubts concerning this PR:

1. if this is a deployment container (so you are not contributing to LNbits and using Docker in your development stack) then we should recommend using requirements.txt instead of Pipenv for installing the dependencies. Pipenv is handy to manage the dependencies during development, and to update/check the package versions required by LNbits... but it does not make sense to install it in a deployment container if we already have the `requirements.txt`file.

2. the same about the `DEBUG` flag deciding it the `--dev` packages shoukd be installed. If it is a deployment container, we don't need all these extra packages because these are only used for testing, linting or code formatting. If you are not a contributor you don't need these, event with Flask running inn DEBUG mode.

Regarding Pipenv having less requirements... as @SteliosRammos says the lock file contains the same info as the requirements.txt file.

PR #57, opened 2020-08-02
use callback instead of callback.base to preserve port

frennkie:
As mentioned in https://github.com/rootzoll/raspiblitz/issues/1348.

In my point of view switching from `withdraw_res.callback.base` to `withdraw_res.callback` should not do any "damage".


frennkie:
This fixed in upstream: https://github.com/python-ln/lnurl/issues/5

eillarra:
requirements.txt has been updated in master: `lnurl==0.3.4`

PR #62, opened 2020-08-07
Update extensions.md

losnappas:
Add a skeleton, better than nothing.


losnappas:
I'm actually not sure if this statement: "views.py: The `/` path will show up as your plugin's home page in lnbits' UI. Other pages you can define yourself. The `templates` folder should explain itself in relation to this." is true, by the way.

arcbtc:
Thanks! So much to do

Issue #30, opened 2020-04-29
Websocket / webhook for invoice payment detection


Kukks:
This seems like a cool project! I'm investigating adding support into BTCPay to use an LNBits based wallet for accepting lightning. We do however some form of push from the server to BTCPay to notify of an invoice being paid. 

On an unrelated note, are you publishing any docker images?

sgeisler:
This will also reduce the load on the backend if users don't have to poll all the time. Most real node backends support some form of notification on payment finalization. For the ones that don't support that there could be a central thread polling all open payments and sending notifications instead of every request causing a separate poll.

arcbtc:
WebSockets need to happen at some point, this is as good an excuse as any. 
Linking LNbits and BTCPay server is an exciting concept

Kukks:
> 
> 
> WebSockets need to happen at some point, this is as good an excuse as any.
> Linking LNbits and BTCPay server is an exciting concept

Yes, I could integrate LNBits as if it was another lightning node in our lightning abstraction. Someone offering a BTCPay server to multiple merchants could then host LNBits alongside and use that as a way for them to accept lightning via their node!

Will need the websocket support and multiarch docker images though

fiatjaf:
Hey, just a heads up: LNbits now has an (still undocumented while we make sure it works) SSE endpoint that you can listen to to get an event every time an invoice payment is received (it listens for notifications in the background from each of the funding sources -- except lntxbot), matches the incoming payments to each LNbits wallet and dispatches the event there.

`curl https://lnbits.example/api/v1/payments/sse -H 'x-api-key: yourkey'` should work. It will give you something like this:

```
curl 'http://localhost:5000/api/v1/payments/sse' -H 'x-api-key: 4b0518a9a640af02b5165fd9879a756c'
event: keepalive

event: payment
[...]

PR #64, opened 2020-08-10
WIP: Use websocket

losnappas:
To implement websockets and "background processing," as the people at btcpayserver put it, how does this look? Currently this pr works for the QR code when you click on "receive," only for c-lightning, and I've only tested `wait_invoice`.

Any better ideas/thoughts welcome. It doesn't look too bad to me, but idk flask or python.

I almost certainly messed up the lockfile and requirements.txt in merge commit.

What do you think? Any hope of a future with websocketing?

Also, development environment uses long polling but it uses websocket otherwise.


losnappas:
Alright, there's support for LND (grpc & rest), as well as c-lightning.

As for the 3rd party wallets, however, I don't really wanna go through making accounts and stuff, so unless someone else steps up, or those are ignored, we might be in limbo for now.

Undoing the changes to `wallet.js` would let us keep the 3rd party wallets, but without websocket stuff. Not sure how well that would play out, though.

fiatjaf:
I'm sorry this PR got ignored, but it wasn't in vain, because I copied the motivation LND snippets from it when implementing my custom async payment listeners stuff, which (hopefully) supersedes this (although WebSockets are not implemented, but SSE is, and the default HTTP polling checks on wallet.js still remain).

Another difference is that the approach currently deployed listens to all invoices instead of to each individually.

See more here: https://github.com/lnbits/lnbits/issues/30#issuecomment-706460322 and in `tasks.py` files.

PR #69, opened 2020-08-17
infinite checking pending transactions notification in webinterface

ChuckNorrison:


ChuckNorrison:
maybe someone find a better way, seems to work for me

ChuckNorrison:
the infinite checking pending transactions is done, but seems there is still an issue with the payments[0]['status'] which is expected and unexpected in a bad state. Also the self funding errors will happen to hanging process payment notifications... seems there is some more to get aware of.

ChuckNorrison:
Very crazy circumstances found with the allow self payment.

If lnbit wallet creates and pays the same invoice, those funds from "nowhere" gets added. Its possible to trick the system and increase this wallet with some fake invoices. 

This is a very bad behaviour we should avoid in some way.

ChuckNorrison:
tested same on lnbits.com. Thats working like a charm.... something is very wrong here. I can double my lnbits accounts as i want... the lnbits.com acts as expected, he recv the fund, and pays at once. Mine only recv the funds but does not pay them for 500 Internal Server error

arcbtc:
lnbits.com is on top of opennode, and they handle internally.

I just put in a pr for self-payments https://github.com/lnbits/lnbits/pull/71

PR #71, opened 2020-08-19
Internal payments

arcbtc:
If payments are internal the pay request is decoded and handled internally.
I had to add an extra field to apipayments table to store the decoded invoice hash, as all these funding sources have their own internal checking_ids the use, the migration file seems to work and no data is lost

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ See [lnbits.org](https://lnbits.org) for more detailed documentation.
 
 Checkout the LNbits [YouTube](https://www.youtube.com/playlist?list=PLPj3KCksGbSYG0ciIQUWJru1dWstPHshe) video series.
 
-LNbits is inspired by all the great work of [opennode.com](https://www.opennode.com/), and in particular [lnpay.co](https://lnpay.co/). Both work as excellent funding sources for LNbits!
+LNbits is inspired by all the great work of [opennode.com](https://www.opennode.com/), and in particular [lnpay.co](https://lnpay.co/). Both work as excellent funding sources for LNbits.
 
 ## LNbits as an account system
 
--- a/lnbits/__init__.py
+++ b/lnbits/__init__.py
@@ -89,4 +89,5 @@ def migrate_databases():
 # ----
 
 if __name__ == "__main__":
-    app.run()
+    app.run(debug=True)
+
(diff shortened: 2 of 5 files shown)


arcbtc:
Thats part of the migration process, so no data is lost for people when they upgrade. It transfers all the data from apipayments to a new table apipayment

fiatjaf:
I'm working on a PR to merge this with #72 and #77.

Also, I think indexing payments by `hash` instead of `checking_id` is very necessary and healthy.
Is it ok to break the external API compatibility to do that? We can bump the version.

fiatjaf:
Actually I think I found a way that prevents any breakage to happen in any third-party service using the LNbits API.

frennkie:
@jiatjaf are you considering using a framework to support database migrations (like Alembic)?

fiatjaf:
I'm personally very against that, but who knows.
Did you see the new migrations framework? I think it is good.

eillarra:
I am not against it, but I don't think we need it right now. Current solution is stable and matches the _kiss_ philosophy of LNbits. We decided not to use SQLAlchemy for the same reason.

If in the future we see that the application has growing pains, we can reconsider it again.

frennkie:
Sounds absolutely fine to me. (I just personally hate SQL so much that I wouldn't touch anything that has no ORM - but that's just me).

frennkie:
Just in case this needs reminding: Manually written SQL queries in any code base are obviously huge red flags for a Pentest/Code Auditor... Is just too easy to introduce security vulnerabilities (e. g. SQL injections).

Frameworks help prevent this - but of course it's also possible to write secure code without them..! :-D

eillarra:
Luckily, the `sqlite3` standard library knows about SQL injections :)
https://docs.python.org/3.7/library/sqlite3.html

eillarra:
Also, it is not completely true that "frameworks help prevent this": you can still make mistakes every time you use raw SQL in an ORM. All ORMs give you the option to write raw SQL, because sometimes it just makes sense (performance being one possible reason).

Writing secure queries has nothing to do with having an ORM or not.

- https://docs.djangoproject.com/en/dev/topics/db/sql/ (check the big note at the top)
- https://docs.sqlalchemy.org/en/13/core/tutorial.html#using-textual-sql

frennkie:
> you can still make mistakes every time you use raw SQL

Exactly the point I wanted to make.

Thanks for clarifying.

Issue #74, opened 2020-08-30
TPoS extension throws 500 Internal Server Error on usage


ChuckNorrison:
After recent updates, i got following error if i enter the desired amount and click OK.

> create_invoice() missing 1 required keyword-only argument: 'description_hash'
> 500 INTERNAL SERVER ERROR

fiatjaf:
Thank you!

PR #76, opened 2020-08-30
add missing description_hash argument

ChuckNorrison:
assign empty bytes array as default value

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -7,7 +7,7 @@ from lnbits.settings import WALLET
 from .crud import get_wallet, create_payment, delete_payment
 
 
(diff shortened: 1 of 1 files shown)


ChuckNorrison:
This will fix https://github.com/lnbits/lnbits/issues/74

fiatjaf:
Oops, sorry, I should have seen this yesterday.

ChuckNorrison:
i saw your default variable as None. I tested this too, but it crashed with a type error. So my approach was to make empty byte array with .encode(), which worked for me. 

I will test LNBits again after LNURLp is supported from LNbits wallet, great work so far and iam very happy with your solution, thanks :+1:

PR #79, opened 2020-08-31
fix infinite pending transactions check

ChuckNorrison:
This should check for status in a more safe way and only result in a check pending transactions notification if payment was "valid".

### DIFF
--- a/lnbits/wallets/lndrest.py
+++ b/lnbits/wallets/lndrest.py
@@ -87,10 +87,16 @@ class LndRestWallet(Wallet):
             return PaymentStatus(None)
 
         payments = [p for p in r.json()["payments"] if p["payment_hash"] == checking_id]
-        print(checking_id)
-        payment = payments[0] if payments else None
+        payment = None
+        if payments:
+            for p in payments:
+                if 'status' in p:
+                    payment = p
+                    break
 
         # check payment.status: https://api.lightning.community/rest/index.html?python#peersynctype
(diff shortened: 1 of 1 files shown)


lnbits/wallets/lndrest.py
```diff
@@ -87,10 +87,16 @@ def get_payment_status(self, checking_id: str) -> PaymentStatus:
...
 
         payments = [p for p in r.json()["payments"] if p["payment_hash"] == checking_id]
-        print(checking_id)
-        payment = payments[0] if payments else None
+        payment = None
+        if payments:
+            for p in payments:
+                if 'status' in p:
+                    payment = p
+                    break
```
fiatjaf, line 95: I believe this is wrong, because if LND is returning more than one payment than something is very wrong and we should raise an error.

fiatjaf: The error is likely to be on the LND call above. It is querying for all payments, but it should be possible to query for a specific payment identified by the `checking_id`.

fiatjaf: On `create_invoice` above `checking_id` is being set to `None`. That is wrong. It should be set to a string that later will be used to query this payment through the API.

ChuckNorrison: i believe you are right, this issue is related to something else. this is just a workaround for missing payment status. I can try to get in deeper to understand the real culprit on this

fiatjaf: Meh, I was very wrong here, sorry about that.

PR #78, opened 2020-08-31
lndhub extension

fiatjaf:
Work in progress.

For operating an lnbits wallet from Bluewallet or Zeus.

Issue #63, opened 2020-08-10
LNURL compatibility for wallets


arcbtc:
First I will move the scanner, to outside the receive send,
![image](https://user-images.githubusercontent.com/33088785/89769520-27ee6900-daf5-11ea-8b9a-5699ac3e8066.png)
Then I will add support for lnurl-pay and lnurl-withdraw
Ideas welcome

PR #81, opened 2020-09-01
Internal payments + extra fields on apipayments + index payments by payment_hash, not checking_id

### LINKED ISSUE #72: Add `tag` field to payments to link them with extensions
What do you think of adding a "tag" field to payments -- incoming or outgoing -- for extensions to fill in themselves, so in the main payments table view we could should the extension icon along with each invoice/payment and improve the user experience?

Should we also have an `extra` field that would support extra metadata in each payment or something like that?

fiatjaf:
This PR contains #71 and closes #72 and #77.

It adds the columns `hash`, `preimage`, `extra` and `bolt11` to the table `apipayments`. `hash` is necessary for #71. `hash` and `bolt11` are necessary for #78. `extra` is kinda necessary for #63 and a good thing for the `lnurlp`. `preimage` is necessary for the overall health of the Lightning Network.

This PR does some big changes: it changes `get_payment` and `get_wallet_payment` to use `payment_hash` instead of `checking_id`, plus `create_invoice` and `pay_invoice` also return a `payment_hash` instead of `checking_id`.

On `POST /api/v1/payments` in the core API a field `payment_hash` is now returned. For backwards-compatibility the field `checking_id` is still present, but it contains the same `payment_hash`, so clients can keep using that same value on `GET /api/v1/payments/<payment_hash>` without changing any code. But this is undocumented behavior now. The canonical way to do things is to get the `payment_hash`.

It also contains changes to all extensions that relied on `checking_id` to make them use `payment_hash` instead.

This PR is a bundle of different things that could contain many hidden bugs mainly because I wanted to put everything in a single migration and also because I'm disorganized.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,25 +1,33 @@
-name: Run Linters
+name: Linters
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
 
 jobs:
-  mypy:
+  black:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v1
-      - name: Run MyPy python type checker
-        uses: jpetrucciani/mypy-check@master
-        with:
-          path: 'lnbits'
-
+      - uses: actions/checkout@v2
+      - run: sudo apt-get install python3-venv
+      - run: python3 -m venv venv
+      - run: ./venv/bin/pip install black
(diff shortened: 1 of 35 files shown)


fiatjaf:
tags

fiatjaf:
2020-09-02-191700_621x410_scrot

fiatjaf:
`black` and `mypy` still fail, but this will have to be fixed later, it's too weird.

PR #84, opened 2020-09-04
migration tracking.

fiatjaf:
Also change the default run command to a single `main.py` file (because I thought I needed it to package the app with pyinstaller but I think it's overall a good idea to not depend on the `flask` command, that is confusing for people who aren't flask developers) that runs the migrations every time you start.


fiatjaf:
This worked from a brand new installation, worked from an installation that was already on core@v2, and works if you run it infinite times sequentially. I didn't test how it does for an installation that is on core@v1, but it should work too.

eillarra (approved):
Looks good.

PR #90, opened 2020-09-07
Transactions

fiatjaf:
It was bothering me a lot that any error from any wallet backend left our database in an unfixable situation, with a pending payment that wasn't really pending but we couldn't cleanup because we weren't sure.

There were also other potential bugs and race conditions that could be caused by not wrapping these sequential database calls into transactions,

This approach works, but I'm not sure it correct from the Flask level.

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -1,6 +1,6 @@
 import importlib
 
-from flask import Flask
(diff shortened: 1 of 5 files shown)


eillarra (approved):
Nice!

Issue #80, opened 2020-09-01
Migrate from Flask to Quart


fiatjaf:
I think a precondition to https://github.com/lnbits/lnbits/pull/64 would be to migrate from Flask to an async framework that supports websockets or server-sent events out-of-the-box.

I think having a websocket API that only works on production environments with gunicorn or whatever isn't developer-friendly and against the LNbits _ethos_ (yes, this word is used when you don't have anything else to say). It would require us to keep supporting the current polling methods plus the websocket, which would rapidly turn into a mess.

It seems there are some frameworks similar to Flask that we could migrate to: Sanic, Vibora and Quart.
Of these, Quart's selling point is having an API identical to Flask, so I guess that is an easy choice.

PR #93, opened 2020-09-14
migrate from flask to quart.

fiatjaf:
Closes https://github.com/lnbits/lnbits/issues/80.


arcbtc:
Getting white screen of death on extensions :(
image

fiatjaf:
You didn't need to delete any dbs, but there were other dependencies involved: `pipenv sync` is the command to be run when you just want to get your local environment synced with some commit.

arcbtc:
Still getting a white screen when clicking extensions
Pulled -> ```pipenv sync``` ->  ```pipenv run python -m lnbits```

fiatjaf:
I'll go ahead and merge this because everything seems to be fine and I want to click the merge button.

The mypy check is failing but it shouldn't. it runs fine on my machine with all the exact same versions, so I'm trying to debug that but it's weird.

PR #94, opened 2020-09-14
Admin setup (WP type form and new DB table to replace .env?)

arcbtc:
## EDIT, skip to updated proposal: https://github.com/lnbits/lnbits/pull/94#issuecomment-692524432 

Shall we add an admin var to .env and an extra "Update balance" button to admin wallets?
Its frustrating for node users to have to send funds out, then back in to get a balance in LNbits, this would allow a simple DB change + privileged admin extensions, such as usermanager
I have added the var, just need to add the dependent "Update balance" button


arcbtc:
added tasteful button that is active when admin
image

fiatjaf:
Please run `make` then push again, just to format the code!

fiatjaf:
I think it will be too hard for users to identify their user id and change the environment variables with it to gain admin powers.

What if there was a CLI command that they could run to list wallets and credit them? We can be sure the person running the CLI command already is the admin anyway, and since this is expected to be done only once and only by some users I think it is reasonable.

arcbtc:
I see your point, its just following the same example as LNBITS_ALLOWED_USERS, which is the same process

arcbtc:
Sorry I cant get prettier to behave, using make and in sublime :(

arcbtc:
Example video
https://youtu.be/jOJNcbTFDp4

arcbtc:
Nah this isn't right, too convoluted 
How about the first user is automatically an admin and has access to an admin extension, which you can credit wallets in etc. If the host screws up they can use CLI to set a new admin?

arcbtc:
EDIT: I have changed the title as there seems to be some support for setting up a WordPress style setup page for initial install, WP do this for all the reasons we should prob do it. 
All the .env variables could be moved to a DB and the setup page either populated on launch, or during install on something like Raspiblitz.
I propose a completely separate form, much like WP 
wordpress-installation-screen

rootzoll:
I like the idea of when you install LNbits the first time you get a "Setup-Page". The setup data could be part of the database or create the env file itself. Maybe Database is better because then later on people just need to care about having a "Database Backup/Export" to move between instances.

fiatjaf:
I like this, but I think we should keep support for environment variables and just show the setup screen or use the variables from the database if values aren't found in the environment.

arcbtc:
Checkout adminvar branch for a demo admin setup form at /admin
image

rootzoll:
@arcbtc like the mockup :) way to go

> I like this, but I think we should keep support for environment variables and just show the setup screen or use the variables from the database if values aren't found in the environment.

@fiatjaf I would recommend having the env variables as an override for the database values. If there is a value in then env or a value in env gezs changed it will be copied that way into the database and the setuo UI dies not allow that value to change (e.g. disable form field). This way all data is in database and easy to backup/migrate but also is controlable from env vars ... and env vars are always an easy way for host systems like RaspiBlitz to set config from the outside.

arcbtc:
I was going to populate the DB from .env on startup, to avoid you having to change anything. The setup page will pull those details and populate the form, so users can set things like Title, etc, but the funding source info will be prefilled

arcbtc:
The only issue with env overriding that afterwards, would be if they want to change the info via the admin extension/DB, the env would override anything they did

arcbtc:
Works pretty well now, saves env in DB, can be edited from ```/admin``` and ```/admin?usr=<admin_usr>``` once the initial setup has been done . 
Not sure how best to replace all the os.enving with DB calls @eillarra?

arcbtc:
Made an admin extension, only accessible by admin, although the gap when your not admin is a rather annoying
Screenshot from 2020-09-18 00-29-13
Screenshot from 2020-09-18 00-29-52

fiatjaf:
I think there will be some people comfortable with environment variables and for these we must keep environment variables working.

But copying these to the db seems... weird.

We can just read from env if env exists and from db if not. If someone is using env for everything we can just not add any row to the db.

arcbtc:
Without copying the variables and moving towards the data being in the db the details cannot be changed through the GUI, which voids the point of a setup page/admin extension

There will also be times where data exists in both

arcbtc:
Admin form now pulls envs, saves them in the db and can be edited.

Whats the cleanest way to;
Use the DB instead of envs?
Use the user defined custom primary/secondary colors in place of deep-purple and purple?

Or is it just a case of changing all the files?

Any pointers would be very welcome!

arcbtc:
```/admin```, ```/admin?usr=<admin-ID>``` and the admin aextension all work here for saving variables https://github.com/lnbits/lnbits/tree/adminvar
When I try and change settings.py though I get an error  https://github.com/lnbits/lnbits/tree/GUIadmin
image

zanderisrael:
Is this dead? Maybe I can pickup with some general idea of how to proceed from here…

arcbtc:
Not dead, just on the list of many things to do. Give me a few days to take a look and update what I can, then please do pickup, would be great to get this moving again

arcbtc:
Started fresh PR and branch https://github.com/lnbits/lnbits-legend/pull/553

PR #111, opened 2020-10-09
Add dependency requirement for CLightning

RohanKapurDEV:


fiatjaf:
This is noted [here](https://github.com/lnbits/lnbits/blob/master/docs/guide/wallets.md#clightning) already.

PR #122, opened 2020-11-10
lnurl-auth from lnbits wallets to services.

fiatjaf:
2020-11-10-001936_616x496_scrot

### DIFF
--- a/lnbits/core/models.py
+++ b/lnbits/core/models.py
@@ -1,4 +1,6 @@
 import json
+import hashlib
+from ecdsa import SECP256k1, SigningKey  # type: ignore
 from typing import List, NamedTuple, Optional, Dict
 from sqlite3 import Row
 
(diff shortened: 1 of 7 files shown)


arcbtc:
Wow, nice work.
As its deterministic and based on the user id, can an lnbits provider check an lnurl-auth call against their db of users.
i.e. If a cafe was running lnbits and giving patrons wallets, could lnurl-auth calls for an additional service be checked against the lnbits db, so an lnurl-auth powered toilet could restrict access to only patrons with the cafes lnbits wallet?

fiatjaf:
It is possible, but the toilet door would have to derive the keys for every wallet in the LNbits DB and see if they matched the key of the person trying to enter.

Or there could be a toilet extension that kept a cache of these keys and allowed the toilet door query them.

lnurl-auth is mostly about making linkability the most impossible thing. In this case it is possible only because the LNbits server knows the seeds for everybody's lnurl-auth keys, and even them, as you see, it is not easy.

I think this may not the best way to restrict entry to the bathroom.

PR #126, opened 2020-11-20
fix #125 qrcode printing

### LINKED ISSUE #125: LNURLw: Printing QR codes seems off
When I create a series of vouchers with the LNURLw plugin and try the printing option first it looks nice: A piece of paper with some qr codes on it ... but when i try to print, it doesnt look the same and is not usable for printing a large amount of voucher.

This is how it looks when I try to print on Firefox (macos):
Bildschirmfoto 2020-11-19 um 17 34 02

Similar result with Chrome(macos):
Bildschirmfoto 2020-11-19 um 17 33 22

First I tested with LNbits on raspiblitz-tag but I also get the same results from the latest master.

rootzoll:
This fixes the qr code voucher printing on my raspiblitz. It would be great to have it merged and moved the raspiblitz tag to. thx.

### DIFF
--- a/lnbits/extensions/withdraw/templates/withdraw/print_qr.html
+++ b/lnbits/extensions/withdraw/templates/withdraw/print_qr.html
@@ -69,12 +69,13 @@
       border-width: 3px;
       border-style: solid;
       border-color: green; 
+      padding: 0px;
     }
     .row {
       border-width: 3px;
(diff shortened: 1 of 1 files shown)


rootzoll:
I dont know why mypy/check fails - its just CSS fixes.

PR #135, opened 2020-12-28
subdomains extension

grmkris:
Work in progress...

Currently completed:
- Domain table
- Subdomain table
- Public frontend for every domain we are selling
- Working CRUD operations
- calls to cloudflare to create a new dns record when invoice is paid
- validation for ip format 
- validation for subdomain format
- validation if subdomain already taken
- code formatting to follow standards
- review comments
- call to cloudflare to check if subdomain is available before creating invoice 
- frontend bug when editing domain information
- 
Todo:  
I will try to automate API docs generation. I'll do this in a seperate PR


lnbits/extensions/subdomains/templates/subdomains/display.html
```diff
@@ -0,0 +1,173 @@
...
+        <h5 class="q-my-none">{{ domain_desc }}</h5>
+        <br />
+        <q-form @submit="Invoice()" class="q-gutter-md">
+          <q-input filled dense v-model.trim="formDialog.data.email" type="email"
+            label="Your email (optional, if you want a reply)"></q-input>
+          <q-input filled dense v-model.trim="formDialog.data.subdomain" type="text" label="subdomain you want">
+          </q-input>
+          <q-input filled dense v-model.trim="formDialog.data.ip" type="text" label="ip of your server">
+          </q-input>
+          <q-input filled dense v-model.trim="formDialog.data.duration" type="number" label="{{ domain_cost }} sats per day">
```
fiatjaf, line 17: This field is confusing, what is the user supposed to write here? The number of days?

fiatjaf:
The design seems good to me, except there is no way to extend the domain lease time.

Also, to call Cloudflare upon an invoice payment, take a look at `lnurlp/__init__.py` and `lnurlp/tasks.py` and `core/tasks.py` if you haven't already.

fiatjaf:
I don't see the need for an email address or how you're using it. Apparently you're not using it for anything too. What is the idea?

grmkris:
> I don't see the need for an email address or how you're using it. Apparently you're not using it for anything too. What is the idea?

The email is optional and would be advisable to use so you can contact the person if the dns registration fails or something like that...   
But yes you are right if I remove email then it is one less thing to worry about. And user can contact the owner anyway by providing the payment hash.

grmkris:
> Also, to call Cloudflare upon an invoice payment, take a look at `lnurlp/__init__.py` and `lnurlp/tasks.py` and `core/tasks.py` if you haven't already.

Thank you will check that.  

> The design seems good to me, except there is no way to extend the domain lease time.  

About extending the lease time currently I'd need a interface for users to check their subdomains status.. or atleast an api so I can integrate a bot into this...   
Instead I'll add a check in create subdomain method that checks if the subdomain is already registered; if it is registered and the IP provided is the same I'll simply extend the duration column.  If the IP doesn't match I'll throw an error....  
When the duration is over i'll drop the domain from cloudflare and database.

lnbits/extensions/subdomains/templates/subdomains/display.html
grmkris: yeah number of days.. will make it clearer

arcbtc:
Looks great. Amazing work in such a short period of time

fiatjaf:
Is there a way to do CNAME records, TXT, MX?
At least CNAME and AAAA should be possible!

grmkris:
> Is there a way to do CNAME records, TXT, MX?
> At least CNAME and AAAA should be possible!

Yes it is possible to do any kind of records.. I just wasn't sure what to include for security reasons...   
I'll give an option to chose what you allow when creating a domain form.

fiatjaf:
I personally only know and use A, AAAA, CNAME, MX and TXT.

grmkris:
hey @fiatjaf I have another question. Upon taking a look at tasks.py file and __inity.py I conclude that crud.py should only be used for crud operations. And tasks that should happen when invoice is paid are executed by specifying those listeners...  
Also while doing this a method that checks if invoice was paid should only do that and the actual listener must update the values in databse... I'll work on this tomorrow...

Also now I added the option so that the operator can choose which types of DNS records they allow to sell..  

Next I have to do:
- some validations on IP and domain format...  
- There are some bugs with frontend when updating the domain form..  
- code formatting
- make some checks when registering new subdomain (subdomain already taken etc...)

grmkris:
Today I completed the following:  
- validation for ip format 
- validation for format
- validation if subdomain already taken
Both validation methods are stored in `util.py` file. I am coming from java spring world and checks like those can be done using custom annotation.. I see that it is similar here with decorators. I didn't bother researching and just called those checks inside the controller...  
 
- refactored code as @fiatjaf recommended. Now I am using listeners to mark invoice paid instead of relying on the frontend pingig the endpoint every x ms...

- code formatting to follow standards (black & prettier)

grmkris:
I added checks to cloudflare before registering a new domain...   This way we make sure that when user pays for the invoice the domain can be registered.  
this way I was also able to remove IP and subdomain validations..

grmkris:
this PR can be merged... 
it has ben sucesfully deployed on: https://lnbits.grmkris.com/ and working with no problems

PR #136, opened 2020-12-31
captcha extension added

pseudozach:
Not sure if it's useful but I finished it and wanted to share it. 
[quick demo](https://www.youtube.com/watch?v=HLxUdppn0k4)

Basically it generates a code snippet any website can integrate to request payment as a proof of humanity aka captcha.
After payment, website can check if checkbox is checked and allow the user to continue. 

Surely at this stage it will be easy to bypass but if people use it, we can improve further.

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -14,7 +14,7 @@ from .proxy_fix import ASGIProxyFix
 from .tasks import run_deferred_async, invoice_listener, internal_invoice_listener, webhook_handler, grab_app_for_later
 from .settings import WALLET
 
-secure_headers = SecureHeaders(hsts=False)
+secure_headers = SecureHeaders(hsts=False,xfo=False)
 
 
 def create_app(config_object="lnbits.settings") -> QuartTrio:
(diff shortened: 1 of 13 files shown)


fiatjaf:
I think the way captcha works is the visitor passes it, then gets assigned an id from the captcha provider, which is a different server, then when the visitor submits his comment or something it sends that id too and the website that is receiving the comment uses that id to check with the captcha provider if it is allowed. Right?

Now I don't know if you did this here, but if you don't it should be easy to adapt. Using payments as recaptcha and payment hashes as IDs is a great idea I think.

pseudozach:
> Now I don't know if you did this here, but if you don't it should be easy to adapt. Using payments as recaptcha and payment hashes as IDs is a great idea I think.

Agreed great idea. I didn't have it before but now it's done. Added a hidden text field "captchapayhash" to the client side snippet and it can be used by the server-side to check with lnbits if the invoice was really paid.

fiatjaf:
I don't know why this wasn't merged yet. It's almost as if this repository has no maintainer!

PR #146, opened 2021-02-11
Bleskomat extension for lnbits

chill117:
This is basically done, but still marked as work-in-progress. My local lnbits setup is using an old version of sqlite3, so some of the database migrations are failing. Once I've had a chance to fully test the extension's UX flow, I will remove the WIP label.


fiatjaf:
Coverage is terrible. Get it to at least 90% so we can merge.

chill117:
I have now tested the full UX flow and it is working as expected. I've removed the WIP label.

chill117:
Are there any blockers that I should address with this pull-request?

fiatjaf:
Is this already ready to merge?

I was waiting for you to finish before reviewing and merging.

fiatjaf:
Interesting. I should steal the fiat exchange rates code from you for my extensions.
Or maybe that should be a core util in LNbits.

chill117:
> Interesting. I should steal the fiat exchange rates code from you for my extensions.
> Or maybe that should be a core util in LNbits.

Sure, it could be extracted into a core component. It's probably something that should be done once and made available to all the extensions.

PR #148, opened 2021-02-12
Display QRcode for easy wallet export

rafaelpac:
This adds a QRcode of the "url_root/wallet?usr=#####&wal=####" so that
the user can create a wallet on desktop and easily use it on his phone.
Also, in a use case where the User Manager extension is used, the
manager can create a wallet and show the QRcode for the user to take it
(e.g. in a hotel front desk the clerk creates it and the user takes it
in his phone browser, to pay for laundry, massage, soda machine...)


fiatjaf:
2021-02-13-214027_510x483_scrot

PR #152, opened 2021-02-26
Creates LNURLw image folks can embed 

arcbtc:
The image is generated from the LNURL id, so unique LNURLws are compatible.
Makes sense to add the same function to LNURLp


arcbtc:
Sorry, need to format with prettier etc. Will also add lnurlw API wndpoint

fiatjaf:
Is this ready to merge?

fiatjaf:
Ready or not, I'm merging.

PR #154, opened 2021-03-07
OfflineShop extension

fiatjaf:
2021-03-07-170551_1184x677_scrot

```
How to use

    Register items.
    Print QR codes and paste them on your store, your menu, somewhere, somehow.
    Clients scan the QR codes and get information about the items plus the price on their phones directly (they must have internet)
    Once they decide to pay, they'll get an invoice on their phones automatically
    When the payment is confirmed, a confirmation code will be issued for them.

The confirmation codes are words from a predefined sequential word list. Each new payment bumps the words sequence by 1. So you can check the confirmation codes manually by just looking at them.

For example, if your wordlist is [apple, banana, coconut] the first purchase will be apple, the second banana and so on. When it gets to the end it starts from the beginning again.

Powered by LNURL-pay. 
```


fiatjaf:
photo_2021-03-14_14-35-26

PR #171, opened 2021-04-08
Adds WatchOnly wallet and SatsPayServer extensions

arcbtc:
WatchOnly wallet uses Embit, and has been audited/contributed to by author of the Embit library Stepan Snigerev 

SatsPayServer can be used to make BTCPayServer like charges, either just to lnbits wallet, or to WatchOnly wallet if activated


Pipfile
```diff
@@ -24,8 +24,12 @@ quart-trio = "*"
 trio = "==0.16.0"
 hypercorn = {extras = ["trio"], version = "*"}
 sqlalchemy-aio = "*"
+<<<<<<< HEAD
+embit = "*"
+=======
 pyqrcode = "*"
 pypng = "*"
+>>>>>>> master
```
fiatjaf, line 32: Just delete these extraneous lines here.

fiatjaf:
ACK, except for the broken Pipfile.lock, which I'll fix.

PR #174, opened 2021-04-11
fix localhost:5000 in captcha.js

pseudozach:
bug from telegram conversation.


fiatjaf:
Please check if this is ok after fixing the conflict.

pseudozach (commented):
tested and works.

PR #185, opened 2021-04-23
note on virtualenv installation

supertestnet:
Added missing step about installing python3 virtual environment


fiatjaf:
Maybe add a comment on that line like `# (on ubuntu) or whatever way you have to install global python packages on your system`.

arcbtc:
Aye, thats true

PR #190, opened 2021-05-10
Adding additional step

keblek:
for some reason my setup needed quart app environment variable


fiatjaf:
That shouldn't be necessary because it's in the `.env` file already, I don't understand why you needed it.

keblek:
I just realized that this was needed in my messed up enviroment and I didnt realize how messed up it was. Tried it on a clean ubuntu install and indeed it works. My apologies.

PR #210, opened 2021-06-11
Fiat invoices + Splitpayments/docs

talvasconcelos:
Allow for invoices to be created in fiat
Updated documentation for split payments extension


fiatjaf:
Merged manually with fixes to the Prettier issue.

PR #221, opened 2021-06-16
Royalties

gitnonymous:


fiatjaf:
I don't understand this. What does it do?

gitnonymous:
> I don't understand this. What does it do?

It's a way to collect fees on every invoice that is created and paid by an extension. When you create an invoice you, you also create a royalties payment, which is stored in a DB. After the invoice is paid, it requests an invoice from the accounts which are in the royalties payment. The accounts return the payment_request and the invoice is paid. On every instance of lnbits it will pay royalties. The Royalty accounts are to be added to the config.json file. The Royalty function pulls the array of Royalty accounts and creates a royalty entry in the DB to be paid when the invoice is paid.

PR #222, opened 2021-06-17
QRCode for LNURLw points to /?lightning 

tradewind886:
discovered issue through twitter where certain wallets can't scan LNURL QR codes from LNBITS

https://twitter.com/wpeace4886/status/1405354548307263493

Not sure why the QRcode includes the domain name, makes for a funky redirect.  Made a change to just use 'lightning:' prefix 

Tested on local instance on multiple wallets, works as intended now


tradewind886:
closed because change interferes with instawallet setup behavior

PR #228, opened 2021-06-20
Fix concurrent withdrawal behavior

evildido:
The current extension authorizes concurrent withdraw and
In this case, 'used' and 'wait time' are not properly updated by api_lnurl_callback function.
As a result, the usage limit can be exceeded if too many concurrent withdrawal are performed at the same time. 

This PR is a proposition to fix this behavior.


fiatjaf:
@arcbtc please take a look at this. I'm not familiar with this extension.

arcbtc:
Looks good to me as well. Nice spot @evildido !

Issue #233, opened 2021-06-28
Saving wallet to localstorage, download wallet file, login system, UI


arcbtc:
Eventually for saving wallets we want a number of options:

- Account as URL
- Save to local storage
- Download wallet file 
- Login user account
- LNURL-auth

Then in the wallet we can have a few icons for the different options:
![image](https://user-images.githubusercontent.com/33088785/123623050-ed70e600-d804-11eb-9e0b-b1252b7c9ae3.png)

On the frontpage we can have:
![image](https://user-images.githubusercontent.com/33088785/123622665-6cb1ea00-d804-11eb-91ff-156816824610.png)

In .env, the admin can delete to limit options
![image](https://user-images.githubusercontent.com/33088785/123623674-a59e8e80-d805-11eb-95c5-94f5c8cc4c69.png)

talvasconcelos:
Login and Admin extension will bring these features.

Issue #117, opened 2020-10-17
Store wallet and user credentials on localStorage


fiatjaf:
I thought it would be enough to rely on browser history to access my LNbits wallets, but after creating a hundred wallets and deleting the database some times and creating wallets again I'm pretty sure I will never find the wallet I want ever again.

Since we already leak previously used wallets into the browser history, what if we also stored them on localStorage and presented the user with a list of previously used wallets at the homepage of LNbits?

If someone wants a temporary instawallet they really need to use private-browsing mode or something.

Or maybe this feature could be turned off by the server owner.

HamishMacEwan:
This would give access to wallets created at the browser by the user who created them only?  That may be implicit, but just wanted to be sure.

That being the case, and with my own management issues (greatly helped by the appearance of the `Wallet name` in the <title>) has improved ease of management a great deal.  I wonder then if adding the `Wallet name` to the URL might be another affordance worth considering.

TL;DR Ack lamation.

fiatjaf:
It would give access to the wallets to anyone using that same browser.

HamishMacEwan:
So that's the same risk as auto-login on a shared browser, which may be reasonably well understood.  Unless you're on a public or kiosk browser, not something that's common or impossible to avoid.  Is the localStorage protected by profile login, as in Firefox and others?

As you note there's private mode too.

jcspocket:
It would not be adding any additional risk really.  In some ways making it more transparent that the risk already exists. 

As fiatjaf mentioned, you can already access browser history to get into old wallets.  Users might not be aware of that.

Making it show the previously accessed wallets on the front page makes it more obvious.

talvasconcelos:
Can this be closed? @fiatjaf ?

PR #236, opened 2021-06-29
Save accounts

talvasconcelos:
- Allow for saving account in localstorage 
- preparing UI for future methods
- fixes issues #233 #117
Draft because I'm having issues with the icons not showing on the wallet page, to choose from various methods


arcbtc:
The main point of saving to localstorage is to not expose your url. So base.js needs to remove the usr/wallet in url
When no variable is selected in .env, then none of the options should be available

PR #240, opened 2021-06-30
Win lose

gitnonymous:
Ready for review and pull


arcbtc:
I'm still unsure how this extension works/what it does. The auto-create button didn't work for me?

arcbtc:
I want to understand, as it looks great. Well done on the api docs. The whole "create user" bit isn't working for me, it might require a little more error reporting, if the wallet id doesnt exist, as I am just putting in random strings

arcbtc:
Got a few errors
image

arcbtc:
You'll need to run black as well. 
I have put you back to draft temporarily

gitnonymous:
> I'm still unsure how this extension works/what it does. The auto-create button didn't work for me?
working on mine even with usr management not enabled.

gitnonymous:
> I want to understand, as it looks great. Well done on the api docs. The whole "create user" bit isn't working for me, it might require a little more error reporting, if the wallet id doesnt exist, as I am just putting in random strings

if you use auto create it will create a "user manager account" the name will be a random id. the random id is then used to create the user in the win/lose db.  if manual you can use a pre-exting ID, say on your server database. You can just put the id in the ID field and leave the accounts blank. it will auto create but use the ID instead of random.  The "full manual". if you already have a lnbits account you can input the id and account details from that to create the win/lose user. Needs to go in the readme doc probably.

gitnonymous:
> > I'm still unsure how this extension works/what it does.

Imagine an arcade machine. You put money in and you get game credits. This is similar but when you receive sats, you instruct winlose to add game credits to the account. Now you can deduct game credits when they lose or if they win you can add credits or sats to their account^^

arcbtc:
Why not just use sats?
What are the use cases?

gitnonymous:
> Why not just use sats?
> What are the use cases?

Case 1. Mobile gaming industry. Pay to play. existing user ID's can be used to create win/lose accounts. You pay with sats to fund the account. Credits make it easy for the player/customer to see how many plays they have left. Companies that have a suite of games. You can use exisiting credits to play.

Case 2. indie gamer can set up their own lnbits and can manage simple api calls to manage users, game credits and receive sats.

Case 3. Betting industry and their customers. They understand $20 equals 20 credits/spins/plays. Ever changing exhange rates on sats is confusing. Whole numbers that don't fluctuate are understood.

PR #248, opened 2021-07-05
Add Umbrel Tor to clearnet guide

talvasconcelos:
Added a link to the guide for accessing LNbits on Umbrel, behind to Tor, through clearnet


arcbtc:
Added guide to install docs, as was worried frontpage will get bloated

PR #253, opened 2021-07-06
Update wallets.md

lc0des:
Correcting wallets.md configuration lnd-rest-macaroon part


arcbtc:
Thanks. Added HEXSTRING as an option, because you can also use macaroon path

PR #254, opened 2021-07-06
updated with troubleshooting

lc0des:
added section troubleshooting
* added tcp layer check 
* added ssl/tls steps


arcbtc:
Useful, but too much bloat for install docs. Create a medium article or something similar we could link to in the docs

PR #260, opened 2021-07-07
Re commit of the support tickets

talvasconcelos:
Just trying to see if a "re merge" of my local code. tickets work flawlessly on my local/test lnbits


arcbtc:
I dont think this will work.
I have the same scenario as you. It seems the webhook doesn't work behind url

PR #262, opened 2021-07-08
Fix LNURL-p and LNURL-w with lntxbot backend

talvasconcelos:
changed cheking_id to a random string

### DIFF
--- a/lnbits/wallets/lntxbot.py
+++ b/lnbits/wallets/lntxbot.py
@@ -1,6 +1,7 @@
 import trio
 import json
 import httpx
+import random
 from os import getenv
 from typing import Optional, Dict, AsyncGenerator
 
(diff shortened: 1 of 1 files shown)


oriolpont:
See my PR: I propose using the hash of the response instead of a random value. This ensures that the payment will get filled only once

oriolpont (changes requested):
There's an issue for invoices that get filled **externally**. There we do want to keep data["payment_hash"], right? `get_payment_status` will query it.

oriolpont (changes requested):
For internal payments, you need to change fee_msat and preimage too, because they are not there in the response and it will throw error without them. See #263 for reference

talvasconcelos:
I believe @fiatjaf fixed at lntxbot!

oriolpont:
Even if it's solved from the Lntxbot side, I still think the Lnbits side could handle the API response better. As it is, we send the request and parse the response strongly assuming that `payment_hash, fee_msat `and` payment_preimage` keys will always be there.

That's risky, for the instance's admin, especially for pay_invoice. If the request resolves successfully, we should probably assume that the backend has got debited, even when we cannot parse the response.

oriolpont (changes requested):
```
        checking_id = data.get("payment_hash") or label
        fee_msat = -(data.get("fee_msat") or 0)
        preimage = data.get("payment_preimage")
```

talvasconcelos:
> ```
>         checking_id = data.get("payment_hash") or label
>         fee_msat = -(data.get("fee_msat") or 0)
>         preimage = data.get("payment_preimage")
> ```

I'll check it with @arcbtc and @fiatjaf and make the changes if they agree...

PR #263, opened 2021-07-08
let Lntxbot have internal transfers correctly handled

oriolpont:
Invoices fulfilled internally within lntxbot users (including within one same user) do not respond with any payment_hash or payment_preimage. Still, if the reply has succeeded status it means it has gone through. We can handle the situation by giving a checking_id based on the hash of the response (it will not be directly queriable, but at the same time it needs to be unique in the DB), assign a blank fee to value 0, and leave the preimage blank

### DIFF
--- a/lnbits/wallets/lntxbot.py
+++ b/lnbits/wallets/lntxbot.py
@@ -99,9 +99,9 @@ class LntxbotWallet(Wallet):
             return PaymentResponse(False, None, 0, None, error_message)
 
         data = r.json()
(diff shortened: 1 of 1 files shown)


oriolpont:
Do NOT merge as is. This assumes that the hash of the response to the api request is unique, which might not be a safe assumption. #262 proposes a proper random value, which guarantees uniqueness checking_id in a rather stronglier way (or at least one that is more resistant to external attack)

talvasconcelos:
@fiatjaf

PR #274, opened 2021-07-22
No more pending tx with LNPay

Praexor:
No more pending tx with LNPay, the solution was proposed by @oriolpont and it is working awesome!! cheers!


fiatjaf:
This is not really a fix, it's a hack. OK if you apply it locally, though.

PR #285, opened 2021-08-04
Stream Alerts: Enforce sensible length limits on Name and Message

Fittiboy:
This adds `maxlength`s of 25 and 255 on the donor name and donation message respectively.
In case of inspect element abuse, anything longer than this gets truncated by the backend.
The limits were chosen like this:
 * Maximum length for Twitch usernames is 25 characters
 * Streamlabs' native tipping page has a maximum configurable message length of 255 characters


Fittiboy:
Apparently this causes some inconsistent bugs, I'll have to do more testing

Fittiboy:
I applied some dark magic, by doing the same thing at a later point and spinning three times in a circle. That somehow seems to have fixed this bug.

PR #295, opened 2021-08-13
Adds flat rate to tickets 

arcbtc:
As suggested by Darthcoin in lnbits telegram.
Pretty much done, just need tidying and testing

### DIFF
--- a/lnbits/extensions/lnticket/crud.py
+++ b/lnbits/extensions/lnticket/crud.py
@@ -49,7 +49,7 @@ async def set_ticket_paid(payment_hash: str) -> Tickets:
         amount = formdata.amountmade + row[7]
         await db.execute(
             """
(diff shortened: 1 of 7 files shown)


Fittiboy:
What are your thoughts on adding a third option with this PR:
1. *Cost per word*
1. **Flat rate**
1. **User choice**

With "User choice," the extension would essentially function as a tipping page.
It's in fact exactly what I've done for the Stream Alerts extension (+ using SatsPay charges to enable on-chain).
I currently accept tips via lnurl-pay links, but most tips I receive don't have messages attached. Some wallets don't support comments, and some don't make it obvious they do. In the worst case, they don't support lnurl-pay at all.
Having a nice tipping page where you can type your message before – like [this](https://fitti.io/twitch) – would be a great addition I think!

PR #302, opened 2021-08-16
Swapping Quart for FastAPI

arcbtc:
FastAPI is much better supported/battle-tested, auto-generates docs, and will make our codebase leaner 
https://amitness.com/2020/06/fastapi-vs-flask/#swagger-documentation


arcbtc:
FastApi also has nice validation, so I think we can do away with our own validation https://github.com/lnbits/lnbits/blob/afc2b78a25b96600b68deb2a6cdf830508dd7d66/lnbits/decorators.py#L35

cryptoteun:
@arcbtc @talvasconcelos are there any extensions or core components already stable enough for testing?

PR #326, opened 2021-08-31
create wallet keys index

talvasconcelos:
Attempting to reduce the DB load on lndhub calls

Created the index for the wallet keys on migrations!

@arcbtc @fiatjaf


arcbtc:
@callebtc fixed here https://github.com/lnbits/lnbits-legend/pull/499

Issue #336, opened 2021-09-11
Payments from same LNbits instance are not honored in function check_invoice_status


zerofeerouting:
# To reproduce

On a new LNBits installation, create a wallet, activate the `TPoS` extension and create a TPoS. Open it and create an arbitrary charge. If that charge gets paid from an external node everything works as expected, the QR code vanishes and a  `x sat received` success message is displayed . If the charge get's paid from another wallet on the same LNbits instance, the `awaiting payment` message and the QR code never go away (although the payment is already completed).

# Cause

The function `check_invoice_status` only checks the funding source (i.e. lnd, c-lightning etc.) if the payment was made, not the internal `payment.status`.

# Proposed Fix

We add

```
if not payment.pending:
    return PaymentStatus(True)
```
on line 320 in `core/services.py` - so that the invoice status is also returned as `True` if an internal payment was made (i.e. the payment is not pending anymore according to the internal db).

I will prepare a PR

PR #337, opened 2021-09-11
honor lnbits payments in check_invoice_status #336

zerofeerouting:
Hope to get this merged so that TPoS and Charges work as expected for internal payments.


arcbtc:
Would this not mean failed payments clear as paid?

talvasconcelos:
@arcbtc from what i can see, please double check it, it only gets set as not pending if it's paid.

```
@property
    def pending(self) -> bool:
        return self.paid is not True
```

zerofeerouting:
Thank you for the feedback. I understood it the same way @talvasconcelos does. Happy to help/improve.

PR #346, opened 2021-09-23
Add actual fee limit (only lnd for now)

Fittiboy:
This sets a hard fee limit for lnd backends. It should entirely prevent negative LNbits wallet balances (obviously for lnd only for now).
Additionally, this makes it possible to sweep the wallet's entire balance, if a 0-fee route can be found.

I haven't done all too extensive testing for this, so it's just a draft PR for now. It would be nice someone could test this more thoroughly, as I probably missed things.

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -158,7 +158,15 @@ async def pay_invoice(
             await internal_invoice_paid.send(internal_checking_id)
         else:
             # actually pay the external invoice
-            payment: PaymentResponse = await WALLET.pay_invoice(payment_request)
+            fee_limit_msat = fee_reserve(invoice.amount_msat)
+            try:
+                payment: PaymentResponse = await WALLET.pay_invoice(
+                    payment_request,
+                    fee_limit_msat=fee_limit_msat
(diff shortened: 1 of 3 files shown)


fiatjaf:
A temporary payment is created at https://github.com/lnbits/lnbits/pull/346/files#diff-0f59f07ce344c28b48f1e3ef2ee36b4428a8a7e9e66ffc53a4ea4cdebff29725R137 before the balances are checked to prevent race conditions (two payments being attempted at the same time). Relying on `old_balance` here is problematic because it bypasses that.

The way to do it would be to set the fee reserve dynamically and use that as LND's fee_limit. But I think for now you could just hardcoded the fee_limit on the LND wallet file to reflect the hardcoded fee reserve and see if that fixes it.

Then we can get fancy later.

fiatjaf:
Eclair and c-lightning also have ways to set fee limits. We'll have to look at these.

For the others maybe we should expose a constant saying how each one behaves regarding fees and use that for calculating the fee reserve.

callebtc:
Closed #447

PR #354, opened 2021-10-05
Display balance with local currency.

jpentland:
This adds a local fiat currency selection dropdown to the top left corner. When any currency other than sats is selected, the balance will be displayed in the local currency below the sats balance. The transaction table also gets a new column containing the converted values for each TX.

Currently the wallet list on the left does not get a conversion since I am not sure yet how this should look.

Summary of current status:

* New API path added: 'api/v1/rates' to get the exchange rate for various currencies
    * This currently requires a valid admin key, in order to prevent non-account holders from spamming the API
    *  Makes use of functions in exchange_rates.py
* base.js contains additional methods to handle local currency selection and calling the new rates API
    * Local currency selection stored in local storage
* base.html has the drop-down added
* wallet.js has a new method formattedValueLocal, which converts a sats value to a local currency, based on rate and local currency
* When creating a new invoice, the selected local currency is automatically selected as default
* When paying an invoice, the invoice value is also displayed in the local currency

Possible improvements needed:
* ✔️ Regular update of rate based on some kind of timer to keep exchange rate up to date when not refreshing the page
* Conversion of wallet balances on left hand menu
* Some extensions could also make use of this. The tpos can make use of the new API path to get exchange rates.
* Some tidy up of the code
* ✔️  only show when a .env variable is enabled
* Add API docs

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,7 @@ PORT=5000
 
 LNBITS_ALLOWED_USERS=""
 LNBITS_DEFAULT_WALLET_NAME="LNbits wallet"
+LNBITS_SHOW_FIAT=true
 
 # Database: to use SQLite, specify LNBITS_DATA_FOLDER
 #           to use PostgreSQL, specify LNBITS_DATABASE_URL=postgres://...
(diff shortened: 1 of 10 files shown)


jpentland:
Don't show dropdown when no wallet is open (api depends on wallet admin key).

jpentland:
Reworked api to use /rates/from/to instead of rates?from=&to= and tidied code in exchange_rates.py

jpentland:
Closing this because there will a new PR with the fastapi branch after fastapi is merged.

PR #355, opened 2021-10-06
adds docker-compose for postgres database

tomichec:


dni:
there is also this duplication https://github.com/lnbits/lnbits-legend/pull/355

PR #368, opened 2021-10-28
Return a valid getinfo JSON response in the lndhub extension

bumi:
The lndhub extension currently returns a "bad auth" error on the /getinfo call.

Ideally we respond with the exact same JSON as LndHub, but we might not have all that information. 
Still we should at least return a valid response here.

### DIFF
--- a/lnbits/extensions/lndhub/views_api.py
+++ b/lnbits/extensions/lndhub/views_api.py
@@ -15,7 +15,7 @@ from .utils import to_buffer, decoded_as_lndhub
 
 @lndhub_ext.route("/ext/getinfo", methods=["GET"])
 async def lndhub_getinfo():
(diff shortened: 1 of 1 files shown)


bumi:
Any chance we can get this in? I think returning a "bad auth" error really does not make sense there.

callebtc:
I think this PR should be merged. Curiously, BlueWallet itself doesn't care about lndhub's own specifications, it seems. Which is probably why this hasn't come up earlier :)

Same fix is needed in the FastAPI branch, btw.

dennisreimann:
This patch somehow went missing. The API is responding with `bad auth` everytime again, see https://github.com/lnbits/lnbits/blob/main/lnbits/extensions/lndhub/views_api.py#L22

dennisreimann:
Filed here: #1182

PR #413, opened 2021-11-12
LndRestWallet fix: add fees to output

callebtc:
Only lines 113 and 115 are changes, the rest is the formatter.

### DIFF
--- a/lnbits/wallets/lndrest.py
+++ b/lnbits/wallets/lndrest.py
@@ -21,7 +21,8 @@ class LndRestWallet(Wallet):
         endpoint = getenv("LND_REST_ENDPOINT")
         endpoint = endpoint[:-1] if endpoint.endswith("/") else endpoint
         endpoint = (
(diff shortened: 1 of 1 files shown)


fusion44:
LGTM - thanks!

Tiny nitpick: I would be not commit the formatting changes as they make it a bit harder to digest the real changes.

callebtc:
> Tiny nitpick: I would be not commit the formatting changes as they make it a bit harder to digest the real changes.

Yes, for sure, I'm using black, but my line length is maybe not the same as your's. What's the convention?

fusion44:
Not sure, I think they are black defaults. Maybe @arcbtc can elaborate?

In cases like this I usually only stage the ranges I actually want to commit.

callebtc:
Black default is 88 characters, I've been using 120 and it seems that you're using even more than that :)

callebtc:
Who does the merges?

fusion44 (approved):
LGTM!

PR #417, opened 2021-11-14
Unit tests for FastAPI branch

chill117:
Install test dependencies as follows:
```bash
./venv/bin/pip install pytest pytest-asyncio requests trio mock
```

And to run the tests:
```bash
make test
```

### DIFF
--- a/tests/core/views/test_generic.py
+++ b/tests/core/views/test_generic.py
@@ -1,0 +1,7 @@
+import pytest
+from tests.conftest import client
+
+@pytest.mark.asyncio
+async def test_core_views_generic(client):
+    response = await client.get("/")
+    assert response.status_code == 200
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,11 +3,11 @@ name: tests
 on: [push, pull_request]
 
 jobs:
-  build:
+  unit:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8]
+        python-version: [3.8]
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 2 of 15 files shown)


chill117:
Fixed merge conflicts, ready for review.

callebtc:
First of all, it's great to see a PR for improving CI and testing. 

If I'm not mistaken though, you've added a new file ` lnbits/wallets/dummy.py` for testing purposes only. Instead, this should ideally be handled via a [mock](https://docs.python.org/3/library/unittest.mock.html) wallet instead of adding one to the main project directory.

chill117:
> If I'm not mistaken though, you've added a new file ` lnbits/wallets/dummy.py` for testing purposes only. Should this not be handled via a [mock](https://docs.python.org/3/library/unittest.mock.html) wallet, instead of adding one to the main project directory?

Thanks for taking a look! Yes, I added a dummy wallet class for testing purposes only. It was necessary to fully test the bleskomat extension and could also later be used to test other extensions or core lnbits functions.

This is the first python project to which I've added automated tests, so I am not fully aware of the best practices in the python ecosystem. Your suggestion - to use python's [unittest.mock](https://docs.python.org/3/library/unittest.mock.html) looks like the right choice. I will make the change and remove the dummy wallet class.

chill117:
@callebtc I removed the dummy wallet class. Ready for another review.

fusion44 (approved):
Looks good to me!

chill117:
@callebtc Do you have some time to have a look at this one once again?

callebtc:
Hey, thanks for pinging me. For sure, a few critical things came up which is why I couldn't.

However, I've used your code to start writing my own tests as well! So thanks for that.

If I'm correct, you've added a few environment things (Dockerfile.dev etc) to this PR as well. Ideally, we should avoid that. I will go through them and comment in the code!

Cheers

tests/core/views/test_generic.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+from tests.conftest import client
+
+@pytest.mark.asyncio
+async def test_core_views_generic(client):
+    response = await client.get("/")
+    assert response.status_code == 200
+    assert "Add a new wallet" in response.text
```
callebtc, line 8: I think we should remove these string tests and just test for http code. Later, we we'll add more tests here.

callebtc (commented):
Lgtm apart from comments, maybe get rid of Dockerfile.env? What do you use it for?

tests/core/views/test_generic.py
chill117: Sure, I'll remove the string assertions.

chill117:
> If I'm correct, you've added a few environment things (`Dockerfile.dev` etc) to this PR as well. Ideally, we should avoid that. I will go through them and comment in the code!
> 
> Cheers
> 
> Edit: I was on mobile and missed your top comment for using the `Dockerfile.dev`. What about not including it?

I use that to create a docker image specifically for testing. A few python dependencies are needed only for testing. Would you prefer to add these new dev-only dependencies somewhere else?

callebtc:
> Would you prefer to add these new dev-only dependencies somewhere else? 

I would actually suggest putting these into a `.github/workflows/test_core.yaml` `.github/workflows/test_extensions.yaml` file that installs everything necessary and runs the tests. What do you think?

chill117:
> > Would you prefer to add these new dev-only dependencies somewhere else?
> 
> I would actually suggest putting these into two `.github/workflows/test_core.yaml` and `.github/workflows/test_extensions.yaml` file that installs everything necessary and runs the tests on github. What do you think?

I cannot run lnbits locally because of sqlite version requirements. So that's why I use docker to run the tests locally. I think it should be possible for developers to test on their local machine - not just GitHub.

callebtc:
> I think it should be possible for developers to test on their local machine - not just GitHub. 

Yes, for sure! That's not connected to this issue. 

Starting a test is entirely possible through the Makefile. It's just that you have pushed _your_ Dockerfile.dev. It doesn't seem to be necessary for tests, therefore I doubt that it should be in this PR. Most devs have a different way of setting up and starting their lnbits instance. I don't use docker for example.

chill117:
> > I think it should be possible for developers to test on their local machine - not just GitHub.
> 
> Yes, for sure! That's not connected to this issue.
> 
> Starting a test is entirely possible through the Makefile. It's just that you have pushed _your_ Dockerfile.dev. It doesn't seem to be necessary for tests, therefore I doubt that it should be in this PR. Most devs have a different way of setting up and starting their lnbits instance. I don't use docker for example.

But as it is now, if someone were to clone the lnbits repository and run the installation steps. They cannot run the tests because they would be missing the necessary dependencies.

chill117:
@callebtc I have removed the Dockerfile.dev and added a short section to [docs/devs/development.md](https://github.com/lnbits/lnbits-legend/blob/FastAPI/docs/devs/development.md) about how to run the tests locally. I also removed the string assertions in the one core test.

callebtc:
I've modified the commands to use the `venv`, not the system python.

PR #437, opened 2021-11-27
Fix UserManager endpoint collision (master)

callebtc:
* Renamed endpoint to `/api/v1/wallets/<wallet_id>`  `/api/v1/transactions/<wallet_id>` to avoid collision with `/api/v1/wallets/<user_id>`.


fusion44:
Looks good to me.

PR #438, opened 2021-11-27
Fix UserManager endpoint collision (FastAPI)

callebtc:
* Added annotations for optional attributes of `User(BaseModel)` to avoid these errors, which caused users to not show up in the UserManager user list. 

```
pydantic.error_wrappers.ValidationError: 2 validation errors for Users
email
  none is not an allowed value (type=type_error.none.not_allowed)
password
  none is not an allowed value (type=type_error.none.not_allowed)
```

* Renamed endpoint to `/api/v1/wallets/<wallet_id>`  `/api/v1/transactions/<wallet_id>` to avoid collision with `/api/v1/wallets/<user_id>`.


fusion44:
Looks good to me.

PR #442, opened 2021-11-29
LNDRest fee limit

callebtc:
Adds `fee_limit` to the the payment.

Policy: 
```python
if invoice.amount_msat > 1000_000:
    lnrpcFeeLimit["percent"] = "1"  # in percent
else:
    lnrpcFeeLimit["fixed"] = "10"  # in sat
```

What changed?
1) Add `fee_limit`. 
2) When payment fails, delete temporary DB entry. Why didn't we do this before? 
3) `check_invoice_status` in `api.py` (`/api/v1/payments`) now `exclude_uncheckable=True`. This excludes DB entries with the `temp_` prefix, that can't be checked (and probably should not be checked?).

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -173,6 +173,8 @@ async def pay_invoice(
                 await delete_payment(temp_id, conn=conn)
         else:
             print("---- ERROR CAUGHT ----")
+            async with db.connect() as conn:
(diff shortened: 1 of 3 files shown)


fiatjaf:
Looks good to me, but please squash before merging.

PR #447, opened 2021-12-02
United fee limit

iWarpBTC:
I was challenged by callebyc so here is my very first PR ever.

- The fee limit is now determined from `fee_reserve()` and passed to WALLET.pay_invoice(), to ensure maintaining fee limit = fee reserve (otherwise it may cause a vulnerability).

- This fee limit is now implemented in `LndRestWallet`, `LndWallet` and `CLightningWallet` (here the 0.5% default fee limit is being overwritten).

Thanks for your feedback and to Kixunil for his guidance.

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -159,7 +159,7 @@ async def pay_invoice(
         await internal_invoice_paid.send(internal_checking_id)
     else:
         # actually pay the external invoice
-        payment: PaymentResponse = await WALLET.pay_invoice(payment_request)
+        payment: PaymentResponse = await WALLET.pay_invoice(payment_request, fee_reserve(invoice.amount_msat))
         if payment.checking_id:
             async with db.connect() as conn:
                 await create_payment(
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -84,11 +84,13 @@ async def pay_invoice(
     description: str = "",
     conn: Optional[Connection] = None,
 ) -> str:
+    invoice = bolt11.decode(payment_request)
+    fee_reserve_msat = fee_reserve(invoice.amount_msat)
+
     async with (db.reuse_conn(conn) if conn else db.connect()) as conn:
         temp_id = f"temp_{urlsafe_short_hash()}"
         internal_id = f"internal_{urlsafe_short_hash()}"
 
-        invoice = bolt11.decode(payment_request)
         if invoice.amount_msat == 0:
             raise ValueError("Amountless invoices not supported.")
         if max_sat and invoice.amount_msat > max_sat * 1000:
@@ -131,7 +133,7 @@ async def pay_invoice(
             # the balance is enough in the next step
             await create_payment(
                 checking_id=temp_id,
-                fee=-fee_reserve(invoice.amount_msat),
+                fee=-fee_reserve_msat,
                 conn=conn,
                 **payment_kwargs,
             )
@@ -159,7 +161,7 @@ async def pay_invoice(
         await internal_invoice_paid.send(internal_checking_id)
     else:
         # actually pay the external invoice
-        payment: PaymentResponse = await WALLET.pay_invoice(payment_request, fee_reserve(invoice.amount_msat))
+        payment: PaymentResponse = await WALLET.pay_invoice(payment_request, fee_reserve_msat)
         if payment.checking_id:
             async with db.connect() as conn:
                 await create_payment(
@@ -339,5 +341,6 @@ async def check_invoice_status(
     return status
 
 
+# WARN: this same value must be used for balance check and passed to WALLET.pay_invoice(), it may cause a vulnerability if the values differ
 def fee_reserve(amount_msat: int) -> int:
     return max(1000, int(amount_msat * 0.01))
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -343,4 +343,4 @@ async def check_invoice_status(
 
 # WARN: this same value must be used for balance check and passed to WALLET.pay_invoice(), it may cause a vulnerability if the values differ
 def fee_reserve(amount_msat: int) -> int:
-    return max(1000, int(amount_msat * 0.01))
+    return max(2000, int(amount_msat * 0.01))
--- a/lnbits/wallets/base.py
+++ b/lnbits/wallets/base.py
@@ -59,6 +59,8 @@ class Wallet(ABC):
     ) -> Coroutine[None, None, InvoiceResponse]:
         pass
 
+    # WARNING: correct handling of fee_limit_msat is required to avoid security vulnerabilities!
+    # The backend MUST NOT spend satoshis above invoice amount + fee_limit_msat.
     @abstractmethod
     def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> Coroutine[None, None, PaymentResponse]:
         pass
--- a/lnbits/wallets/base.py
+++ b/lnbits/wallets/base.py
@@ -60,7 +60,7 @@ class Wallet(ABC):
         pass
 
     @abstractmethod
-    def pay_invoice(self, bolt11: str) -> Coroutine[None, None, PaymentResponse]:
+    def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> Coroutine[None, None, PaymentResponse]:
         pass
 
     @abstractmethod
--- a/lnbits/wallets/clightning.py
+++ b/lnbits/wallets/clightning.py
@@ -88,7 +88,8 @@ class CLightningWallet(Wallet):
 
     async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse:
         invoice = lnbits_bolt11.decode(bolt11)
-        fee_limit_percent = (invoice.amount_msat // fee_limit_msat) // 100
+       fee_limit_percent = fee_limit_msat / invoice.amount_msat * 100
+
 
         payload = {
             "bolt11" : bolt11,
--- a/lnbits/wallets/clightning.py
+++ b/lnbits/wallets/clightning.py
@@ -84,7 +84,7 @@ class CLightningWallet(Wallet):
             error_message = f"lightningd '{exc.method}' failed with '{exc.error}'."
             return InvoiceResponse(False, label, None, error_message)
 
-    async def pay_invoice(self, bolt11: str) -> PaymentResponse:
+    async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse:
         try:
             r = self.ln.pay(bolt11)
         except RpcError as exc:
--- a/lnbits/wallets/clightning.py
+++ b/lnbits/wallets/clightning.py
@@ -10,6 +10,8 @@ import json
 from os import getenv
 from typing import Optional, AsyncGenerator
 
+from lnbits import bolt11 as lnbits_bolt11
+
 from .base import (
     StatusResponse,
     InvoiceResponse,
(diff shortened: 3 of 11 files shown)


Kixunil (approved):
Looks good!

In the future it'd be nice to implement support for other backends. Not sure if it's relevant for all of them. At least the most popular should be fixed by this.

lnbits/wallets/lnpay.py
```diff
@@ -76,7 +76,7 @@ async def create_invoice(
 
         return InvoiceResponse(ok, checking_id, payment_request, error_message)
 
-    async def pay_invoice(self, bolt11: str) -> PaymentResponse:
+    async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse:
```
callebtc, line 79: What is we make `fee_limit_msat` an optional argument in the other functions, so we don't need it when it can't be used. Then, we could also still use `pay_invoice(bolt11)` like the before.

lnbits/wallets/clightning.py
```diff
@@ -84,9 +86,17 @@ async def create_invoice(
             error_message = f"lightningd '{exc.method}' failed with '{exc.error}'."
             return InvoiceResponse(False, label, None, error_message)
 
-    async def pay_invoice(self, bolt11: str) -> PaymentResponse:
+    async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse:
+        invoice = lnbits_bolt11.decode(bolt11)
+        fee_limit_percent = (invoice.amount_msat // fee_limit_msat) // 100
```
callebtc, line 91: This looks inverted. Also: why use `//`? `maxfeepercent`'s default is itself a `float` (0.5).

```suggestion
       fee_limit_percent = fee_limit_msat / invoice.amount_msat * 100

```

lnbits/wallets/clightning.py
```diff
@@ -84,9 +86,17 @@ async def create_invoice(
...
+
+        payload = {
+            "bolt11" : bolt11,
+            "maxfeepercent" : fee_limit_percent
+        }
+        
         try:
-            r = self.ln.pay(bolt11)
+            r = self.ln.call("pay", payload)
```
callebtc, line 99: I can't test this :) Did you?

callebtc:
Thanks a lot for the PR. Very cool that you have implemented it in the LndRPC and c-lightning backends!

What happens if `fee_reserve` is set to 0? I am thinking that we could still have a max-fee policy in that case, maybe we should use a separate function to calculate the max fee (which should always be at least some amount)? What do you think?

lnbits/wallets/clightning.py
iWarpBTC: I tested it only on Polar with c-lightning v0.10.0 and pylightning 0.0.7.3. I've tested it again right now with your inverted fee calc.

lnbits/wallets/clightning.py
iWarpBTC: Much better!

lnbits/wallets/lnpay.py
Kixunil: I don't believe it should be optional because it's absolutely required to be processed for security. Other functions not using it is actually a security vulnerability. It should be addressed in followup PRs or documented.

callebtc: Another option would be to leave out this argument completely and compute the max_fee inside each `WALLET.pay_invoice`. That way, the user / programmer of extensions / etc. cannot mess it up by setting it wrong or using `0` because they don't want to think about it.

Kixunil:
> What happens if `fee_reserve` is set to 0?

In principle it should forbid any fee (thus only allow free transactions). In practice I'm worried that the ridiculous design of Golang to set variables to zero by default could cause zero to mean "unlimited" which would be very wrong. If I guess right then `fee_reserve` should enforce minimum reserve 1 sat or somehow make sure LND enforces the fee.

lnbits/wallets/lnpay.py
Kixunil: > compute the `max_fee_msat` inside each `WALLET.pay_invoice`

That'd be very dangerous because if the computed fees disagree with what's being checked (e.g. after refactor) the vulnerability is back again. This is the very reason I recommended to do the refactoring in this PR.

>  the user / programmer of extensions / etc. cannot mess it up by setting it wrong or using `0` because they don't want to think about it.

It should be clearly documented that doing this == easily-exploitable security vulnerability. Sadly, Python doesn't warn about unused variables, so there's a limit on what can be done about it but hopefully it'll be still visible. Perhaps rename the variable to `must_use_fee_limit_msat`.

callebtc: > This is the very reason I recommended to do the refactoring in this PR.

Could you elaborate on this? What refactor are you referring to? I'm out of the loop.

callebtc:
Just to reiterate explicitly: for all payments up to 100 sat, the max_fee will be 1 sat. Above that, it will be 1%. Is this a smart policy? 

I know there are payments below 100 sat that require 1-2 sat fees. These would not go through anymore. Maybe we should think about increasing the min value to 2 sat.

lnbits/wallets/lnpay.py
Kixunil: Any followup code change that could possibly happen in the future. I'm speaking from experience - I did introduce various bugs in my code multiple times when I didn't uphold so called "single source of truth" - having exactly one place to compute a value and distribute it using variables. In this case it's even bigger risk because we know for a fact that it can lead to serious security vulnerability.

iWarpBTC:
I've tried zero fee_reserve with lnd and got no_route with a hop involved, but direct payment passes through. So says the API Ref:

> If this field is left to the default value of 0, only zero-fee routes will be considered. This usually means single hop routes connecting directly to the destination.

Kixunil:
How about exposing the limit to the user and allow setting it? There can still be a reasonable default (e.g. what you suggest).

Kixunil:
@iWarpBTC that's great news, no danger there!

callebtc:
Do you guys agree that we should increase the minimal value to 2 sat? I'm open to reasons against it.

Kixunil:
I agree. Would be cool to expose the limit later. Actually, it's required for lnbits backend to be secure.

iWarpBTC:
I don't see any reasons against it...

lnbits/core/services.py
```diff
@@ -339,5 +341,6 @@ async def check_invoice_status(
     return status
 
 
+# WARN: this same value must be used for balance check and passed to WALLET.pay_invoice(), it may cause a vulnerability if the values differ
 def fee_reserve(amount_msat: int) -> int:
     return max(1000, int(amount_msat * 0.01))
```
callebtc, line 346:
```suggestion
    return max(2000, int(amount_msat * 0.01))
```

callebtc: Here we go, I checked with the others, they are fine with it as well.

lnbits/wallets/clightning.py
```diff
@@ -84,9 +86,17 @@ async def create_invoice(
...
             return InvoiceResponse(False, label, None, error_message)
 
-    async def pay_invoice(self, bolt11: str) -> PaymentResponse:
+    async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse:
+        invoice = lnbits_bolt11.decode(bolt11)
+        fee_limit_percent = fee_limit_msat / invoice.amount_msat * 100
+
+        payload = {
+            "bolt11" : bolt11,
+            "maxfeepercent" : fee_limit_percent
```
callebtc, line 95: Doesn't this need to be a string? Precision might cause issues.
```suggestion
            "maxfeepercent" : "{:.1f}".format(fee_limit_percent)
```

lnbits/wallets/lnpay.py
callebtc: Ok, I agree!

lnbits/wallets/clightning.py
iWarpBTC: I did some tests with various precision (out of curiosity) and didn't induce any issue. But since fee_limit_percent is a float now, maybe it's better to have control over the precision.

lnbits/wallets/base.py
```diff
@@ -60,7 +60,7 @@ def create_invoice(
         pass
 
     @abstractmethod
-    def pay_invoice(self, bolt11: str) -> Coroutine[None, None, PaymentResponse]:
+    def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> Coroutine[None, None, PaymentResponse]:
```
Kixunil, line 63: I suggest adding this comment above `def pay_invoice`:

```python
# WARNING: correct handling of fee_limit_msat is required to avoid security vulnerabilities!
# The backend MUST NOT spend satoshis above invoice amount + fee_limit_msat.
```

iWarpBTC:
With clightning we should also set the `exemptfee` to 0 (default is 5000 msat) in order to ensure the backend never spends above the lnbits wallet balance. Agree?

[https://lightning.readthedocs.io/lightning-pay.7.html](url)

lnbits/wallets/clightning.py
iWarpBTC: Maybe I would also revert this change, it seems to me that clightning handles well many digits precision... Bad idea?

Snímek obrazovky pořízený 2021-12-03 10-33-45

Kixunil:
Not familiar with C-lightning but from [the documentation](https://github.com/ElementsProject/lightning/blob/master/doc/lightning-pay.7.md#description) it seems to be the case.

lnbits/wallets/lndgrpc.py
```diff
@@ -144,8 +144,12 @@ async def create_invoice(
         payment_request = str(resp.payment_request)
         return InvoiceResponse(True, checking_id, payment_request, None)
 
-    async def pay_invoice(self, bolt11: str) -> PaymentResponse:
-        resp = self.rpc.send_payment(payment_request=bolt11)
+    # WARNING: correct handling of fee_limit_msat is required to avoid security vulnerabilities!
+    # The backend MUST NOT spend satoshis above invoice amount + fee_limit_msat.
+    async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse:
+        fee_limit_fixed = ln.FeeLimit(fixed=fee_limit_msat//1000)
```
arcbtc, line 150: Should this be fixed=fee_limit_msat/1000

Kixunil: Actually, LND supports msats: https://api.lightning.community/#lnrpc-feelimit
So `fee_limit_fixed = ln.FeeLimit(fixed_msat=fee_limit_msat)` should work.

Kixunil:
@iWarpBTC got confirmation from Rusty that indeed it needs to be set to 0.

lnbits/wallets/lndgrpc.py
iWarpBTC: I am afraid [the library](https://github.com/adrienemery/lnd-grpc-client/blob/master/lndgrpc/rpc_pb2.py#L501) is not ready for this. It needs to be a satoshi integer... But please correct me if I am being wrong.

Kixunil: Looks like that library is not updated. So I'd suggest use integer now and try to look for a better maintained library.

lnbits/wallets/clightning.py
iWarpBTC: Or `"maxfeepercent" : "{:.11f}".format(fee_limit_percent)` to be on the safe side?

Kixunil (approved):
Looks good now.

lnbits/wallets/clightning.py
callebtc: We should set the precision ourselves, the more precise the better. 11 digits seems high, but why not. If you've tested it (I can't), it should be fine :)

callebtc:
> With clightning we should also set the `exemptfee` to 0 (default is 5000 msat) in order to ensure the backend never spends above the lnbits wallet balance. Agree?

Agree.

lnbits/wallets/lndgrpc.py
callebtc: Yes, needs integer, should be `//`.

callebtc:
After precision is set, I will have one more look at all changes and we're ready to merge.

lnbits/wallets/lndgrpc.py
callebtc: > 

We're already switching to another library (see `fastapi` branch).

PR #450, opened 2021-12-04
New extension: Autopay - scheduled LNURL payments

crowphale:
**What is this extension?**
You can schedule repeated payments in the future, using LNURL (type
payRequest): add LNURL, with amount, future datetime and repeating frequency (hourly, daily, weekly), and LNBits will execute the payments.

**Feedback for this PR - potential issues:**
- Currently the scheduling is done by spawning a process that hits the webserver with curl. This is not great, likely doesn't work with Tor. I tried really hard to make this work with Quart/asyncio (just put fn on the event loop that will get called every x seconds/minutes, but I couldn't figure it out). Once LNBits updates to Quart 0.16, [this](https://stackoverflow.com/questions/70075859/scheduling-periodic-function-call-in-quart-asyncio/70086619#70086619) should work, hopefully.
- I mostly wanted to learn how LNBits work, and wanted this functionality so that I can setup automatic saving/allowance for my kids. I've never used Quart/Trio/Vue/Quasar/etc before so likely it's not following best practices. Please let me know, or even better just send improvements.

End-to-end tested locally with Polar.

### DIFF
--- a/lnbits/extensions/autopay/__init__.py
+++ b/lnbits/extensions/autopay/__init__.py
@@ -1,0 +1,30 @@
+from quart import Blueprint
+from lnbits.db import Database
+
+from .context import AutopayContext
+
+
+db = Database("ext_autopay")
+
+autopay_ext: Blueprint = Blueprint(
+    "autopay", __name__, static_folder="static", template_folder="templates"
+)
+
+context = AutopayContext(db)
+
+
+import sys
+
+if "pytest" not in sys.modules:
+    # Only load these in prod, for some reason it fails under pytest
+    from .views_api import *  # noqa
+    from .views import *  # noqa
+
+    # Spawn a process that periodically hits the scheduler API
+    # TODO: use asyncio once lnbits updates to quart 0.16
+    # See https://stackoverflow.com/questions/70075859/scheduling-periodic-function-call-in-quart-asyncio
+    from lnbits.settings import HOST, PORT
+    import subprocess
+    scheduler_check_freq_secs = '60'
+    p = subprocess.Popen(['watch', '-n', scheduler_check_freq_secs, f"curl -k -X POST https://{HOST}:{PORT}/autopay/api/v1/scheduler"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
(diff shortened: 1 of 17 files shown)


crowphale:
Are there any instructions for how to test this with Umbrel? (Ideally without messing it up)

arcbtc:
Haha, I'm making the exact same extension 😅! 
We'll use yours though, happy to have you abroad.
I was going to use a listener that checks to see the payment has breached the set limit. Feel free to see if there is anything you can use. All the ui stuff loads up, but listeners not working yet.

https://github.com/arcbtc/lnbits/tree/lnurlpayout/lnbits/extensions/lnurlpayout

arcbtc:
Have you looked at the way I started to do the autopayout, using a listener?

arcbtc:
You have also made for Master, and we will be switching to FastAPI branch in a few days. I can help you convert it, or you could take a look at mine and see if you can use that

arcbtc:
I PR'd it so you can decide what you want to do?
https://github.com/lnbits/lnbits-legend/pull/455

crowphale:
Hey @arcbtc, sorry for the delay. I was offline for a while.

crowphale:
@arcbtc By "doing autopayout using listener" you mean using the https://github.com/arcbtc/lnbits/blob/331a3832232f96f360c12c9a640042c33d67f1c7/lnbits/extensions/lnurlpayout/tasks.py#L13 for executing the payments? (Rather than my current hack of spawning a `curl` process)
I guess that could work, but doesn't that call back get called whenever there is some payment to the current LNBits instance? That seems hacky as well, because what I want is to just periodically check the time and whether a payment should be executed, so if nobody is sending payments to this instance of LNBits, then payments won't be executed. I might be wrong.

crowphale:
Btw, I mostly made this to learn more about LNBits, so happy to abandon this PR/extension if lnurlpayout supports this already.

I don't have a lot of time to continuously spend hours on further development, but happy to help here and there. I can look into switching it to FastAPI (looks like good fit for LNBits so +1 for using it), just send me link to your branch and any instructions or what I should keep in mind. Also, if you wanna make further changes to this extension, please go ahead.

arcbtc:
The listener solution is reliant on the payment breaching a threshold, so rather than check periodically, you could just set the threshold very low. We're running a similar hacky check for splitpayments, but I am told its very low on resources, however I will pursue a more elegant solution. 
I'll merge my one into FastAPI, but then also port your one to FastAPI in a few weeks. 
Things like WordPress have different extensions performing similar functionality and users pick/decide, I think its healthy 👍

PR #451, opened 2021-12-04
Add email notification capabilities

fusion44:
This PR adds capabilities to send notification email to a user provided email address.

**Why?**
* user might want to get notified when a subscription must be renewed (e.g. lnaddress)
* warn the user when funds are low in case he has some automatic recurring payments running

**How?**
Email support will be **completely optional** and is **opt-in**. If a user doesn't want to set his email then he doesn't have to. Settings will be stored in the .env file and can be disabled there. When email is disabled, no new extensions are necessary.

We have these possible states:
- Globaly disabled - must be disabled in the UI
- Globally enabled but users will be able to turn this on and off at will

To reduce UI clutter there should only be one way to manage the email address:
image

The users profile page can be a whole new page or just a Modal depending on the amount of settings there are.

TODOS:
- [x] implement send via SMTP 
- [ ] implement send via [SendGrid](https://sendgrid.com)
- [ ] add email notifications to LNbits core where appropriate (ideas welcome!)
- [ ] implement an API to verify email addresses
- [ ] charge an optional fee for this service if enabled for a wallet

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -72,3 +72,16 @@ LNTXBOT_KEY=LNTXBOT_ADMIN_KEY
 # OpenNodeWallet
 OPENNODE_API_ENDPOINT=https://api.opennode.com/
 OPENNODE_KEY=OPENNODE_ADMIN_KEY
+
+# Enable Email services. For busy servers use sendgrid as an email provider. 
+# Using a consumer mail provider with large volumes of mails will get you kicked.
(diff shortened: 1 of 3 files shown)


arcbtc:
Closing for now to clear the deck, but we should reopen at some point, if emails can be done in a way that is optional

PR #454, opened 2021-12-05
Withdraw extension error handling

javad94:
Withdraw extension error handling based on lnurl-rfc


bitkarrot:
To the lnbits team - can we do versioning on APIs, because this pull request has data formats returning with a different format and would break all previous versions

callebtc:
@bitkarrot's comment is true, it might break some API's. @javad94, any particular reason for this PR, except for making the response nicer?

callebtc:
Closing due to no activity.

PR #481, opened 2022-01-02
Extension Subdomains: Update crud.py, critical sql error

dni:
i talked to @grmkris via email, and something fishy is going on with cloudflare api also. im looking into it now.

fix sql error: 

sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedTable)
relation "domain" does not exist
LINE 1: ...omain_name FROM subdomains.subdomain s INNER JOIN domain d O...
[SQL: SELECT s.*, d.domain as domain_name FROM subdomains.subdomain s
INNER JOIN domain d ON (s.domain = d.id) WHERE s.id = %s]


dni:
closed it for now found other things too

dni:
ok everything fine now, :D it fixes the problem as expected!

PR #485, opened 2022-01-06
fix copilot not getting created

talvasconcelos:
copilot wasn't being created. user in create_copilot needed `""`


arcbtc:
You're replacing the <user id> with just a `user` string
The user id should be passed, and if its not should be fixed
https://github.com/lnbits/lnbits-legend/blob/6af8899cd1910a0351562bdc472d380961a38d2e/lnbits/extensions/copilot/views_api.py#L59

talvasconcelos:
Good catch, it wasn't... the all user was getting passed! I only fixed for the DB naming issue!

PR #491, opened 2022-01-13
Upgrade conversion endpoint

talvasconcelos:
small upgrade to conversion endpoint to allow fiat -> sats and sats -> fiat

also allow to pass a list of currencies to get converted

updated data schema:
`{unit: <string defaults to 'sat'>, amount: <float>, to: <string defaults to 'usd'>}`
sats -> fiat
`{"amount": 30000, "to": "eur, gbp"}`
fiat -> sat
`{"amount": 10, "unit": "eur"}`

### DIFF
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -516,12 +516,13 @@ class ConversionData(BaseModel):
 async def api_fiat_as_sats(data: ConversionData):
     output = {}
     if data.unit == 'sat':
-        output["sats"] = data.amount
+        output["sats"] = int(data.amount)
         output["BTC"] = data.amount / 100000000
         for currency in data.to.split(','):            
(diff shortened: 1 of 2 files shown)


lnbits/core/views/api.py
```diff
@@ -504,9 +508,21 @@ async def api_list_currencies_available():
 
 
 class ConversionData(BaseModel):
-    unit: str
+    unit: str = Query('sat')
     amount: float
+    to: str = Query('usd')
```
fiatjaf, line 513: can we have `from` and `to` just to be sane? for backwards compatibility is it possible to default `from` to `unit`?

talvasconcelos: fixed, i think this is not being used so no backward compatibilty needed! 
`from` is a reserved keyword, I've set an alias on pydantic model.

PR #498, opened 2022-01-25
fix livestream startup issue

pseudozach:
fixed some small bugs in livestream extension that was throwing errors when I downloaded a fresh clone of lnbits and tried to launch it.


callebtc:
thank you, this is needed!

PR #499, opened 2022-01-27
LNDHhub for fastapi with LIMIT

callebtc:
**Description**
* This PR adds a `limit` clause to the `get_payments` crud in order to ... limit the number of payments to receive from the db. 
* This is necessary since the `lndhub` extension used to requests all payments of a user and just sliced returned list (very inefficient). 
* The default number of tx's to receive is set from 200 to 20, because more is usually not needed and the client can still request more if they like. 
* It also removes the unnecessary `delete_expired_invoices()` which went through _all_ invoices and checked with the backend previously. 

**Comments:**
* Add pagination to the crud as well? This would certainly come handy for the web interface which does not need to fetch all transactions if they are not shown anyway. 

**Todo:**
* [x] Testing
* [x] Check whether `get_payments` on `fastapi` still works with lndhub since it expected a `list` on quart but now gets a `dict`. I had some issues before. 

**Info:** Ready to merge!

### DIFF
--- a/lnbits/extensions/lndhub/views_api.py
+++ b/lnbits/extensions/lndhub/views_api.py
@@ -13,7 +13,7 @@ from lnbits import bolt11
 from lnbits.core.crud import delete_expired_invoices, get_payments
 from lnbits.core.services import create_invoice, pay_invoice
 from lnbits.decorators import WalletTypeInfo
-from lnbits.settings import WALLET
+from lnbits.settings import WALLET, LNBITS_SITE_TITLE
 
 from . import lndhub_ext
 from .decorators import check_wallet, require_admin_key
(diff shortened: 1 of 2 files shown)


arcbtc:
How we doing?

lnbits/extensions/lndhub/views_api.py
```diff
@@ -116,19 +117,25 @@ async def lndhub_balance(
...
         outgoing=True,
         incoming=False,
+        limit=limit,
+        offset=offset,
         exclude_uncheckable=True,
     ):
         await payment.set_pending(
             (await WALLET.get_payment_status(payment.checking_id)).pending
         )
+        await asyncio.sleep(0.1)
```
JssDWt, line 137: Hey @callebtc, I noticed this line. Does that mean it'll take 2 seconds to get 20 payments from LndHub? Is there a reason to keep this?

callebtc: Yes, I added this to prevent DoS in an earlier version but it could be omitted now.

PR #501, opened 2022-01-28
UI is working

talvasconcelos:
Things won't work as there's some differences between migrations.py and models.py


arcbtc:
Ace, will start working into this again, Thanks for converting

PR #505, opened 2022-02-01
fix scan/paste LNURLw in lnbits wallet

talvasconcelos:


arcbtc:
Thanks. Used part of this fix

PR #506, opened 2022-02-02
fix reducing balance for non fake invoices

talvasconcelos:


arcbtc:
Thanks, Used part of this fix

PR #516, opened 2022-02-10
api: Fix conversion from fiat to satoshis

jpentland:
Conversion to satoshis incorrectly used data.to as fiat currency rather
than data._from.

Note: The api currently assumes you are either converting to or from satoshis. This could be fixed to allow arbitrary conversions, so I will add that if needed.


jpentland:
I noticed this PR already contains a fix for this: https://github.com/lnbits/lnbits-legend/pull/495

callebtc:
Thanks for noticing this and also finding the duplicate (which is closed now). Merging.

PR #519, opened 2022-02-11
fix paywall ext not fetching invoice

talvasconcelos:


arcbtc:
Looks like you're on an old branch

PR #521, opened 2022-02-13
Support for encrypted macaroons

callebtc:
This PR adds support for AES-256 encrypted macaroons using `pycryptodomex`. The encryption scheme is compatible with the Javascript library `CryptoJS/aes`. Only the LND backend is supported.

There is a new directory `lnbits/wallets/macaroon` that holds functions for loading macaroons, determining their encoding, and a new class `AESCipher` for de- and encrypting macaroons.

## How it works

User adds an encrypted version of the macaroon in the `.env` file as `LND_GRPC_MACAROON_ENCRYPTED` or `LND_REST_MACAROON_ENCRYPTED`. If these are set, they will overwrite other unencrypted macaroon settings. Opon start, LNBits will ask for the decryption password and decrypt the macaroon. If the password is wrong, an error is returned. The decrypted macaroon is kept in RAM only.

## How to encrypt your macaroon

You have two options: You can encrypt the macaroon with CryptoJS, or with the script in LNBits. 
In Javascript, use

```js
import AES from "crypto-js/aes.js";
AES.encrypt(macaroon, password).toString()
```

To encrypt your macaroon with LNBits, use 
```bash
./venv/bin/python lnbits/wallets/macaroon/macaroon.py
```
You will be asked for the macaroon and an encryption password. 

## Additional changes
* `load_macaroon` refactored into `lnbits/wallets/macaroon/macaroon.py`
* LNBits now also supports bech64 macaroons 🎉.
* `LNDRestWallet` now supports macaroon files instead of only the hex.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -55,11 +55,15 @@ LND_GRPC_ENDPOINT=127.0.0.1
 LND_GRPC_PORT=11009
 LND_GRPC_CERT="/home/bob/.config/Zap/lnd/bitcoin/mainnet/wallet-1/data/chain/bitcoin/mainnet/tls.cert"
 LND_GRPC_MACAROON="/home/bob/.config/Zap/lnd/bitcoin/mainnet/wallet-1/data/chain/bitcoin/mainnet/admin.macaroon or HEXSTRING"
+# To use an AES-encrypted macaroon, set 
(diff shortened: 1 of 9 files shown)


arcbtc:
Ah, I see, it should be pycryptodomex

PR #522, opened 2022-02-13
Add env var as admin password

AaronDewes:
This adds an environment variable `LNBITS_ADMIN_LOGIN_KEY`. This key can be entered on the wallet creation screen to make your account an admin.

grafik
grafik

The button shown in the screenshot is only visible if `LNBITS_ADMIN_LOGIN_KEY` is set and not an empty string.


arcbtc:
I like it a lot. We'll be working on an optional admin extension UI in the next few weeks, so will be transposing the .envs into a db. I'll make sure to add this feature

PR #524, opened 2022-02-16
Fix fiat amount entry in invoice creation

callebtc:
This PR fixes two fiat unit bugs:

* Units were `int()` so 2.30€ == 2€ but nobody seems to have noticed until now 🤷.
* Amount was necessarily `> 1` irrespective of the unit, which means that `$0.25` invoices failed to be created.

Extra:
* The label of the invoice creation dialog changes based on the selected unit (I am very proud of this)
* Allow the creation of invoices without a memo (which is what the UI hints at). Defaults to "LNbits".
* Input changes to '#.##' if a fiat currency is selected.

Todo:
* [x] Test if the label still works if `LNBITS_DENOMINATIOIN` is actually set!

image

### DIFF
--- a/lnbits/core/templates/core/_api_docs.html
+++ b/lnbits/core/templates/core/_api_docs.html
@@ -61,7 +61,7 @@
         <code
           >curl -X POST {{ request.base_url }}api/v1/payments -d '{"out": false,
           "amount": &lt;int&gt;, "memo": &lt;string&gt;, "webhook":
-          &lt;url:string&gt;}' -H "X-Api-Key: <i>{{ wallet.inkey }}</i>" -H
+          &lt;url:string&gt;, "unit": &lt;string&gt;}' -H "X-Api-Key: <i>{{ wallet.inkey }}</i>" -H
           "Content-type: application/json"</code
         >
       </q-card-section>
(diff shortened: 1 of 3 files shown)


lnbits/core/templates/core/_api_docs.html
```diff
@@ -61,7 +61,7 @@ <h5 class="text-caption q-mt-sm q-mb-none">Curl example</h5>
         <code
           >curl -X POST {{ request.base_url }}api/v1/payments -d '{"out": false,
           "amount": &lt;int&gt;, "memo": &lt;string&gt;, "webhook":
-          &lt;url:string&gt;}' -H "X-Api-Key: <i>{{ wallet.inkey }}</i>" -H
+          &lt;url:string&gt;, "unit": &lt;string&gt;}' -H "X-Api-Key: <i>{{ wallet.inkey }}</i>" -H
```
callebtc, line 64: unit in the api docs

lnbits/core/views/api.py
```diff
@@ -123,7 +123,7 @@ async def api_payments(wallet: WalletTypeInfo = Depends(get_key_type)):
 
 class CreateInvoiceData(BaseModel):
     out: Optional[bool] = True
-    amount: int = Query(None, ge=1)
+    amount: float = Query(None, ge=0)
```
callebtc, line 126: we need a float here and >0 for fiat

PR #531, opened 2022-02-19
Delete expired invoices after one pending check

callebtc:
PR fixes two bugs:

* Check for expired invoices regularly, instead of only upon startup.
* Expired invoices should only be deleted after one pending check. The backend could have been offline for a very long time and expired invoices could have been paid in the mean time.

### DIFF
--- a/lnbits/tasks.py
+++ b/lnbits/tasks.py
@@ -60,6 +60,7 @@ def register_invoice_listener(send_chan: asyncio.Queue):
 
 
 async def webhook_handler():
+    handler = getattr(WALLET, "webhook_listener", None)
     if handler:
         return await handler()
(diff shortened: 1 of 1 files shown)


fiatjaf:
I like the bug fix, but I don't think we should call this regularly. It's a very intensive operation. It fetches all the invoices from the database, parses their bolt11 string and checks the expiration time for each.

arcbtc:
Could we add another `loop.create_task`for it and run every 24hrs? That way it will run on startup as well

callebtc:
Missed the comments, sorry. @fiatjaf what do you think of increasing the interval to 24 hours like @arcbtc said?

callebtc:
Changed it to a one-time expiry check after the first pending check.

PR #532, opened 2022-02-21
Retry backend connection on startup

callebtc:
Retry the connection to the backend every 5 seconds if the connection fails. Before, we used to `sys.exit(4)`.

This is useful for Umbrel users who don't use a proper systemd startup chain like a normal person, which means that lnbits doesn't know when lnd is actually ready (and unlocked) to receive connections.


louneskmt:
Tested and working. Thanks @callebtc!

PR #538, opened 2022-02-22
Fix LNURLw QR code 

talvasconcelos:
the QR encoded string had the URL attached, so it failed on scanning with lnbits!


arcbtc:
Damn, nice spot!

PR #560, opened 2022-03-16
WIP: check if uncheckable to avoid error

callebtc:
Checking the status of an uncheckable (== `internal_` or `temp_`) payment will result in an error. This change will avoid calling the payment source payment check and just return  returns `PaymentStatus(None)`. If called by `core/views/api.py:api_payment()`, this will not change anything.

Question: What is the response of the payment source when trying to check an uncheckable payment?

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -319,7 +319,7 @@ async def check_invoice_status(
     wallet_id: str, payment_hash: str, conn: Optional[Connection] = None
 ) -> PaymentStatus:
     payment = await get_wallet_payment(wallet_id, payment_hash, conn=conn)
-    if not payment:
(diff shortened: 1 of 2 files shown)


lnbits/core/services.py
```diff
@@ -319,7 +319,7 @@ async def check_invoice_status(
     wallet_id: str, payment_hash: str, conn: Optional[Connection] = None
 ) -> PaymentStatus:
     payment = await get_wallet_payment(wallet_id, payment_hash, conn=conn)
-    if not payment:
+    if not payment or payment.is_uncheckable:
```
callebtc, line 322: This prevents calling `WALLET.get_invoice_status` in the next line if the payment is not checkable.

PR #565, opened 2022-03-19
api_payments_create API return error for wrong key

callebtc:
Instead of assuming that the user wants to create an invoice if they use an invoice key, we now check whether invoice creation was actually intended (or they just used the wrong key when trying to _make_ a payment)

### DIFF
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -23,7 +23,8 @@ from lnbits.decorators import (
     WalletInvoiceKeyChecker,
     WalletTypeInfo,
     get_key_type,
-    require_admin_key
+    require_admin_key,
+    require_invoice_key,
 )
 from lnbits.helpers import url_for, urlsafe_short_hash
 from lnbits.requestvars import g
@@ -231,12 +232,9 @@ async def api_payments_pay_invoice(bolt11: str, wallet: Wallet):
     status_code=HTTPStatus.CREATED,
 )
 async def api_payments_create(
-    wallet: WalletTypeInfo = Depends(get_key_type),
+    wallet: WalletTypeInfo = Depends(require_invoice_key),
     invoiceData: CreateInvoiceData = Body(...),
 ):
-    if wallet.wallet_type < 0 or wallet.wallet_type > 2:
-        raise HTTPException(status_code=HTTPStatus.BAD_REQUEST, detail="Key is invalid")
-
     if invoiceData.out is True and wallet.wallet_type == 0:
         if not invoiceData.bolt11:
             raise HTTPException(
@@ -246,8 +244,13 @@ async def api_payments_create(
         return await api_payments_pay_invoice(
             invoiceData.bolt11, wallet.wallet
         )  # admin key
-    # invoice key
-    return await api_payments_create_invoice(invoiceData, wallet.wallet)
+    elif not invoiceData.out:
+        # invoice key
+        return await api_payments_create_invoice(invoiceData, wallet.wallet)
+    else:
+        raise HTTPException(
+            status_code=HTTPStatus.BAD_REQUEST, detail="Key type is invalid"
+        )
 
 
 class CreateLNURLData(BaseModel):
@@ -391,7 +394,11 @@ async def api_payment(payment_hash, X_Api_Key: Optional[str] = Header(None)):
         return {"paid": False}
 
     if wallet and wallet.id == payment.wallet_id:
-            return {"paid": not payment.pending, "preimage": payment.preimage, "details": payment}
+        return {
+            "paid": not payment.pending,
+            "preimage": payment.preimage,
+            "details": payment,
+        }
     return {"paid": not payment.pending, "preimage": payment.preimage}
 
 
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -367,13 +367,8 @@ async def api_payments_sse(
 
 
 @core_app.get("/api/v1/payments/{payment_hash}")
-async def api_payment(payment_hash, X_Api_Key: Optional[str] = Header(None)):
-    wallet = None
-    try:
-        if X_Api_Key.extra:
-            print("No key")
-    except:
-        wallet = await get_wallet_for_key(X_Api_Key)
+async def api_payment(payment_hash, wal: WalletTypeInfo = Depends(require_invoice_key)):
+    wallet = wal.wallet
     payment = await get_standalone_payment(payment_hash)
     await check_invoice_status(payment.wallet_id, payment_hash)
     payment = await get_standalone_payment(payment_hash)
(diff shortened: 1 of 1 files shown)


lnbits/core/views/api.py
```diff
@@ -231,12 +232,9 @@ async def api_payments_pay_invoice(bolt11: str, wallet: Wallet):
     status_code=HTTPStatus.CREATED,
 )
 async def api_payments_create(
-    wallet: WalletTypeInfo = Depends(get_key_type),
+    wallet: WalletTypeInfo = Depends(require_invoice_key),
     invoiceData: CreateInvoiceData = Body(...),
 ):
-    if wallet.wallet_type < 0 or wallet.wallet_type > 2:
-        raise HTTPException(status_code=HTTPStatus.BAD_REQUEST, detail="Key is invalid")
-
```
callebtc, line 239: We can get rid of these lines when using `require_invoice_key`

lnbits/core/views/api.py
```diff
@@ -246,8 +244,13 @@ async def api_payments_create(
...
         )  # admin key
-    # invoice key
-    return await api_payments_create_invoice(invoiceData, wallet.wallet)
+    elif not invoiceData.out:
+        # invoice key
+        return await api_payments_create_invoice(invoiceData, wallet.wallet)
+    else:
+        raise HTTPException(
+            status_code=HTTPStatus.BAD_REQUEST, detail="Key type is invalid"
+        )
```
callebtc, line 253: This is the error for the edge case mentioned in the OP.

fusion44: Nitpick: Maybe it'd be more informative to the user with error message detail="Requires invoice key" instead of detail="Key type is invalid"

lnbits/core/views/api.py
```diff
@@ -246,8 +244,14 @@ async def api_payments_create(
...
-    # invoice key
-    return await api_payments_create_invoice(invoiceData, wallet.wallet)
+    elif not invoiceData.out:
+        # invoice key
+        return await api_payments_create_invoice(invoiceData, wallet.wallet)
+    else:
+        raise HTTPException(
+            status_code=HTTPStatus.BAD_REQUEST,
+            detail="Invoice (or Admin) key required.",
+        )
```
callebtc, line 254: This is the edge case checking, now with more expressive error.

lnbits/core/views/api.py
```diff
@@ -364,13 +368,8 @@ async def api_payments_sse(
...
 @core_app.get("/api/v1/payments/{payment_hash}")
-async def api_payment(payment_hash, X_Api_Key: Optional[str] = Header(None)):
-    wallet = None
-    try:
-        if X_Api_Key.extra:
-            print("No key")
-    except:
-        wallet = await get_wallet_for_key(X_Api_Key)
+async def api_payment(payment_hash, wal: WalletTypeInfo = Depends(require_invoice_key)):
+    wallet = wal.wallet
```
callebtc, line 372: @arcbtc: why not use this? It seems to work according to my tests.

arcbtc: Because it doesn't require a key. You can find out if an invoice is paid or not without a key, or add a key and get all the data back.
ie `curl -X GET https://legend.lnbits.com/api/v1/payments/afb06542db0671c78bc1d16c8e8380c8f1a5238bec74755cd4f91bfd3a39d1a1 `
vs
`curl -X GET https://legend.lnbits.com/api/v1/payments/afb06542db0671c78bc1d16c8e8380c8f1a5238bec74755cd4f91bfd3a39d1a1 -H "X-Api-Key: 8a5e8e06d37540d6b945a0f0593d00f9" -H "Content-type: application/json"`

lnbits/core/views/api.py
arcbtc: Yep good

lnbits/core/views/api.py
arcbtc: Yep, good

arcbtc:
All great apart from the removal of `X_Api_Key: Optional[str] `, thats a useful feature

lnbits/core/views/api.py
callebtc: Gotcha! I've returned it back to its original state with a bit a cleaner code. Is it ok now?

lnbits/core/views/api.py
```diff
@@ -359,19 +363,19 @@ async def api_payments_sse(
     request: Request, wallet: WalletTypeInfo = Depends(get_key_type)
 ):
     return EventSourceResponse(
-        subscribe(request, wallet), ping=20, media_type="text/event-stream"
+        subscribe(request, wallet.wallet), ping=20, media_type="text/event-stream"
```
callebtc, line 366: This was a wrong type annotation.

lnbits/core/views/api.py
```diff
@@ -322,7 +326,7 @@ async def api_payments_pay_lnurl(
 
 
 async def subscribe(request: Request, wallet: Wallet):
-    this_wallet_id = wallet.wallet.id
+    this_wallet_id = wallet.id
```
callebtc, line 329: This was a wrong type annotation.

lnbits/core/views/api.py
```diff
@@ -359,19 +363,19 @@ async def api_payments_sse(
...
-            print("No key")
-    except:
-        wallet = await get_wallet_for_key(X_Api_Key)
+    # We use X_Api_Key here because we want this call to work with and without keys
+    wallet = await get_wallet_for_key(X_Api_Key) if X_Api_Key is not None else None
     payment = await get_standalone_payment(payment_hash)
+    if payment is None:
+        raise HTTPException(
+            status_code=HTTPStatus.NOT_FOUND, detail="Payment does not exist."
+        )
```
callebtc, line 378: for mypy

lnbits/core/views/api.py
```diff
@@ -359,19 +363,19 @@ async def api_payments_sse(
...
 @core_app.get("/api/v1/payments/{payment_hash}")
 async def api_payment(payment_hash, X_Api_Key: Optional[str] = Header(None)):
-    wallet = None
-    try:
-        if X_Api_Key.extra:
-            print("No key")
-    except:
-        wallet = await get_wallet_for_key(X_Api_Key)
+    # We use X_Api_Key here because we want this call to work with and without keys
+    wallet = await get_wallet_for_key(X_Api_Key) if X_Api_Key is not None else None
```
callebtc, line 373: Does exactly what you've described before (@arcbtc) but cleaner. Is there a reason for the `X_Api_Key.extra` check?

fusion44 (approved):
LGTM!

arcbtc:
LGTM also

arcbtc (approved):

PR #569, opened 2022-03-22
Converting AdminUI to FastAPI (WIP)

talvasconcelos:
Still a WIP...

Converting AdminUI extension to FastAPI


talvasconcelos:
AdminUI is working! Only changing the funding wallet, requires a server restart!

There's a few files I messed about, in order to have the configs take place, when the ADMIN_UI variable is true. It works but, I believe the way we handle the .env settings, and how we use them throughout LNbits should be refactored! It doesn't make much sense to be reading the .env file every time we call some VARIABLE, and it made it harder to have the AdminUI settings to actually work. There's probably still some cleanup and testing to be made!

lnbits/config.py
```diff
@@ -0,0 +1,62 @@
...
+    force_https: bool = Field(default=True, env="LNBITS_FORCE_HTTPS")
+    service_fee: float = Field(default=0, env="LNBITS_SERVICE_FEE")
+    hide_api: bool = Field(default=False, env="LNBITS_HIDE_API")
+    denomination: str = Field(default="sats", env="LNBITS_DENOMINATION")
+    # Change theme
+    site_title: str = Field(default="LNbits", env="LNBITS_SITE_TITLE")
+    site_tagline: str = Field(default="free and open-source lightning wallet", env="LNBITS_SITE_TAGLINE")
+    site_description: str = Field(default=None, env="LNBITS_SITE_DESCRIPTION")
+    default_wallet_name: str = Field(default="LNbits wallet", env="LNBITS_DEFAULT_WALLET_NAME")
+    theme: List[str] = Field(default="classic, flamingo, mint, salvador, monochrome, autumn", env="LNBITS_THEME_OPTIONS")
```
AaronDewes, line 40: This fails when no value is specified because the default is not a Python list.

```suggestion
    theme: List[str] = Field(default=["classic", "flamingo", "mint", "salvador", "monochrome", "autumn"], env="LNBITS_THEME_OPTIONS")
```

Making it a list like this fixed it for me

lnbits/extensions/admin/migrations.py
```diff
@@ -1,91 +1,152 @@
...
-    allowed_users = None
-    admin_user = None
-    default_wallet_name = None
-    data_folder = None
-    disabled_ext = None
-    force_https = True
-    service_fee = 0
-    funding_source = ""
+    # users/server
+    user = conf.admin_users[0]
```
AaronDewes, line 11: This throws an error if conf.admin_users is empty

AaronDewes (changes requested):
Two minor issues I found:

talvasconcelos:
@AaronDewes thanks for your input, I'll fix this and commit again!

AaronDewes:
I'll probably have some more feedback later on, I'm currently playing around with this UI on my node. Great work so far!

PR #574, opened 2022-03-27
Updates to documentation

fusion44:
Smaller updates and refinements to the existing documentation.


arcbtc:
Its all very messy, we currently link to the dev install as default, as the other install guide became overwhelming
https://github.com/lnbits/lnbits-legend#running-lnbits

arcbtc:
In fact, much of the install guide could go in a "Common troubleshooting" doc, maybe we should add that

Issue #573, opened 2022-03-26
Add Toman & IRrial (iran currency)


alefvanoon:
Exchange api for IRrial:
https://api.nobitex.ir/market/stats?srcCurrency=btc&dstCurrency=rls
https://api.exir.io/v1/ticker?symbol=btc-irt

https://en.wikipedia.org/wiki/Iranian_toman
https://en.wikipedia.org/wiki/Iranian_rial

PR #576, opened 2022-03-30
add IRT api

lilcheti:
As mentioned in #573 we would be happy if you add the Iranian Toman Currency. Blue wallet also uses this [api](https://api.exir.io/v1/ticker?symbol=btc-irt) for IRT.


NavidGoalpure:
It will be great

fusion44 (approved):
LGTM!

PR #578, opened 2022-03-31
lnbits.org -> https://lnbits.com

kilrau:
simple url fix


arcbtc:
Thanks, but .org used to point to docs. Will try to get fixed

PR #596, opened 2022-04-16
Adds generic QR maker endpoint, url can be embedded 

arcbtc:
Useful for developers using LNbits, so they do not have to use additional QR libraries. The url can be embedded in `` tags in pages, emails, etc

An example can be seen here:

https://nostrfy.com/api/v1/qrcode/some%20data%20that%20will%20make%20a%20qrcode

and embedded using ``


lnbits/core/views/api.py
```diff
@@ -7,6 +7,9 @@
 from urllib.parse import ParseResult, parse_qs, urlencode, urlparse, urlunparse
 
 import httpx
+import pyqrcode
```
fusion44, line 10: Any reason you've chosen pyqrcode instead of https://pypi.org/project/qrcode/?

pyqrcode seems to be pretty much unmaintained. Last release was in mid-2016.

lnbits/core/views/api.py
```diff
@@ -580,3 +583,23 @@ async def api_fiat_as_sats(data: ConversionData):
...
+
+    return StreamingResponse(
+        _generator(stream),
+        headers={
+            "Content-Type": "image/svg+xml",
+            "Cache-Control": "no-cache, no-store, must-revalidate",
+            "Pragma": "no-cache",
+            "Expires": "0",
+        },
+    )
```
fusion44, line 605: Looks good to me. In the long run we might have to rate limit this call. 

Or make it a paid service :thinking:

lnbits/core/views/api.py
arcbtc: Because of bloat pillow

lnbits/core/views/api.py
arcbtc: Yeah, I was thinking that, could be great. Can I rate limit a specific endpoint per ip in fastapi?

fusion44: There are packages that add rate limiting capabilities to routes. As it is an additional dependency I'd only introduce this if we see abuse in future and we absolutely need to.

slowapi: https://github.com/laurentS/slowapi

PR #638, opened 2022-05-26
fix a problem for some currencies in Opennode

ashkanjalaliQ:
to further integrate LNbits, all rates must be taken from one service, otherwise some rates will be different. For example, the IRT(Iranian Toman) rate in Opennode is wrong.
(the first and second commits are wrong)


callebtc:
This PR breaks the function `api_fiat_as_sats` in its original use case and introduces loads of logic that is hard to understand. Closing until it becomes evident that this is a useful change.

Issue #640, opened 2022-05-27
Spark Integration Improvement Suggestion


mrbitcoiner:
When LNBITS is using sparko as funding source and pays a invoice, only the bolt11 is passed to the spark RPC, so the maximum allowed fee for the payment will be 0,5% (c-lightning pay plugin default fee limit). That reduces the probability of the payment being settled. I believe that the solution is pass the parameter maxfee or exemptfee + maxfeepercent with the lnbits default fee reserve as value.

PR #641, opened 2022-05-27
Added maxfee parameter to Spark pay method 

mrbitcoiner:
Working on c-lightning 0.11.0.1

Fixing the issue #640

### DIFF
--- a/lnbits/wallets/spark.py
+++ b/lnbits/wallets/spark.py
@@ -109,7 +109,10 @@ class SparkWallet(Wallet):
 
     async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse:
         try:
-            r = await self.pay(bolt11)
+            r = await self.pay(
+                bolt11=bolt11,
+                maxfee=fee_limit_msat,
+            )
         except (SparkError, UnknownError) as exc:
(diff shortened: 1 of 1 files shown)


callebtc:
Thanks! 

LGTM

callebtc:
According to https://readthedocs.org/projects/lightning/downloads/pdf/master/  from CLN 0.11.0.1 on`maxfee` can be used instead of `maxfeepercent`. Maybe we can use the same for `clightningwallet` as well. 

@mrbitcoiner I assume you have tested it? Do you think we should use `maxfeepercent` for backends that haven't upgraded to the newest version?

mrbitcoiner:
> According to https://readthedocs.org/projects/lightning/downloads/pdf/master/ from CLN 0.11.0.1 on`maxfee` can be used instead of `maxfeepercent`. Maybe we can use the same for `clightningwallet` as well.
> 
> @mrbitcoiner I assume you have tested it? Do you think we should use `maxfeepercent` for backends that haven't upgraded to the newest version?

I've tested, running on my c-lightning v0.11.0.1 instance through spark and `maxfee` is working fine.

callebtc:
Merging.

Issue #456, opened 2021-12-08
FastAPI branch: sometimes a HTML repsonse returned in API


cryptoteun:
I performed some tests on the FastAPI branch and sometimes I get a HTML response instead of a json. 

Some examples:
**lnurlp**

```
POST /lnurlp/api/v1/links
body: {}
Response = HTML
```

```
POST /lnurlp/api/v1/links
body: {}
Response = HTML
```

```
POST /lnurlp/api/v1/links
body: valid request
Response = Json, lnurlp 9 created: but strange thing with this tag:    "lnurl": {} (will raise separate issue)
```

When I try to retrieve lnurlp via: GET /lnurlp/api/v1/links/9
HTML response with error: <h5 class="q-my-none">&#39;PayLink&#39; object has no attribute &#39;_asdict&#39;

**lnurlw**
```
POST /withdraw/api/v1/links
body: {}
Response = HTML
```
```

POST /withdraw/api/v1/links
body: some missing fields
Response = HTML
```

```
POST /withdraw/api/v1/links
body: valid request
Response = HTML response which doesn't conain an error
```

**paywall**
```
POST /paywall/api/v1/paywalls
body: {}
Response = HTML

```
```
POST /paywall/api/v1/paywalls
body: some missing fields
Response = HTML
```

```
POST /paywall/api/v1/paywalls
body: valid request
Response = Json
```

bitkarrot:
Just tested 

Curl example
curl -X GET https://fast.laisee.org/api/v1/links/<withdraw_id> -H "X-Api-Key: <api key>"

on latest main and it still returns HTML and not json data

Curl example on `List all withdraw` links works however
GET /withdraw/api/v1/links

PR #653, opened 2022-06-02
Fix api_link_retrieve in withdraw extension

iWarpBTC:
It's apparently extra in there and causing responses of HTML instead of a json like in #456


iWarpBTC:
duplicity

PR #655, opened 2022-06-07
Fixed problem to exchanging IRT to BTC.

cybertech1981:
As mentioned in lnbits#573 we would be happy if you adding the following to the project for the reasons mentioned:
1) Add Nobitex exchange, because many Iranians use this exchange and it is almost a price reference for digital currencies of this exchange.
2) Modify the output type of "fiat_amount_as_satoshis()" Function from int to float due to decimal output in converting Toman to Satoshi in int mode returns zero output.
For example : 1 IRT => 0.0104 Sat

Thanks & Regards
Amir Makouei

### DIFF
--- a/lnbits/utils/exchange_rates.py
+++ b/lnbits/utils/exchange_rates.py
@@ -180,6 +180,12 @@ class Provider(NamedTuple):
 
 
 exchange_rate_providers = {
+    "nobitex": Provider(
+        "Nobitex",
+        "nobitex.ir",
+        "https://api.nobitex.ir/v2/orderbook/{FROM}IRT",
+        lambda data, replacements: data["lastTradePrice"],
+    ), 
     "exir": Provider(
         "Exir",
         "exir.io",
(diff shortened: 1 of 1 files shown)


lnbits/utils/exchange_rates.py
```diff
@@ -289,8 +295,8 @@ async def get_fiat_rate_satoshis(currency: str) -> float:
     return float(100_000_000 / (await btc_price(currency)))
 
 
-async def fiat_amount_as_satoshis(amount: float, currency: str) -> int:
-    return int(amount * (await get_fiat_rate_satoshis(currency)))
+async def fiat_amount_as_satoshis(amount: float, currency: str) -> float:
+    return float(amount * (await get_fiat_rate_satoshis(currency)))
```
callebtc, line 299: Satoshis need to be integers, lnbits doesn't support anything below 1 sat.

callebtc (changes requested):
Thank you for the PR! Please have a look at the comment. We can't change the type of the function, it might break other things.

lnbits/utils/exchange_rates.py
cybertech1981: I change to :


image

callebtc:
You can't round up a 0 satoshi invoice to 1 satoshi. 

Closing this PR since the use case doesn't really seem obvious. Sub-satoshi amounts should not be used. The implementation should rather round the fiat value to something reasonable, instead of affecting all other `fiat_to_satoshi` calls.

PR #661, opened 2022-06-09
Repo: Clean up and remove old files

callebtc:
Removes unnecessary files from the repo. Most are remainders of the old quart API. If you think some of these should stay, please provide a reason what they are used for.

Delete:
```
pytest.ini 
package.json 
package-lock.json 
mypy.ini 
app.json 
Procfile 
.prettierrc 
.editorconfig
```

Move:
```
conv.py -> tools/conv.py
```


fiatjaf:
Are you giving up on all code format uniformization, linting etc?

.editorconfig
```diff
@@ -1,16 +0,0 @@
...
-trim_trailing_whitespace = true
-indent_size = 2
-indent_style = space
-
-[*.md]
-trim_trailing_whitespace = false
-
-[*.py]
-indent_size = 4
-indent_style = space
```
fusion44, line 16: Seems in most editors this file is only usable with an additional plugin anyway. Not sure anyone actually uses this? I don't.

.prettierrc
```diff
@@ -1,12 +0,0 @@
...
-  "arrowParens": "avoid",
-  "insertPragma": false,
-  "printWidth": 80,
-  "proseWrap": "preserve",
-  "singleQuote": true,
-  "trailingComma": "none",
-  "useTabs": false,
-  "jsxBracketSameLine": false,
-  "bracketSpacing": false
-}
```
fusion44, line 12: I'd not remove this. I guess pretty much everyone is using prettier to format the .vue files.

Procfile
```diff
@@ -1 +0,0 @@
-web: hypercorn -k trio --bind 0.0.0.0:5000 'lnbits.app:create_app()'
```
fusion44, line 1: Used when starting an instance on Heroku. But the command is wrong since we've switched to FastAPI. We could remove that and re-add if someone is interested in it.

app.json
```diff
@@ -1,7 +0,0 @@
-{
-  "scripts": {
-    "dokku": {
-      "predeploy": "quart migrate && quart assets"
-    }
-  }
-}
```
fusion44, line 7: Not used with FastAPI. Purrrge!

package-lock.json
```diff
@@ -1,32 +0,0 @@
...
-  },
-  "dependencies": {
-    "prettier": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/prettier/-/prettier-2.1.1.tgz",
-      "integrity": "sha512-9bY+5ZWCfqj3ghYBLxApy2zf6m+NJo5GzmLTpr9FsApsfjriNnS2dahWReHMi7qNPhhHl9SYHJs2cHZLgexNIw==",
-      "dev": true
-    }
-  }
-}
```
fusion44, line 32: Also has some prettier stuff in it. Not sure we should remove that.

fusion44 (commented):
I'd keep the prettier files since they contain some specific formatting rules which are probably used in the project. I think everything else can be removed.

package-lock.json
callebtc: This file is matched by `.gitignore` though. Should I add it?

callebtc:
Closing due to second attempt in PR #748.

PR #663, opened 2022-06-10
lnurlp: support for floating point fiat values

### LINKED ISSUE #662: LNURLp links with floating point fiat currency not created
Creating an LNURLp with a value of for example 1.5€ does not work.

callebtc:
Allows for creation of floating point fiat values in lnurlp links (such as 0.25 EUR).

Fixes #662.

### DIFF
--- a/lnbits/extensions/lnurlp/lnurl.py
+++ b/lnbits/extensions/lnurlp/lnurl.py
@@ -29,11 +29,11 @@ async def api_lnurl_response(request: Request, link_id):
             status_code=HTTPStatus.NOT_FOUND, detail="Pay link does not exist."
         )
 
-    rate = await get_fiat_rate_satoshis(link.currency) if link.currency else 1
+    rate = await get_fiat_rate_satoshis(link.currency) / 100 if link.currency else 1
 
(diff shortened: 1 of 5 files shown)


lnbits/extensions/lnurlp/lnurl.py
```diff
@@ -29,11 +29,11 @@ async def api_lnurl_response(request: Request, link_id):
...
         )
 
-    rate = await get_fiat_rate_satoshis(link.currency) if link.currency else 1
+    rate = await get_fiat_rate_satoshis(link.currency) / 100 if link.currency else 1
 
     resp = LnurlPayResponse(
         callback=request.url_for("lnurlp.api_lnurl_callback", link_id=link.id),
-        min_sendable=math.ceil(link.min * rate) * 1000,
+        min_sendable=round(link.min * rate) * 1000,
         max_sendable=round(link.max * rate) * 1000,
```
callebtc, line 37: If someone knows why this is `math.ceil` please tell. Having `math.ceil` in one and `round` in the causes problems for fiat since min can become larger than max.

PR #677, opened 2022-06-22
SCRUB extension

talvasconcelos:
SCRUB® extension (we're minting the term!)

Turns an LNbits wallet as a passthrough to an LNURLpay or LNaddress!


arcbtc:
image
Getting an error

arcbtc:
Cant get funds to scrub from wallet

talvasconcelos:
> 

your listener isn't working! it doesn't trigger whatever happens in tasks.py

talvasconcelos:
ahh damn you, @callebtc !!! 

now his tests are forcing me to add scrub to conv.py...

callebtc:
> ahh damn you, @callebtc !!!
> 
> now his tests are forcing me to add scrub to conv.py...

That sounds great! Hah, gotcha!

Btw, it would amazing if you could also add a sample extension db file to `tests/data/mock_data.zip` which is used for the migration test!

callebtc:
> Getting an error

This Exception should be fixed in main now, please rebase!

PR #679, opened 2022-06-23
Change percentage data type from int to float

exitfiat:
This enables split percentages <1%, could be useful for some usecases.
Tested this on my raspiblitz, but i can't code and this is my first pull request so pls check if this is correct :D


talvasconcelos:
The HTML file also needs to comply with this change and it probably worked ok for you if you're on SQLite but, Postgres is a bit more demanding on the type of data you feed it! So, the migrations.py file also needs some work done!

callebtc:
Dupe of #690. Closing.

Issue #680, opened 2022-06-24
create lnurlp is not giving lnurl in response


sawrubgupta:
Hey 
api endpoint :- lnurlp/api/v1/links
Method - POST

I am not getting lnurl here in response 

{
    "id": 6,
    "wallet": "f34a80cde81f4c2299819f23dacaf25b",
    "description": "brewry payment",
    "min": 100,
    "served_meta": 0,
    "served_pr": 0,
    "webhook_url": "api/v1/lnurlp/lnurlpWebhook",
    "success_text": null,
    "success_url": null,
    "currency": "INR",
    "comment_chars": 0,
    "max": 10000,
    "lnurl": {}
}

see lnurl field is empty
Please add lnurl here in response so we dont have to hit different api for this

Issue #681, opened 2022-06-24
In legend server, two api calls can be changed to one


sawrubgupta:
Hey
In lnbits you are hitting two api's, 
one to create lnurl link 
and one to fetch lnurl from lnurlp id

so you can simply return lnurl in reponse of create lnurlp link api

check lnbits legend code for reference.

PR #685, opened 2022-06-27
return lnurl in create link

talvasconcelos:
fix issue #680 and #681


talvasconcelos:
Hi, thank you for your input. Can you please be a bit more specific on where it's breaking?

callebtc:
@bezysoftware do you really rely on this field being empty (which is a bug)?

bezysoftware:
Apologies I misread the PR, I wrongly assumed it would change the response object, instead of populating a missing field. Please ignore my previous comment

talvasconcelos:
what's it breaking?!

PR #689, opened 2022-06-28
Corrected example CURL requests in multiple extension API docs

blackcoffeexbt:
The move to FastAPI changed the API URLs. This change updates the example URLs for LNbits extensions


callebtc:
Thank you for this painstaking PR!

PR #691, opened 2022-06-28
Fix Return 1 as a Minimum of Satoshi amount.

cybertech1981:
As mentioned in lnbits#655 :

1) I changed the code when you did not accept to change the output type of the "fiat_amount_as_satoshis ()" function to float.
because of When the fiat currency price like IRT is lowest from satoshi, the function returns 0 and addon like BTCPOS can't calculate correct amount of satoshi and create invoice.
For example : 1 IRT => 0.0104 Sat then Function return 0,

2) Add Nobitex exchange, because many Iranians use this exchange and it is almost a price reference for digital currencies of this exchange.

### DIFF
--- a/lnbits/utils/exchange_rates.py
+++ b/lnbits/utils/exchange_rates.py
@@ -180,6 +180,12 @@ class Provider(NamedTuple):
 
 
 exchange_rate_providers = {
+    "nobitex": Provider(
+        "Nobitex",
+        "nobitex.ir",
+        "https://api.nobitex.ir/v2/orderbook/{FROM}{TO}",
+        lambda data, replacements: data["lastTradePrice"],
+    ), 
     "exir": Provider(
         "Exir",
         "exir.io",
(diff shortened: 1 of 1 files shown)


callebtc:
Hey! 
If `1 IRT => 0.0104 sat` then we shouldn't assume that it's 1 sat. At that point, currency conversion doesn't make any sense anyway. 

Another point: Adding an exchange just for one ticker (IRT) seems like a bad idea since it will receive requests for all other tickers as well. I think one exchange for IRT was added already. As long as we don't specify in the code which currency should be requested from which exchange, I would be against adding another one for a single ticker.

lnbits/utils/exchange_rates.py
```diff
@@ -290,7 +296,11 @@ async def get_fiat_rate_satoshis(currency: str) -> float:
 
 
 async def fiat_amount_as_satoshis(amount: float, currency: str) -> int:
-    return int(amount * (await get_fiat_rate_satoshis(currency)))
+    amount_as_sattoshi = amount * (await get_fiat_rate_satoshis(currency))
+    amount_as_sattoshi = round(amount_as_sattoshi)
+    if amount_as_sattoshi==0:
+        amount_as_sattoshi=1
+    return int(amount_as_sattoshi)
```
callebtc, line 303:
```suggestion
    return int(round(amount * (await get_fiat_rate_satoshis(currency)))) or 1 
```

callebtc:
Closing.

cybertech1981:
Hello
Why dont merge ?

PR #693, opened 2022-06-29
always send json error response

talvasconcelos:


callebtc:
Closing after DMs with Tal until we decide to tackle this again.

PR #697, opened 2022-07-03
Enhancements to TPOS Extension

leesalminen:
This PR adds new functionality to the TPOS extension. 

- Add new tip % option to TPOS extension.      
        - When adding a new TPOS, a user can choose 1 or more tip % options to be displayed to the customer.
        - When adding a new TPOS, a user can choose a wallet to send all collected tips to.
        - Demo Image:
        
Screen Shot 2022-07-03 at 11 01 38 AM

Screen Shot 2022-07-03 at 11 06 31 AM

- UI Refresh on TPOS extension.
        - Moved the share button to the top navigation, next to the TPOS name, and changed the icon to a more recognizable one.
        - Re-arranged the buttons on the keypad to be more ergonomic.
        - Added Progressive Web App (PWA) support for the terminal page.
        - Demo image: 
        
Screen Shot 2022-07-03 at 11 00 31 AM

PR #713, opened 2022-07-07
Fix/tpos tips #697

talvasconcelos:
Small fixing for the recently merged #697 

- Tip now add to the invoice value
- show tip amount
- if a user had existing TPoS created, without the tip options, it doesn't break anymore
- some models typing and small fastpai-ish corrections


lnbits/extensions/tpos/tasks.py
```diff
@@ -60,7 +60,7 @@ async def on_invoice_paid(payment: Payment) -> None:
         payment_request="",
         payment_hash=payment.payment_hash,
         amount=tipAmount,
-        memo=payment.memo,
+        memo="Tip",
```
leesalminen, line 63: May I propose here doing something like `f"Tip {payment.memo}"` here? In my use case, `payment.memo` has a unique identifier for a specific order. This helps the business just reconcile everything.

leesalminen:
Hi @talvasconcelos,

Sorry for any issues here. Would be happy to take on any of this as I caused it in the first place.

lnbits/extensions/tpos/tasks.py
talvasconcelos: Fixed.

PR #721, opened 2022-07-08
Add Nix and Poetry to the project

MatthewCroughan:
This PR begins using Nix via `flake.nix`, upgrades a few library dependencies of the project, fixes some of their usage, and also begins using [Poetry](https://python-poetry.org/) for the project. This makes the project much more reproducible.

# Notable Python changes

1. CSS is no longer compiled every single time lnbits runs. Instead, it is implemented in `build.py` as part of `poetry run` or `poetry build`. This means the compiled CSS is bundled with the program and distributed, rather than wrongly trying to mutate the source code directory (LNBITS_PATH) on every launch.
2. FastAPI and Starlette have been updated to their latest version
3. I have corrected some of the references to relative paths by using the `StaticFiles()` function from Starlette, see the commit comment here 2e237d215d8d56a5a004e22e8b58925f8b0fb56f
4. `lnbits/server.py ` has been added to provide a CLI entrypoint to the program, this is what will be ran via `poetry run lnbits`

# What are the Nix additions?

This adds a Nix flake which provides:
- A `nixosModule` which allows someone to add `services.lnbits.enable = true;` to their NixOS configuration, and voila, they're now running lnbits as a systemd service.
- A NixOS VM test which can be built with `nix flake check` which
  1. Spawns two QEMU virtual machines running NixOS. One named `client` the other named `server`
  5. Waits for the `server` to be available
  6. Uses `curl` on the client to prove that lnbits is working by expecting  `<title>LNbits</title>`
- A `nix develop` shell which simply provides prettier
- An output `.#lnbits` that can be built with `nix build`. The result can then be ran as follows
 
  ```
  user: matthew
  ❯ nix build .#lnbits

  user: matthew
  ❯ ./result/bin/lnbits --help
  Usage: lnbits [OPTIONS]

    Launched with `poetry run start` at root level

  Options:
    --port TEXT  Port to run LNBits on
    --host TEXT  Host to run LNBits on
    --help       Show this message and exit.

  user: matthew
  ❯ LNBITS_DATA_FOLDER=$(mktemp -d) ./result/bin/lnbits
  2022-07-08 10:27:13.27 | INFO | Starting LNbits
  2022-07-08 10:27:13.27 | INFO | Host: 127.0.0.1
  2022-07-08 10:27:13.27 | INFO | Port: 5000
  ```

# General changes and considerations for reproducibility

1. Relative paths are present all over the codebase. I've tried to fix some of this in 2e237d215d8d56a5a004e22e8b58925f8b0fb56f, as the comment says, this means that the code can only be ran from within the git directory, as lnbits assumes it will have access to `./templates` in some cases, for example.
2. LNBITS_PATH shouldn't exist, instead `importlib` or `StaticFiles()` should be used since they already work reliably to provide an absolute path to the source code.
3. The code assumes wrongly that the source code (LNBITS_PATH) is modifiable, which it will not be when distributed with a package manager. I have fixed at least one case of this in this PR, since I have removed the `transpile_scss()` function and placed it in `build.py` as its own step of the build via Poetry, rather than at runtime of the program.
4. There are a lot of unnecessary files that need to be gradually removed now that Poetry + Nix is available. These files can't be removed immediately until the `Makefile` is fixed, although arguable the Makefile should also be destroyed. Here's a list of files should no longer be needed once we adopt using Poetry and Nix
  - requirements.txt
  - package.json
  - package-lock.json
  - Pipfile
  - Pipfile.lock
  - Makefile

5. Nix should be used natively in CI via https://hercules-ci.com/dashboard, CI will take seconds and everything will be better, trust me.

# What is Nix?

Nix builds are reproducible and deterministic. They do not use the host's environment for any reason, as they are sandboxed. A build on my machine will result in the same hash on your machine, as on another machine every time, since all inputs are controlled. This is the goal, and it is successful at this in the majority of cases. There are some exceptions with Java and reading system time, but issues like these are easily identified and worked around.

The Nix thesis can be read here:
https://edolstra.github.io/pubs/phd-thesis.pdf


dni:
this is a very interesting PR :+1: , i just wanted to leave this here, because it will take more time for me evaluate this PR.

but it is a very promising one.

MatthewCroughan:
@afilini I haven't done extensive testing of the program at runtime, so it's entirely possible that patch is still required. But I have made a VM Test for this PR that you can run with `nix build github:matthewcroughan/lnbits-legend/matthewcroughan/nixify#checks.x86_64-linux.vmTest`. This proves that making a `curl` request to the uvicorn webserver doesn't produce any errors about jinja templating. It would be expected that it would produce errors if your patch was still necessary. But I'm not too familiar with Jinja, so maybe this patch is still needed. Either way, we should fix the original Python code in all cases, since we have this opportunity.

@callebtc Poetry wants to be at the root of the project, as the main system for building and distributing the Python code. The flake.nix can be put in a subdirectory, but it's very unergonomic as all the nix commands would have to refer to this subdirectory. `build.py` is at the root, because the source code for this repository is laid out poorly and it is not possible to import from `.commands` without triggering `__main__.py` to run the LNBits main code. I did not want to modify too much of the original python code, so I left this as is.

I have explained what build.py does here in the PR body above. I'll copy-paste here so you can see it:
> CSS is no longer compiled every single time lnbits runs. Instead, it is implemented in build.py as part of poetry run or poetry build. This means the compiled CSS is bundled with the program and distributed, rather than wrongly trying to mutate the source code directory (LNBITS_PATH) on every launch.

arcbtc:
Working great. This is a huge upgrade. Switching to Poetry alone makes this worthwhile. We should dump pipenv and venv support completely. People will have two deployment options, Nix and Poetry, both make us more responsible, without overcomplicating for developers. Poetry will allow us to push to PyPi. Having nix support will get us ready for the inevitable NixNode revolution.

Reproducibility/stability win!

arcbtc:
Note, I did have to add:
`experimental-features = nix-command flake` to `/etc/nix/nix.conf` when running `nix build .#lnbits`

PR #727, opened 2022-07-11
WIP: Fix payment timeouts

callebtc:
WIP


callebtc:
This PR is dangerous. Only the timeout increase should be considered, the rest is BS. Closing.

PR #730, opened 2022-07-12
error in `/usermanager/api/v1/extensions`

leesalminen:
Requests to `POST /usermanager/api/v1/extensions` would silently fail due to a missing `await`.


leesalminen:
Created an issue for this: https://github.com/lnbits/lnbits-legend/issues/744

callebtc:
Good catch! LGTM

PR #733, opened 2022-07-13
New Extension: Invoicing

leesalminen:
This extension allows users to create "traditional" invoices (not in the lightning sense) that contain one or more line items. Line items are denominated in a user-configurable fiat currency. Each invoice contains one or more payments up to the total of the invoice. Each invoice creates a public link that can be shared with a customer that they can use to (partially or in full) pay the invoice.

Screen Shot 2022-07-12 at 8 14 34 PM

Screen Shot 2022-07-12 at 8 01 30 PM

Open to feedback on this with regards to available features, etc, while in draft PR mode!

## Todo:
* [x] Make sure to rebase to `main`
* [x] Run `make format`
* [x] Write tests
* [x] Add migrations for new extensions to `tools/conv.py`
* [x] Write `README.md` for extension


leesalminen:
> This is awesome! Congrats!
> 
> It would be cool to have a printable version of the invoice, for the merchant, probably needed for accounting.

That makes sense. I've now pushed an update to the extension that has `@media print` CSS rules, and a print button next to the share button. I've attached a sample.


[invoice.pdf](https://github.com/lnbits/lnbits-legend/files/9104733/invoice.pdf)

nanda77:
Not adding any tax and discount input?

callebtc:
Great PR! Couple comments:

* I have done an unsolicited Edit of your top post with a checklist of things to do before a merge.

* As you  might have noticed, we have tests now! With this PR, there is a [postgres test](https://github.com/lnbits/lnbits-legend/runs/7326562725?check_suite_focus=true) that's failing.

    `psycopg2.errors.UndefinedObject: type "invoice_status_type" does not exist` context: `ERROR at setup of test_core_views_generic`

* Could you also a couple of tests for this extension specifically? We don't have many extension tests yet, but it would be great to test at least the API and some expected responses. 

Let me know if you need any help. This pipeline is new territory for LNbits but it will make life a lot easier in the future!

arcbtc:
Is there a guide to making these tests? I notice there has been a pause here. Tests are important of course, but we dont want to discourage people making extensions by creating too much complexity. If tests are some extensions are expected to have an example is needed first.

callebtc:
I wouldn't say it's expected but definitely encouraged as a best practice. Most extensions don't have tests. There is no guide yet but i could write one in the dev docs.

For now, the existing tests should serve as a basic example. It's quite easy really. The simplest thing to do is to call the API endpoints and see if you get what you expect. More testing is always possible.

nanda77:
Would be nice if we can add a predefine items into the extension..... who is going to type items each and every time. I got a restaurant with lots of items, if I can select those items quickly when creating invoice, thats better.

talvasconcelos:
> Would be nice if we can add a predefine items(with price) into the extension..... who is going to type items each and every time. I got a restaurant with lots of items, if I can select those items quickly when creating invoice, thats better.

I get your point, but that would make a full fledged Point of Sale software... I see it like, the merchant has already a PoS system/software running, where he makes the entries of products! He could then enter a pointer to the invoice generated on the PoS, a number or something! 

What you describe would have merchants, adding products to the DB!

talvasconcelos:
Hi @leesalminen , I'm trying to make this pass checks, and maybe tweak it a litlle bit, if that's ok! Can i rename this to accounting, instead of invoices ?

talvasconcelos:
woohooo!! only tests missing!! 

@callebtc will take bleskomat as guide but, i'll bug you if i need insight...

talvasconcelos:
@leesalminen could you write the README for this extension?

talvasconcelos:
@leesalminen could you write the README for this extension? 

Also, what's the reason for having the invoice created as draft as default? even I set it to be open on creation, i need to go and update it by hand to change status to open! can I change this behaviour?

leesalminen:
> Also, what's the reason for having the invoice created as draft as default? even I set it to be open on creation, i need to go and update it by hand to change status to open! can I change this behaviour?

I've just pushed a fix for this bug. Had to set a specific property in `models.py`

leesalminen:
> @leesalminen could you write the README for this extension?

Just pushed up a readme.

callebtc:
Great extension! I was just playing with it and found that this happens when you enter a fiat value smaller than 0.01
image

callebtc:
Great job! LGTM. Merging after all checks are green.

callebtc:
Found a little issue with the mobile layout. Can it be adjusted for smaller screens? I guess that's also something we could do better in LNbits generally.

leesalminen:
> Found a little issue with the mobile layout. Can it be adjusted for smaller screens? I guess that's also something we could do better in LNbits generally.
> 
> 

See https://github.com/lnbits/lnbits-legend/pull/874

Issue #352, opened 2021-10-04
lnd default invoice expiry is 10 minutes


pseudozach:
Is there any reason why default lnd default invoice expiry time is 10 minutes? can it be changed to 1 hour?
https://github.com/lnbits/lnbits/blob/19d432b404ede4ecc5206e14c24f40d42346ebc1/lnbits/wallets/lndgrpc.py#L129

I can also send a PR, just wanted to check if there's any justification. Short invoice expiry time like 10 minutes is bad UX. Unless there's any good reason, it could even be 1 day?

arcbtc:
Most people seem to like invoices failing after a shorter period of time, as when the invoice is pending the funds are not available

pseudozach:
right but in some use-cases:
- submarine swaps that may take more than 10 minutes (multiple block confirmation time)
- non-interactive/emailed payment request: sender/receiver may be in different timezones so the response may be up to 24 hours.
10 minute expiry time causes inconvenience, and since by default lnd creates 24hour valid invoices but lnbits artificially reduces that. 
Could it be at least configurable? Would you be open to it? Default can still stay at 10 minutes.

sawrubgupta:
We need to set a parameter by which user can enter whatever expiry time he wants. That is going to be more convenient.

Semisol:
> non-interactive/emailed payment request: sender/receiver may be in different timezones so the response may be up to 24 hours.

You could also use LNURLp.

> Could it be at least configurable? Would you be open to it? Default can still stay at 10 minutes.

For user configurable: That would be nice, but have a hard limit on maximum. Too much time may result in invoices that won't be used for days.
For admin configurable: I'm all for it.

dni:
custom expiry is merged

PR #753, opened 2022-07-18
Configure invoice expiry from ENV variable

cryptoteun:
PR makes the invoice expiry configurable. Is also a fix for #352 

Please review carefully:

- I'm not a DEV so not sure if I followed coding standards/best practices
- Tested on CLN, that worked well. 
- But not tested on a LND instance.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -40,6 +40,10 @@ LNBITS_BACKEND_WALLET_CLASS=VoidWallet
 # VoidWallet is just a fallback that works without any actual Lightning capabilities,
 # just so you can see the UI before dealing with this file.
 
+# Set the default expiry of the invoices create by LNbits (in seconds)
+# Applies to SparkWallet, CLightningWallet, LndRestWallet
+INVOICE_EXPIRY=3600
+
 # Set one of these blocks depending on the wallet kind you chose above:
 # SparkWallet
 SPARK_URL=http://localhost:9737/rpc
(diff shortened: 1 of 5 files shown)


dni:
cool, i started a branch last week, where i make expiry also configurable, on a invoice to invoice basis. ENV variable is nice, but i think the ability to define your timeout on each invoice, would be very awesome also.

cryptoteun:
> 

Yeah, that's why I made this one only for lnd and cln. But I have to check tonight why the build is failing.
It worked local so perhaps it's something with setting the ENV for the regtest. @callebtc, any idea?

dni:
it failing becuase it cant get the env variable, i guess for regtest a custom one is used where your variable is missing?

callebtc:
Thank you, very useful PR. I'd like to tackle this one with you. 

Currently working on an updated payment interface for LND wallets so that will have to be rebased to PR #745 at some point.

dni:
ok, i guess, this one has to be changed if we decide to use a env variable. https://github.com/lnbits/lnbits-legend/blob/main/.github/workflows/regtest.yml

callebtc:
Concept ACK but the way it is solved here is suboptimal (see my comment above). 

I reckon that it would be better to pass the desired expiry to the wallet init in `settings.py`.

We also would need to add tests for this PR.

cryptoteun:
Spoke about it with Calle, but don't know how to resolve correctly. Calle will look into it

PR #754, opened 2022-07-19
Update requirements.txt for a pip3 install

SatoriHoshiAiko:
This is a recomendation to include all the pipfile locking dependencies into requirements.txt so all the required packages can be installed directly with a "pip3 install -r requirements.txt"

See further that installation covers pipenv and venv, however, on macOS Monterey pipenv fails to lock the dependency secp256k1 using pip instead of pip3.

This is a broken dependency, and can be tested (on macOS Monterey) with pip install secp256k1, and it will reproduce the error.

After further investigation it appears that this package doesn't install properly using setuptools, and it needs --use-pep517, since setuptools is deprecated on that package. I haven't tested this method, but since pipenv is going to attempt to lock without this flag or option, the bug will consistently reproduce on macOS.

To resolve this larger circumstance, my recommendation for macOS is to brew install python@3.9 - 3.10 and 3.11 will not work and would be unsupported without further updates to other dependencies. 

The best way is to be utilizing pip3 and not pip. Otherwise we have the bug with secp256k1. Pipenv automatically call on pip.

So instead installation can be done with sudo pip3 install -r requirements.txt

Where this will update the full dependency tree including all the modules and dev modules.

Then lnbits is simply run with:

python3 -m uvicorn lnbits.__main__:app --port 5000 --host 0.0.0.0

And directly using the pip3 packages and not a pip environment or v environment at all. 

Please see how this might fit better, and provide an equally simple method to installing and running where pipenv fails.


SatoriHoshiAiko:
Will need to add the following two lines:

jsonschema==4.4.0
pylightning

SatoriHoshiAiko:
I have tested and pip3 install -r requirements.txt is successful.

I am able to launch LNBits on macOS now with:

brew install python@3.9
git clone https://github.com/lnbits/lnbits-legend
cd lnbits 
cd pip3 install -r requirements.txt
python3 -m uvicorn lnbits.__main__:app --port 5000 --host 0.0.0.0

SatoriHoshiAiko:
MY last comment can be added as Option 3 for installation instructions as an afterthought. Of course modifying pp3 packages directly and not using a pipenv or venv can break your system python3 of other things require it (in this case, I had to modify jsonschema and docker-compose-1.29.1 broke with 4.4. But assuming a person is okay with minor issues elsewhere that can be easily fixed..

We ought to make a direct and non pipenv or venv installation method available. Since these just simply do not work, and create a wall of "uninstallable" on macOS, and I am sure there are other cases than just my own test environment.

Option 3 can have a "Warning - This will modify your python3 default packages and dependency tree. Use this is you are having problems with loading modules. Known bug - secp256k1 will not resolve using pipenv.

cd lnbits-legend
sudo pip3 install -r requirements.txt
python3 -m uvicorn lnbits.main:app --port 5000 --host 0.0.0.0

dni:
hello! 

as far as i know ```pylightning``` is only needed when you use clightning backend, it was mentioned in the installation guide but i could not find it, @callebtc did you remove it?

in my opinion support for direct installs on macosx, should not be a requirement because it is never needed in production.

we could add a notice in the install guide for macosx users, but i would not change the requirements.txt at all.

callebtc:
Hi. We can't recommend installing LNbits without a virtual environment. That will just mess up a users computer. I am running it in venv on macOS so it's definitely possible. The only thing the user needs to take care of is secp256k1, no need to update the requirement file for that. A simple guide would do (for venv or Pipenv)

The requirement file that you've edited has lots of duplicates now, I don't see what you have actually added.

The packages requirements for each wallet are listed here: https://github.com/lnbits/lnbits-legend/blob/main/docs/guide/wallets.md

Closing.

PR #760, opened 2022-07-19
WIP: Enable `mypy` linting of the entire code base

dni:
pr that fixes mypi.ini, adds type-protobuf and fixes mypy checks.

current status: all mypy tests pass

there are quite a lot of  ```#type: ignore``` that we probably should look at


callebtc:
What is the purpose of this PR ? Does it change anything? Please write more extensive descriptions so it can be properly reviewed.

It has 15 file changes but the description says that only an ini file changes.

Thanks :)

dni:
the purpose of this PR is the re-enable mypy checks, at the moment the status is ```Found 30 errors in 6 files (checked 41 source files)```
screenshot-1658296879

callebtc:
Nice! Please get accustomed to writing descriptions in the top post and updating the top post as you go along. Otherwise it gets hard to follow the progress!

See [here](https://github.com/lnbits/lnbits-legend/pull/728) for an example of how to do it ideally.

tests/core/views/test_api.py
```diff
@@ -140,3 +141,27 @@ async def test_decode_invoice(client, invoice):
...
+# check api_payment() internal function call (NOT API): payment status
+@pytest.mark.asyncio
+async def test_api_payment_with_key(invoice, inkey_headers_from):
+    # check the payment status
+    response = await api_payment(
+        invoice["payment_hash"], inkey_headers_from["X-Api-Key"]
+    )
+    assert type(response) == dict
+    assert response["paid"] == True
+    # not key, that's why no "details"
```
callebtc, line 166:
```suggestion
```

tests/core/views/test_api.py
```diff
@@ -140,3 +141,27 @@ async def test_decode_invoice(client, invoice):
...
+
+
+# check api_payment() internal function call (NOT API): payment status
+@pytest.mark.asyncio
+async def test_api_payment_without_key(invoice):
+    # check the payment status
+    response = await api_payment(invoice["payment_hash"])
+    assert type(response) == dict
+    assert response["paid"] == True
+    # not key, that's why no "details"
```
callebtc, line 153:
```suggestion
    # no key, that's why no "details"
```

requirements.txt
```diff
@@ -48,4 +48,4 @@ uvicorn==0.15.0
 uvloop==0.16.0
 watchgod==0.7
 websockets==10.0
-zipp==3.5.0
\ No newline at end of file
+zipp==3.5.0
```
callebtc: Remove non-change

Pipfile
```diff
@@ -35,9 +35,11 @@ pycryptodomex = "*"
...
-pytest-cov = "*"
+mock = "*"
 mypy = "*"
+pytest = "*"
 pytest-asyncio = "*"
+pytest-cov = "*"
 requests = "*"
-mock = "*"
+types-mock = "*"
+types-protobuf = "*"
```
callebtc, line 45: `Pipfile` shouldn't be updated without an appropriate update of of the `Pipfile.lock`.

lnbits/wallets/fake.py
```diff
@@ -80,8 +80,10 @@ async def create_invoice(
 
     async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse:
         invoice = decode(bolt11)
+        # TODO: no data here?
+        data: Dict = {"privkey": "missing"}
         if (
-            hasattr(invoice, "checking_id")
+            invoice.checking_id is not None
```
callebtc, line 86: Is this safe? What if invoice doesn't have the attribute "checking_id"?

lnbits/decorators.py
```diff
@@ -153,9 +156,11 @@ async def get_key_type(
         raise
 
     try:
-        checker = WalletInvoiceKeyChecker(api_key=token)
-        await checker.__call__(r)
-        wallet = WalletTypeInfo(1, checker.wallet)
+        invoice_checker = WalletInvoiceKeyChecker(api_key=token)
+        await invoice_checker.__call__(r)
+        wallet = WalletTypeInfo(1, invoice_checker.wallet)
```
callebtc, line 161: Why rename variable here?

lnbits/decorators.py
```diff
@@ -134,9 +136,10 @@ async def get_key_type(
     token = api_key_header if api_key_header else api_key_query
 
     try:
-        checker = WalletAdminKeyChecker(api_key=token)
-        await checker.__call__(r)
-        wallet = WalletTypeInfo(0, checker.wallet)
+        admin_checker = WalletAdminKeyChecker(api_key=token)
+        await admin_checker.__call__(r)
+        wallet = WalletTypeInfo(0, admin_checker.wallet)
+        assert wallet.wallet is not None
```
callebtc, line 142: Why rename variable here? Because of same name below? Is that a problem?

lnbits/decorators.py
```diff
@@ -52,12 +53,13 @@ async def __call__(self, request: Request) -> Wallet:
...
             #        Possibly store it in a Redis DB
-            self.wallet = await get_wallet_for_key(key_value, self._key_type)
-            if not self.wallet:
+            wallet = await get_wallet_for_key(key_value, self._key_type)
+            if not wallet:
                 raise HTTPException(
                     status_code=HTTPStatus.UNAUTHORIZED,
                     detail="Invalid key or expired key.",
                 )
+            self.wallet = wallet
```
callebtc, line 62: You set it only if no error happens. Before it was set even if `wallet` is `False` or `None`. Does that matter?

lnbits/decorators.py
```diff
@@ -29,20 +30,20 @@ def __init__(
...
                 description="Wallet API Key - QUERY",
             )
         else:
-            self.model: APIKey = APIKey(
+            key = APIKey(
                 **{"in": APIKeyIn.header},
                 name="X-API-KEY",
                 description="Wallet API Key - HEADER",
             )
-        self.wallet = None
```
callebtc, line 43: Why do you remove this? See below. Is the default value `None`? Does it matter?

lnbits/decorators.py
```diff
@@ -29,20 +30,20 @@ def __init__(
...
+        self.model: APIKey = key
```
callebtc, line 44: You set it only if no error occurs. Before it was set either way. Does that matter?

lnbits/core/views/generic.py
```diff
@@ -149,8 +151,8 @@ async def wallet(
         )
 
     logger.debug(f"Access wallet {wallet_name}{'of user '+ user.id if user else ''}")
-    wallet = user.get_wallet(wallet_id)
-    if not wallet:
+    userwallet = user.get_wallet(wallet_id)
```
callebtc, line 154: Why rename variable?

lnbits/core/views/api.py
```diff
@@ -606,7 +626,7 @@ class ConversionData(BaseModel):
 async def api_fiat_as_sats(data: ConversionData):
     output = {}
     if data.from_ == "sat":
-        output["sats"] = int(data.amount)
+        output["sats"] = data.amount
```
callebtc, line 629: You don't cast the amount here. Why?

lnbits/core/views/api.py
```diff
@@ -387,16 +408,13 @@ async def api_payments_sse(
 async def api_payment(payment_hash, X_Api_Key: Optional[str] = Header(None)):
     # We use X_Api_Key here because we want this call to work with and without keys
     # If a valid key is given, we also return the field "details", otherwise not
-    wallet = None
-    try:
-        if X_Api_Key.extra:
-            logger.warning("No key")
-    except:
-        wallet = await get_wallet_for_key(X_Api_Key)
+    wallet = await get_wallet_for_key(X_Api_Key) if type(X_Api_Key) == str else None
```
callebtc, line 411: Let's hope this is safe. But I've tested it multiple times now, it should be ok.

lnbits/core/views/api.py
```diff
@@ -291,7 +308,7 @@ async def api_payments_pay_lnurl(
                 timeout=40,
             )
             if r.is_error:
-                raise httpx.ConnectError
+                raise httpx.ConnectError("LNURL Callback Connection Error")
```
callebtc, line 311:
```suggestion
                raise httpx.ConnectError("LNURL callback connection error")
```

lnbits/core/views/api.py
```diff
@@ -378,6 +397,8 @@ async def payment_received() -> None:
 async def api_payments_sse(
     request: Request, wallet: WalletTypeInfo = Depends(get_key_type)
 ):
+    # FIXME: wallet.wallet can be None here
+    assert wallet.wallet is not None
```
callebtc, line 401: Do you need these?

Pipfile
```diff
@@ -36,9 +36,11 @@ pycryptodomex = "*"
 
 [dev-packages]
 black = "==20.8b1"
-pytest = "*"
-pytest-cov = "*"
+mock = "*"
```
callebtc, line 39: Need to check whether `Pipfile.lock` gets recreated when installing with `Pipenv`. Test on a fresh instance.

lnbits/bolt11.py
```diff
@@ -23,6 +23,7 @@ class Route(NamedTuple):
 
 class Invoice(object):
     payment_hash: str
+    checking_id: Optional[str] = None
```
callebtc, line 26: `checking_id` should not be in invoice object. Why is this here?

lnbits/core/models.py
```diff
@@ -106,6 +106,8 @@ def from_row(cls, row: Row):
 
     @property
     def tag(self) -> Optional[str]:
+        if self.extra is None:
+            return ""
```
callebtc, line 110: What's happening here?

lnbits/core/views/api.py
dni: ye, wallet can be None here, if we do not assert the code it will break anyway, there are 2 solutions to the problem.

1. we try to make the wallet type Wallet, not Optional[Wallet]
2. we check everywhere before using wallet.wallet if its not None.

lnbits/core/views/api.py
dni: i added tests for that,  in this commit, right? the ones you gave me.

lnbits/wallets/fake.py
dni: that is why i added it to invoice model, as mentioned in your review below.

lnbits/bolt11.py
dni: your are using it on fakewallet            
```python
 hasattr(invoice, "checking_id")
 invoice.checking_id is not None
```

Pipfile
dni: it does, if you delete it before

lnbits/core/models.py
dni: self.extra, can be None, if so ```self.extra.get()``` would error.
so i just check it, and return a empty string if it is ```None``` instead of the tag.

dni: maybe, i can return None? because the return Value is Optional[str]

lnbits/core/views/api.py
dni: mypy complains, if we set amount to float in ConversionData, then cast it to int and afterwards using it as float again.

lnbits/core/views/generic.py
dni: if not mypy complains ```lnbits/core/views/generic.py:154: error: Incompatible types in assignment (expression has type "Optional[Wallet]", variable has type "Wallet")```
i guess ```create_wallet``` above return Optional[Wallet] and ```user.get_wallet()``` return Wallet, mypy doesnt like reusing same variable with different types.

lnbits/decorators.py
dni: no, imo, it doesnt do anything at all, in the __call__ func we try to get it anyways.

lnbits/decorators.py
dni: it is not about the error, mypy doesnt like defining attribute twice.
```lnbits/decorators.py:39: error: Attribute "model" already defined on line 33```

File: lnbits/decorators.py (hunk 8 above)
Comment (dni) [reply lnbits/decorators.py:62]: good question, my guess is, if you do not set self.wallet it is just None, is my assumption wrong?

File: lnbits/decorators.py (hunk 7 above)
Comment (dni) [reply lnbits/decorators.py:142]: its similar to the user wallet, mypy doesnt like if you use the same variable name with different types, in this case checker first was WalletAdminKeyChecker and the it is used later with type WalletInvoiceKeyChecker

File: lnbits/wallets/fake.py (hunk 5 above)
Comment (dni) [reply lnbits/wallets/fake.py:86]: i think this piece of code is terribly broken, it was using data variable which was not defined at all.

Comment (callebtc) [reply lnbits/wallets/fake.py:86]: Do not change models (anywhere in this PR). Stick to the goals of this PR which is type annotation and safety. Please remove this attribute from the invoice.

File: lnbits/decorators.py (hunk 7 above)
Comment (callebtc) [reply lnbits/decorators.py:142]: Gotcha!

File: lnbits/decorators.py (hunk 8 above)
Comment (callebtc) [reply lnbits/decorators.py:62]: Do not change the logic. Stick to the same as before. Please revert.

File: lnbits/core/views/api.py (hunk 12 above)
Comment (callebtc) [reply lnbits/core/views/api.py:629]: I have the feeling that something else might break because of this.

File: Pipfile (hunk 16 above)
Comment (callebtc) [reply Pipfile:39]: If it's not "up to date" it will modify it. Is the Pipfile.lock a fresh one? I usually build them on a fresh `multipass` instance... but I'm sure there are better ways.

File: lnbits/bolt11.py (hunk 17 above)
Comment (callebtc) [reply lnbits/bolt11.py:26]: Oops that's very bad practice. Please do not change the model.

File: lnbits/core/views/api.py (hunk 15 above)
Comment (callebtc) [reply lnbits/core/views/api.py:401]: Option 1 would be a lot more elegant. Is it in the scope of this PR?

File: lnbits/decorators.py (hunk 9 above)
Comment (callebtc) [reply lnbits/decorators.py:43]: Please leave it if you are unsure.

File: lnbits/core/views/api.py (hunk 12 above)
Comment (dni) [reply lnbits/core/views/api.py:629]: can we somehow write a test for this? i'll take a look

File: lnbits/wallets/fake.py (hunk 5 above)
Comment (dni) [reply lnbits/wallets/fake.py:86]: can you look at the code there? i have absolutely no clue why and how this piece of code work.... i think it is completely wrong.

File: lnbits/decorators.py (hunk 8 above)
Comment (dni) [reply lnbits/decorators.py]: e63c4c9f

File: lnbits/decorators.py (hunk 9 above)
Comment (dni) [reply lnbits/decorators.py:43]:
```lnbits/decorators.py:57: error: Incompatible types in assignment (expression has type "Optional[Wallet]", variable has type "None")
```

lnbits/core/views/api.py
dni: 00dba54a
pushed changed with option 1, add a couple of ignore, we really should do a followup PR fixing those ignore types up

lnbits/core/views/api.py
dni: apparently just changing the order fixed the mypy error :) 36cb1eb4

lnbits/core/views/api.py
```diff
@@ -181,11 +181,11 @@ async def api_payments_create_invoice(data: CreateInvoiceData, wallet: Wallet):
 
     lnurl_response: Union[None, bool, str] = None
     if data.lnurl_callback:
-        if "lnurl_balance_check" in data:
+        if hasattr(data, "lnurl_balance_check"):
```
callebtc, line 184: This is not the same thing. 
image

dni: reverted

fusion44 (approved):
LGTM

dni:
ready :)?

Issue #752, opened 2022-07-18
(#725) should write lightning:LNURL instead of LNURL to NFC-Tag


arbadacarbaYK:
for the sake of UX. Cant code, have no idea how to include it though 😅😂

callebtc:
@leesalminen what do you think? 

The current implementation seems to trigger lightning wallets in my phone already but adding a custom `lightning:` URI might be better for more compatibility. Any opinions on this?

arbadacarbaYK:
no then meaning with Lightning:LNURLXXXX

arbadacarbaYK:
Without Lightning:
+ Most wallets are shown as a selection to choose from
- some (Muun and Breez) only react when already opened and  tapped (iphone11) and do not show up in the list
- Having no wallet installed i am redirected to lnbits 

With Lighting:
+ All wallets show up in the list
- Havin no wallet installed nothing happens at all

Some want to be triggered explicitly 👸🏻

Is there a reason why this is not already the case for all QRs shown / to be printed ?

arbadacarbaYK:
not so happy with LNURLw-prefix #761 because LUD17 is so far just supported by only a few wallets whereas lightning: works fine for all 😅

callebtc:
[I completely agree](https://github.com/lnbits/lnbits-legend/pull/762) 🤬

PR #762, opened 2022-07-19
LNURLw: change prefix to `lightning:` until wallets fix their bugs

callebtc:
* [This](https://github.com/fiatjaf/lnurl-rfc/pull/66#issuecomment-868407854=) says you should use `lnurlw:`
* [This](https://github.com/fiatjaf/lnurl-rfc/blob/luds/17.md) says you should use `lnurlw://`
* Wallet of Satoshi parses `lnurlw://abc` to `//abc`
* Breez can handle it all
* I'm gonna use `lightning:` because I got mad.

### DIFF
--- a/lnbits/extensions/withdraw/static/js/index.js
+++ b/lnbits/extensions/withdraw/static/js/index.js
@@ -250,7 +250,7 @@ new Vue({
         })
 
         await ndef.write({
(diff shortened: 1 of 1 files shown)


callebtc:
Fixes #752

PR #791, opened 2022-07-27
Offlineshop - allow for decimals in fiat unit

talvasconcelos:
Offlineshop extension couldn't handle decimals, in items... now it does!


callebtc:
Yo, love this PR, finally! 

Can you run make format and add adjust the migration to conv.py pls! Ideally you'd also update the extension db in mock_data.zip so the tests fail for the next person as well :)

PR #796, opened 2022-07-28
fix black, and pretty formatting errors in current main

dni:
fix black, and pretty formatting errors in current main


arcbtc:
Done via your mypy commit

PR #802, opened 2022-07-28
WIP: Added cache busting for JS and CSS assets

blackcoffeexbt:
Added a cache busting Jinja filter to add ?v=XXXXXX to each CSS and JS asset.

This prevents hard caching of JS/CSS after deploying an update causing users to experience bugs or not see updates.

The cache busting code uses a truncated md5 hash of whatever file appended to the file URL. For example:

/static/js/base.js?v=89d0ts

This md5 hash is cached in memory so persists whilst the LNbits instance is running. This helps prevent performance issues by excess md5 calculations of files.


dni:
i am wondering if there is some sort of middleware or automated way for using that, because now every extension dev would need to include those, that might lead to some dev maybe not knowing, or forgetting about that. which leads to inconsistencies.

do you have an idea for that?

PR #810, opened 2022-07-30
Fix staticfiles for docker install

cryptoteun:
During last updates somehow the staticfiles were broken.
I made this PR based on @callebtc suggested quick fix.


cryptoteun:
Crap, already see two new PR's for this isssue and one for poetry. I forget to check PR's before making this one, only checked main branche :-(

Issue #511, opened 2022-02-07
Add missing description_hash argument that is required for Lightning Address


john-zaprite:
I'm not sure if this is the same as #76 or not, but I'm trying to implement Lightning Address (allowing users to connect to their own nodes: MyNode, Umbrel, etc.) and it looks like LNbits is not including the `description_hash` in the raw data when generating an invoice.

As per the LNurl-pay spec:
> `pr` must have the `h` tag (`description_hash`) set to `sha256(utf8ByteArray(metadata))`.
https://github.com/fiatjaf/lnurl-rfc/blob/legacy/lnurl-pay.md

Is it possible to add this into the data?
Is this a bigger issue, where it depends on the underlying lightning implementation?

dni:
have you tested it this merge solved your issue?

PR #812, opened 2022-07-30
Mega-merge 1: Test invoice creation with description hash (WIP)

callebtc:
# The Mega-merge

This is the first PR of the Mega-merge series. The ultimate goal of the Mega-merge series it to use updated CLN and LND libraries and fix the use of `description_hash` for all wallets (which ultimately enables LNURL).

The Mega-merge series is separated into following parts in order to avoid total chaos and make review easier. All successive PRs have their predecessor merged into them already which is why the number of changed files is deceivingly large for the later PRs (they should go down as we merge one by one). 

Dependency graph:

```
                      ┌───┐
Tests    Refactor   ┌─┤MM3│ CLN
┌────┐     ┌───┐    │ └───┘
│MM1 ├─────►MM2├────►
└────┘     └───┘    │ ┌───┐
                    └─┤MM4│ LND
                      └───┘
```

### Mega-merge 1: Test invoice creation with `description_hash` (PR #812)
The purpose of this (present) PR is to add tests for invoice creation via the endpoint `/api/v1/payments` with `description_hash`. The tests are **expected to fail** but should succeed after the successive fixes in following merges. 

### Mega-merge 2: Refactor `description_hash` handling (PR #814)
This PR refactors how `description_hash` in handled in LNbits. Since most wallets (LND, Eclair, web-wallets) expect the **hash** of the description and others (CLN and FakeWallet/bolt11) expect the description itself, the process of hashing needs to be refactored to the very end of the pipeline into the wallets itself. Before, LNbits was hashing the description in all LNURL operations before calling `create_invoice`. After this PR, LNbits will pass the description and the wallet decides to hash it or not

### Mega-merge 3: Update CLN to use `description_hash` (PR #792)
This PR updates the CLN library and enables the native use of `description_hash`. This PR is the reason why the previous two PR's exist. They are separated so that the previous parts can be merged before the lib is updated. 

### Mega-merge 4: Reenable LND GRPC and update invoice status checking (PR #745)
This PR is not directly related to the previous ones and could be merged separately. Here we updated the RPC stub to the latest version and use `TrackPaymentV2` to check whether an outgoing payment has succeeded or not.

### To be continued...
The end of this series (which I will tackle some other day after we merged all this) will be a new payment checking flow in which the payment will be created in the backend asynchronously and a task is kicked off that regularly checks the status of the payment. In theory, this should get rid of stuck payments. A precondition for this would be that all payments are written to the LNbits database with their `payment_hash` as the `checking_id` and not some post-hoc label that we only know when the payment is sent to the backend. More thoughts on this are written down in the notes of PR #745.

### DIFF
--- a/lnbits/bolt11.py
+++ b/lnbits/bolt11.py
@@ -216,7 +216,11 @@ def lnencode(addr, privkey):
                 expirybits = expirybits[5:]
             data += tagged("x", expirybits)
         elif k == "h":
-            data += tagged_bytes("h", hashlib.sha256(v.encode("utf-8")).digest())
+            # previously: assume that description_hash is a string that needs to be hashed
+            # data += tagged_bytes("h", hashlib.sha256(v.encode("utf-8")).digest())
+            data += tagged_bytes(
+                "h", unhexlify(v)
+            )  # assume that description_hash is a hash already
         elif k == "n":
             data += tagged_bytes("n", v)
         else:
--- a/lnbits/bolt11.py
+++ b/lnbits/bolt11.py
@@ -216,11 +216,7 @@ def lnencode(addr, privkey):
                 expirybits = expirybits[5:]
             data += tagged("x", expirybits)
         elif k == "h":
-            # previously: assume that description_hash is a string that needs to be hashed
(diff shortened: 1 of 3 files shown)


john-zaprite:
Does this solve #511?

PR #814, opened 2022-07-31
Mega-merge 2: Invoice/refactor description hash (WIP)

callebtc:
This PR refactors how the `description_hash` is handled during invoice creation. Before, we calculated the hash of the LNURL metadata and passed it to the wallets. Since some wallets don't want the hash but the actual content, we can't do this and the hashing step has to be refactored into the wallet backends. This is what this PR does. I hope it does not break the extensions, that's the most important thing to test in this PR.

The `regtest/CLightningWallet` test is **expected to fail** since it does not support `description_hash` yet which will happen in the next PR (PR #792).


akovalenko:
@callebtc did you test lnaddress after this merge?
 In the current main ae1a2ac383e210aba8d329801c252f33d9fef05e, I had to revert lnbits/extensions/lnaddress/lnurl.py like this:
```diff
diff --git a/lnbits/extensions/lnaddress/lnurl.py b/lnbits/extensions/lnaddress/lnurl.py
index 49fd30e0..a22c0373 100644
--- a/lnbits/extensions/lnaddress/lnurl.py
+++ b/lnbits/extensions/lnaddress/lnurl.py
@@ -70,9 +70,9 @@ async def lnurl_callback(address_id, amount: int = Query(...)):
                 json={
                     "out": False,
                     "amount": int(amount_received / 1000),
-                    "description_hash": (
+                    "description_hash": hashlib.sha256((
                         await address.lnurlpay_metadata(domain=domain.domain)
-                    ).encode("utf-8"),
+                    ).encode('utf-8')).hexdigest(),
                     "extra": {"tag": f"Payment to {address.username}@{domain.domain}"},
                 },
                 timeout=40,

```
to make it work again.

PR #817, opened 2022-07-31
Adds Cliche as a funding source

arcbtc:
Adds the wonderful lite-node Cliche as a funding source,
https://github.com/fiatjaf/cliche

### DIFF
--- a/lnbits/wallets/cliche.py
+++ b/lnbits/wallets/cliche.py
@@ -5,6 +5,7 @@ from typing import AsyncGenerator, Dict, Optional
 
 import httpx
 from loguru import logger
+from websocket import create_connection
 
 from .base import (
     InvoiceResponse,
@@ -13,7 +14,6 @@ from .base import (
     StatusResponse,
     Wallet,
 )
-from websocket import create_connection
 
 
 class ClicheWallet(Wallet):
@@ -26,7 +26,7 @@ class ClicheWallet(Wallet):
         try:
             ws = create_connection(self.endpoint)
             ws.send("get-info")
-            r =  ws.recv()
+            r = ws.recv()
         except Exception as exc:
             return StatusResponse(
                 f"Failed to connect to {self.endpoint} due to: {exc}", 0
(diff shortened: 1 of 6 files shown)


arcbtc:
I added a new dependency https://github.com/websocket-client/websocket-client
httpx and starlette do not have well supported websocket "clients"

lnbits/wallets/cliche.py
```diff
@@ -0,0 +1,134 @@
...
+            )
+            r = ws.recv()
+        else:
+            ws = create_connection(self.endpoint)
+            ws.send(f"create-invoice --msatoshi {amount*1000} --description {memo}")
+            r = ws.recv()
+        data = json.loads(r)
+        checking_id = None
+        payment_request = None
+        error_message = None
```
callebtc, line 62: You define them here but only use them in line 68.

lnbits/wallets/cliche.py
```diff
@@ -0,0 +1,134 @@
...
+        try:
+            if data["error"]["message"]:
```
callebtc, line 64: Bad code style: you don't know yet whether "error" is present in `data` or whether "message" is present in `data["error"]`.

Always use `data.get("error")` if you don't know if it's in there, instead of `data["error"]`.

Better:
```suggestion
            if data.get("error") is not None and data["error"].get("message"):
```

lnbits/wallets/cliche.py
```diff
@@ -0,0 +1,134 @@
...
+                logger.error(data["error"]["message"])
+                error_message = data["error"]["message"]
+                return InvoiceResponse(
+                    False, checking_id, payment_request, error_message
+                )
+        except:
+            checking_id, payment_request = (
+                data["result"]["payment_hash"],
+                data["result"]["invoice"],
```
callebtc, line 73: Same here, this line might thrown an exception, because you don't know if these fields are present in the dictionary. Use `.get("key") is None` first to be sure, otherwise this will crash when cliche misbehaves and the exception won't be caught in lnbits.

arcbtc: Done

lnbits/wallets/cliche.py
arcbtc: Otherwise 79 throws an error

arcbtc:
Thanks for reviewing!

PR #822, opened 2022-08-01
purge click

arcbtc:


dni:
i think there is nothing wrong with click, do you still want to remove it?

PR #824, opened 2022-08-01
Fix/tipjar desc and satspay

talvasconcelos:
Fix Tipjar description and remove offending quotes that invalidate `JSON.parse`

Also a small fix for Satspay, for when onchain wallet isn't available... mempool was crashing it!

### DIFF
--- a/lnbits/extensions/satspay/views.py
+++ b/lnbits/extensions/satspay/views.py
@@ -2,14 +2,13 @@ from http import HTTPStatus
 
 from fastapi.param_functions import Depends
 from fastapi.templating import Jinja2Templates
-from starlette.exceptions import HTTPException
-from starlette.requests import Request
-from starlette.responses import HTMLResponse
-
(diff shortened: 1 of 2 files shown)


lnbits/extensions/satspay/views.py
```diff
@@ -40,6 +40,8 @@ async def display(request: Request, charge_id: str):
             "request": request,
             "charge_data": charge.dict(),
             "wallet_inkey": inkey,
-            "mempool_endpoint": onchainwallet_config.mempool_endpoint,
+            "mempool_endpoint": onchainwallet_config.mempool_endpoint
+            if onchainwallet_config
```
dni, line 44: imho that if kinda looks confusing, i would heck for that above and use a variable, just for clarity.

dni: close to happy, :D black is not :)

dni:
LGTM

PR #829, opened 2022-08-02
use poetry for development and tests

dni:
this pr includes changes for github workflows, Makefile and docs, to use poetry for development.


callebtc:
Merged in #837

PR #830, opened 2022-08-02
make instant wallet available in custom vouchers

talvasconcelos:
Make the custom designed vouchers QR have the instant wallet capability, when scanned with a normal QR scanner!


dni:
does that work for both now?

talvasconcelos:
> does that work for both now?

yes, initially it would only work for the normal ones. As the QR size is a bit smaller, i tested without the `root_url/?lightning=` but makes sense to have it in!

Issue #847, opened 2022-08-05
empty post requests throwing internal server errors


dni:
the following testcase errors. it should work on all post endpoints

```python
@pytest.mark.asyncio
async def test_what(client):
    url = "/copilot/api/v1/copilot"
    response = await client.post(url)
```

error:
```
sqlalchemy.exc.InterfaceError: (sqlite3.InterfaceError) Error binding parameter 0 - probably unsupported type.
[SQL:
        SELECT *, COALESCE((SELECT balance FROM balances WHERE wallet = wallets.id), 0) AS balance_msat
        FROM wallets
        WHERE adminkey = ? OR inkey = ?
        ]
[parameters: (Security(APIKeyQuery), Security(APIKeyQuery))]
```

it does work on legend server.
![screenshot-1659681456](https://user-images.githubusercontent.com/1743657/183018993-ef2217af-ea9c-40f5-aae1-808b6760e7b3.jpg)

callebtc:
The PR in #848 should include the fix for this issue. If the error is thrown in the crud I guess it should have never reached it.

PR #848, opened 2022-08-05
added test for issue #847 and fixed it!

dni:
testcase to trigger this error as mentioned here
https://github.com/lnbits/lnbits-legend/issues/847

### DIFF
--- a/lnbits/decorators.py
+++ b/lnbits/decorators.py
@@ -199,7 +199,7 @@ async def require_invoice_key(
     api_key_header: str = Security(api_key_header),  # type: ignore
     api_key_query: str = Security(api_key_query),  # type: ignore
 ):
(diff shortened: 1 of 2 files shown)


callebtc:
This PR should ideally include the fix, not just a test that fails.

lnbits/decorators.py
```diff
@@ -201,6 +201,12 @@ async def require_invoice_key(
 ):
     token = api_key_header if api_key_header else api_key_query
```
callebtc, line 202:
```suggestion
    token = api_key_header or api_key_query
```

callebtc: Python ninja

callebtc (changes requested):

PR #851, opened 2022-08-06
feat: implement pre-commit

fusion44:
# Add pre-commit dev package
### What is pre-commit?
[pre-commit](https://pre-commit.com) is a package to manage pre commit hooks.

### Why pre-commit?
A user can install the hook to run certain tools automatically before each commit. This prevents accidentally committing unformatted code to the repository. This potentially could replace our manual calls to black, pretter, etc.

### How to use pre-commit?
`pre-commit` is installed as a dev dependency using poetry. After the install the `pre-commit` command line tool becomes available. This does not automatically install the git hook.

#### Install the git hook
Use `pre-commit install` or `make install-pre-commit` to install the hook.

```bash
$: pre-commit install
pre-commit installed at .git/hooks/pre-commit
```

#### Uninstall the git hook
Use `pre-commit uninstall` to remove the hook.

```bash
$: pre-commit uninstall
pre-commit uninstalled
```

### Gotchas
* when ran as a git hook, only staged files in git will be checked before a commit
* when committing via a gui tool you'll see a somewhat non-descriptive error when trying to commit. The bot applies the fixes and for next commit no error should appear.

### What tools does pre-commit run?
All tools are defined in the [.pre-commit-config.yaml](https://github.com/lnbits/lnbits-legend/blob/9c9a81a9feabc16ae8d1cc606fb9e59e97c443d6/.pre-commit-config.yaml) file.

Tools:
* a few hooks included in pre-commit
* black
* isort
* prettier

### Github action
[pre-commit ci](https://pre-commit.ci) is a Github action that runs automatically on each push event. If it detects a problem in a PR it'll automatically create a PR to fix the problems. [example](https://github.com/fusion44/blitz_api/pull/114/commits)

The CI system will also keep the `.pre-commit-config.yaml` file up to date. [example](https://github.com/fusion44/blitz_api/pull/132)

### ToDo:
* [ ] Update dev documentation
* [ ] Integrate mypy

https://user-images.githubusercontent.com/503952/183246754-ea5a354a-8c1c-4139-8d8b-929bdae179ad.mp4


callebtc (approved):
LGTM

Makefile
```diff
@@ -27,6 +30,10 @@ checkblack:
 checkisort:
 	poetry run isort --check-only .
 
+install-pre-commit:
+	pre-commit install
```
dni, line 34: i think we should run this with `poetry run pre-commit install`

Makefile
```diff
@@ -18,6 +18,9 @@ mypy:
 isort:
 	poetry run isort .
 
+pre-commit:
+	pre-commit run --all-files
```
dni, line 22: i think we should run this with poetry run pre-commit install

PR #855, opened 2022-08-09
do not use invoice mocks for regtest tests

dni:
don't use mock invoice for real wallet tests inside regtest

* introduced `is_regtest` variable into test/helpers.py
* fixed tests for bleskomat
* test_pay_invoice_adminkey added regtest if
* use logger.error for failed backend payments


tests/extensions/bleskomat/test_lnurl_api.py
```diff
@@ -126,11 +127,12 @@ async def test_bleskomat_lnurl_api_action_success(client, lnurl):
...
-    WALLET.pay_invoice.reset_mock()
-    response = await client.get(f"/bleskomat/u?k1={secret}&pr={pr}")
-    assert response.json() == {"status": "OK"}
-    wallet = await get_wallet(bleskomat.wallet)
-    assert wallet.balance_msat == 50000
-    bleskomat_lnurl = await get_bleskomat_lnurl(secret)
-    assert bleskomat_lnurl.has_uses_remaining() == False
-    WALLET.pay_invoice.assert_called_once_with(pr, 2000)
+    if not is_regtest:
+        WALLET.pay_invoice.reset_mock()
```
callebtc, line 131: Why is the whole block in the `if`? Can't this test be run if it's regtest? Should be enough to just `reset_mock()` and go with it. 

Also, this (inside a test case) is not the place to reset the mock. It should be done in `conftest.py`.

tests/core/views/test_api.py
```diff
@@ -153,7 +153,11 @@ async def test_pay_invoice_adminkey(client, invoice, adminkey_headers_from):
     response = await client.post(
         "/api/v1/payments", json=data, headers=adminkey_headers_from
     )
-    assert response.status_code < 300  # should pass
+    if is_regtest:
+        # no self payments
+        assert response.status_code == 520
```
callebtc, line 158: This is a sad test case but true :(

We need to create the invoice somewhere else (in another lnbits instance :) and pay it to be sure this test case does it's job.

callebtc (changes requested):

tests/extensions/bleskomat/test_lnurl_api.py
dni: there is no pay_invoice mock in regtest, so it will fail.

tests/core/views/test_api.py
dni: i agree, can we but those 2 things into an issue, and we will figure it out later? i really need to move forward with boltz extension and i need to PR to make tests.

callebtc:
I think we have to solve this differently. See [here](https://github.com/lnbits/lnbits-legend/blob/main/tests/core/views/test_api.py#L203) for an example of how to check for conditions in tests. I've tried resetting mocks in conftest but they aren't saved globally. 

Closing for now until we find a good way to reset the mocks in different conditions, hope that's ok!

dni:
ok i'm fine with it, i have no idea yet and how we should do it

PR #857, opened 2022-08-09
Backend: Unstuck outgoing payments

callebtc:
This PR is an attempt to solve the **ongoing stuck payment crisis**.  

I have added a lot of notes (for myself) in #745.

### Status
* [x] works with `LndRestWallet`
* [x] works with `LndWallet`
* [x] works with `CLightningWallet`
* [x] works with `EclairWallet`
* [x] works with `LntxbotWallet`
* [x] works with `OpennodeWallet`
* [x] works with `LnpayWallet`
* [x] works with `ClicheWallet`
* [x] works with `SparkWallet`
* [x] works with `LnbitsWalelt`

### Flow
- extract `payment_hash` from `payment_request`
- create temporary payment with `temp_id = payment_hash`
- attempt payment with backend
- backend responds with status (`ok in [True, False, None]` with `None == IN_FLIGHT`) and a `checking_id` 
- if `checking_id == temp_id` and `payment.ok` (**Success**), then update `pending=False` and `fee, preimage` in database entry
- if `checking_id != temp_id` (opennode, lnpay) and `payment.ok`, update `pending=False` etc database entry and **overwrite** `temp_id <- checking_id`
- if `payment.ok == False` (**Failure**), delete payment `temp_id` from database
- if `payment.ok == None` (**Pending**), and `checking_id != temp_id`, overwrite only `temp_id <- checking_id` and let payment sit there for next check
- if `payment.ok == None` and `checking_id == temp_id`, let it sit there

### New code
Most of this is implemented in `services.py`.
* `temp_id` is now simply the `payment_hash`
* `internal_id` is now `internal_{payment_hash}`
* print warning if backend returns other `checking_id` than `payment_hash` (this will be the case for Opennode and LNPay)
* payment is now considered settled if also `payment.ok` instead of only `payment.checking_id` before
* instead of using `create_payment` to create the real payment entry, we use `update_payment_details` now to update the `pending, fee, preimage` and set `new_checking_id` if the wallet returned something different than `payment_hash`
* we don't delete the temporary payments anymore because the don't exist
* `is_uncheckable` now only triggers on `internal_`

### Check
- `wallet.pay_invoice` **must** return `payment.ok in [True, False, None]` for `success, failure, inflight`
- `wallet.pay_invoice` **must** return `payment_hash` as `checking_id`. If it can't (OpenNode, LnPay) it should return what it can to lookup in the later `get_payment_status`.
- If for `payment.is_out`, `PaymentStatus` from `get_payment_status` or `pay_invoice` **ever** returns `PaymentStatus(False)`, the outgoing payment will be considered failed and be deleted. 🚨 **`PaymentStatus(False)` should only be set if you are 100% sure that the payment actually failed!**
- `wallet.get_payment_status` **must** deal with `payment.checking_id` and this should be using the `payment_hash` lookup function of each backend. If not supported, then use `checking_id` the appropriate way.
- `wallet.get_payment_status` **must** return `paid, fee_msat, preimage` if it can.

## Checking payment

Task running a regular payment success check (happens on every browser reload, every 30m, and restart):

### Flow:
* Take `checking_id` and check with backend. If `payment.ok`, then update all payment details, do not overwrite `checking_id`

### New code
* `WALLET.check_payment_status` now returns an updated model `PaymentStatus` which has now three attributes, `paid` (old), `fee` (new), and `preimage` (new).
* To implement this, a new crud called `update_payment_details` was added.
* This is called by the new method `Payment.update_status`.
* `Payment.update_status` is now called instead of `Payment.set_pending` in `Payment.check_pending` and in `services.py:check_transaction_status()`

### Todo
* [x] make sure that all wallets that support lookup by `payment_hash` return it as `checking_id`
* [x] implement updated `WALLET.check_payment_status` for all wallets
* [x] delete warning in `services.py`
* [x] make sure that all `WALLET.pay_invoice` return `PaymentResponse(False, None, None, None, error_message)` upon a failure and not `PaymentResponse(False, "", 0, "", error_message)` or something
* [x] test the `checking_id` overwriting if the backend returns something other than `payment_hash`


callebtc:
Ready for review

lnbits/wallets/opennode.py
```diff
@@ -119,14 +123,16 @@ async def get_payment_status(self, checking_id: str) -> PaymentStatus:
         if r.is_error:
             return PaymentStatus(None)
 
+        data = r.json()["data"]
         statuses = {
             "initial": None,
             "pending": None,
             "confirmed": True,
-            "error": False,
+            "error": None,
```
fiatjaf, line 131: Why `None` here and not `False`?

callebtc: Good question, I did not have to change that. It makes no difference for invoices to be `False` or `None`, I suppose I wanted to be sure that is' an invoice-related error ("failed") instead of some system response "error". Will revert to be sure.

fiatjaf: Nevermind, I think it makes sense.

fiatjaf:
Looks good to me.

PR #873, opened 2022-08-13
fix mypy for extensions

dni:
- explicitly exclude all extensions from mypy
- fix example extension mypy
- fix subdomains extension mypy + 1 type error fixed
- fix mypy discordbot
- fix mypy copilot

### DIFF
--- a/lnbits/extensions/copilot/crud.py
+++ b/lnbits/extensions/copilot/crud.py
@@ -10,7 +10,7 @@ from .models import Copilots, CreateCopilotData
 
 async def create_copilot(
     data: CreateCopilotData, inkey: Optional[str] = ""
-) -> Copilots:
+) -> Optional[Copilots]:
     copilot_id = urlsafe_short_hash()
     await db.execute(
         """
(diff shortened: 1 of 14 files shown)


callebtc:
Why do you exclude the extensions explicitly?

dni:
i exclude them so the new ones willl have to pass mypy, while we slowly keep removing them altogether.

lnbits/extensions/copilot/crud.py
```diff
@@ -71,15 +71,16 @@ async def update_copilot(
...
     items = [f"{field[1]}" for field in data]
-    items.append(copilot_id)
-    await db.execute(f"UPDATE copilot.newer_copilots SET {q} WHERE id = ?", (items))
+    if copilot_id:
+        items.append(copilot_id)
+    await db.execute(f"UPDATE copilot.newer_copilots SET {q} WHERE id = ?", (items,))
     row = await db.fetchone(
         "SELECT * FROM copilot.newer_copilots WHERE id = ?", (copilot_id,)
     )
     return Copilots(**row) if row else None
```
callebtc, line 80: This function doesn't look like the `copilot_id` is actually optional. Won't this just crash if you don't pass it?

dni: changed it, ```copillot_id``` is now not ```Optional[str]```

lnbits/extensions/discordbot/crud.py
```diff
@@ -65,8 +65,9 @@ async def get_discordbot_users(user_id: str) -> List[Users]:
 
 async def delete_discordbot_user(user_id: str) -> None:
     wallets = await get_discordbot_wallets(user_id)
-    for wallet in wallets:
-        await delete_wallet(user_id=user_id, wallet_id=wallet.id)
+    if wallets:
+        for wallet in wallets:
+            await delete_wallet(user_id=user_id, wallet_id=wallet.id)
```
callebtc, line 70: This shouldn't be necessary? The loop only executes if `len(wallets)>0`.

PR #876, opened 2022-08-14
Boltcards extension, round 2

leesalminen:
This takes over where https://github.com/lnbits/lnbits-legend/pull/674 left off.


iWarpBTC:
Thanks for the resurrection. In the last commit the whole extension is adapted to work smoothly with [bolt-nfc-android-app](https://github.com/boltcard/bolt-nfc-android-app) for easily card setup. Because of it there is change in the link that is being written on the cards. Also the db was changed and migration from previous state hasn't been taken care of. Looking forward to all the suggestions and shortcomings revealed.

leesalminen:
This is awesome, thanks so much @iWarpBTC. I will do some testing on this tomorrow.

leesalminen:
Initial testing locally this evening works great using your README. Will keep poking at it tomorrow.

leesalminen:
I've created a handful of cards now, things are looking good! Been testing with the NXP android app and the `boltcard/bolt-nfc-android-app`. Still waiting on my desktop USB reader to test with that.

leesalminen:
Test server for this PR now live at: https://boltcardstest.bitcoinjungle.io

PeterRounce:
The link in the extension 'about box' for `more details` gives a 404 error.
https://github.com/lnbits/lnbits-legend/tree/quart/lnbits/extensions/boltcards

Maybe it could go to this document:
https://github.com/leesalminen/lnbits-legend/blob/ext-boltcards-2/lnbits/extensions/boltcards/README.md

---

Links in the 'about box' are unreadable (dark mode).
Screenshot 2022-08-19 08 03 55

blackcoffeexbt:
@leesalminen 

I've been testing the PR this morning. Looks good but I've run into an error related to failing CMAC verification when trying to make a withdrawal using Breez. The flow works perfectly with the LNbits TPOS.
U
![photo_2022-08-19_09-42-19
ploading photo_2022-08-19_09-42-06.jpg…]()

I've attached some screenshots here. I've used https://github.com/boltcard/boltcard with Breez and it's worked fine so maybe you need something the work around whatever weirdness Breez is doing with the CMAC?

leesalminen:
> @leesalminen
> 
> I've been testing the PR this morning. Looks good but I've run into an error related to failing CMAC verification when trying to make a withdrawal using Breez. The flow works perfectly with the LNbits TPOS.
> 
> I've attached some screenshots here. I've used https://github.com/boltcard/boltcard with Breez and it's worked fine so maybe you need something the work around whatever weirdness Breez is doing with the CMAC?
> 
> photo_2022-08-19_09-42-19
> 
> photo_2022-08-19_09-42-06

We talked about this on Telegram, but it appears the issue is Breez lower-casing the `p` and `c` values. I've pushed a change that re-upper-cases those values on the back end. @iWarpBTC just wanted to share this with you as well for other implementations.

leesalminen:
> The link in the extension 'about box' for `more details` gives a 404 error. https://github.com/lnbits/lnbits-legend/tree/quart/lnbits/extensions/boltcards
> 
> ## Maybe it could go to this document:
> https://github.com/leesalminen/lnbits-legend/blob/ext-boltcards-2/lnbits/extensions/boltcards/README.md
> or this one, which has a lot of information
> https://github.com/boltcard/boltcard/blob/main/README.md
> Links in the 'about box' are unreadable (dark mode). Screenshot 2022-08-19 08 03 55

Thanks for this! I will make a note to fix this pre-launch. We probably want to point the link to `lnbits/lnbits-legend/main`. (can't do until after merge)

thilohille:
I did a small test and setup 2 tags. The first is working fine with breez since the case related fix.
The second tag however is not working. Both have been setup with the bolt-nfc-android-app.
On the second (non-working) tag the default-keys are in place, but as far as i understand, that shouldnt be a problem.
image
i had another error ("unknown card") but didnt take screenshot :-/

leesalminen:
Hey there, thanks for the report here. Taking a look in the database, there is one card in there (ID a46t2unxpa2v5Tx6fgLQTZ) with the k0, k1 and k2 set to `0` as opposed to `00000000000000000000000000000000`.  Wondering if you set it that way in the database by accident?

leesalminen:
Hi again @thilohille, I was able to dig into this some more and confirmed the problem you were having was due to incorrect user input on a card. I've gone ahead and added some more validation on the way in, and some more error handling when using the user input to avoid this from happening in the future. Your card should work now!

lnbits/extensions/boltcards/views_api.py
```diff
@@ -0,0 +1,202 @@
...
+    cards_ids = []
+    for card in cards:
+        cards_ids.append(card.id)
+
+    return [hit.dict() for hit in await get_hits(cards_ids)]
+
+
+# /boltcards/api/v1/scan?p=00000000000000000000000000000000&c=0000000000000000
+@boltcards_ext.get("/api/v1/scan")
+async def api_scane(p, c, request: Request):
```
titusz, line 134: Is `api_scane` a typo or is it intentional?

lnbits/extensions/boltcards/views_api.py
```diff
@@ -0,0 +1,202 @@
...
+    # some wallets send everything as lower case, no bueno
+    p = p.upper()
+    c = c.upper()
+    card = None
+    counter = b""
+
+    # since this route is common to all cards I don't know whitch 'meta key' to use
+    # so I try one by one until decrypted uid matches
+    for cand in await get_all_cards():
```
titusz, line 143: This could become an issue with many registered cards. The bolt server seems to use a single decrypt key for its service: https://github.com/boltcard/boltcard/blob/main/lnurlw_request.go#L66.

An alternative would be to change the API signature to something like: `/api/v1/scan/<card-id>` and and add the <card-id> to the static part of the URI to be programmed to the card. 

I checked how the URIs look on cards programmed by the coincorner app. This is what they look like: `lnurlw://coincorner.io/lnurl/withdraw/auth/563f36e7-faf0-4002-a316-012c29a4759a?picc_data=W2N54LKZYLIKWBXK4HQEFX44UG2VBXKK&cmac=WZEDHGN6TVWO32JT

Not sure but my guess is they use the UUID to map the request to the card.
`

thilohille:
> Hey there @thilohille, thanks for the report here. Taking a look in the database, there is one card in there (ID a46t2unxpa2v5Tx6fgLQTZ) with the k0, k1 and k2 set to `0` as opposed to `00000000000000000000000000000000`. Wondering if you set it that way in the database by accident?

Thanks @leesalminen, the api is now reporting now card found. After changing the alphabetical characters in the cards UID to uppercase it works with breez.
I think the problem ist in boltcards/views_api.py line 148:
`                if card_uid.hex().upper() == cand.uid:`
It assume the UID in lnbits is always uppercase. I copied the uid from the Boltcard NFC Programme, and its lowercase.

lnbits/extensions/boltcards/views_api.py
titusz: I should note that adding a static-uid to the URL would be privacy trade-off. It allows the reader of the card to recognize the individual card. With the current approach only the service can recognize the card.

lnbits/extensions/boltcards/views_api.py
iWarpBTC: That was intentional when there was also api_scan for non encrypted picc data and api_scane was for encrypted. Forgot to change it when deleting the non encrypted function.

lnbits/extensions/boltcards/views_api.py
iWarpBTC: Yeah that was what I considering. I like having common link template for all the card. But it could be  an issue for many card. 

I think it could be optional to have card in the link and therefor 'fast' solo link for every card.

leesalminen: @iWarpBTC I did like the initial PR that had both API options available for use.

leesalminen:
> > Hey there @thilohille, thanks for the report here. Taking a look in the database, there is one card in there (ID a46t2unxpa2v5Tx6fgLQTZ) with the k0, k1 and k2 set to `0` as opposed to `00000000000000000000000000000000`. Wondering if you set it that way in the database by accident?
> 
> Thanks @leesalminen, the api is now reporting now "no card found". After changing the alphabetical characters in the cards UID to uppercase it works with breez. I think the problem ist in boltcards/views_api.py line 148: ` if card_uid.hex().upper() == cand.uid:` It assume the UID in lnbits is always uppercase. I copied the uid from the Boltcard NFC Programme, and its lowercase there. Maybe you should upper() cand.uid as well?

Hi @thilohille, thanks for the additional find here! I've just pushed up a change as you described, performing a `.upper()` on the user input card uid.

lnbits/extensions/boltcards/views_api.py
leesalminen: One possible concern I'm having here is with a project I'm working on with Praia Bitcoin (Bitcoin Beach Brasil) to bulk create Bolt Cards on LNBits (say, 500 at a time). For that use case, it's great as it is now, as it reduces the steps manual steps required to get a card set up. If we do put the card UID back into the output, then we'd potentially have to enter the card UID into the template in the android app when programming the card.

titusz: For clarity, I think there are basically 3 different approaches:

1. SUN parameters only - each card with the same key per service deployment
2. SUN parameters only - each card with custom encryption key
3. SUN parameters + static card-id - each card with custom encryption key

titusz: I actually like option 1 because it still offers link privacy against the reader, easier mass provisioning of cards and fast lookup. At the cost of server key leakage. But if that can leak, individual card keys can also ...

iWarpBTC: @leesalminen The initial PR had unencrypted option, which is not compatible with bolt-nfc-android-app writter. It would be number 4 approach when I follow up on @titusz's list.

I came to the conclusion that a shared encryption key is not much suitable for a service like lnbits. But we can think about how it could be appropriately implemented for such a service.

On the other hand adding a route for the approach 3 would be quick and painless (if I'm not missing anything).

leesalminen: Hey @iWarpBTC, wondering if it'd be possible to put an optional checkbox in the android programming app "append card UID to url" in the "Write NFC" screen there? Then we could offer 2 routes in LNBits: `/api/v1/scan` (as it is now) and `/api/v1/scan/{card_uid}`.

iWarpBTC: I think it would be possible but we don't need that. We can write there the card's UID or even our own ID manually now. We can have anything before `?p=00000000000000000000000000000000&c=0000000000000000`. I hope the link (with or without card's ID) could be part of the QR in the future. Before that, we can display the link that should be written in the card's details in lnbits (maybe with a copy button).

leesalminen: I just pushed up to the branch so that it will allow both ways `/boltcards/api/v1/scan` where it loops over all the cards and decrypts (best privacy, slow if lots of cards in db) and `/boltcards/api/v1/scan/{card_uid}` where it only fetches and decrypts with 1 card record (small potential privacy leak, more efficient)

lnbits/extensions/boltcards/crud.py
```diff
@@ -82,6 +82,14 @@ async def get_card(card_id: str) -> Optional[Card]:
 
     return Card.parse_obj(card)
 
+async def get_card_by_uid(card_uid: str) -> Optional[Card]:
+    row = await db.fetchone("SELECT * FROM boltcards.cards WHERE uid = ?", (card_uid,))
```
thilohille, line 87: Is the select case insensitive?

lnbits/extensions/boltcards/views_api.py
leesalminen: function renamed

lnbits/extensions/boltcards/views_api.py
iWarpBTC: I'm just wondering if revealing the UID and guessing the counter could endanger the encryption key... Unfortunately I'm not that familiar with AES...

lnbits/extensions/boltcards/crud.py
leesalminen: Card UIDs should be stored in the db as uppercase. So, I just pushed a change that uppercases the card UID when creating/saving/searching by it, making it case insensitive from a user's perspective.

lnbits/extensions/boltcards/views_api.py
leesalminen: > I'm just wondering if revealing the UID and guessing the counter could endanger the encryption key... Unfortunately I'm not that familiar with AES... But probably I'm being too paranoid here.

I'm definitely not qualified to make anything approaching an authoritative statement on that.

In an abundance of caution, we could change this to use LNBit's internal card ID (not used in the encryption) instead.

The only downside I see to that is that when a user programs the cards, they'll need to use this opaque identifier only found in LNBits. A future update to the Android app though as you mention could resolve that UX concern by embedding it into the QR code they scan.

leesalminen: Pushed this

leesalminen: Crap, I forgot that breez wallet lowercases everything in the string, so lookup on the card_id fails. I'm back to card_uid?

leesalminen:
OK, test results I've tried since the latest updates:

- Breez iOS 
Tried both with and without the {card_uid} appended to the lnurlw:// string. Both work

- BTCPayServer Android
Tried both with and without the {card_uid} appended to the lnurlw:// string. Both work

- LNBits TPOS Android
Tried both with and without the {card_uid} appended to the lnurlw:// string. Both work

- Bitcoin Jungle Wallet iOS & Android
Tried both with and without the {card_uid} appended to the lnurlw:// string. Both work

lnbits/extensions/boltcards/views_api.py
iWarpBTC: Ouch! I did not realize card_id is case sens. So it is card_uid or a new dedicated id. Up to you I guess.

titusz:
Also tested with Breez iOS and TPOS without {card_uid}. 👍

lnbits/extensions/boltcards/views_api.py
leesalminen: > Ouch! I did not realize card_id is case sens. So it is card_uid or a new dedicated id. Up to you I guess.

I put the card_uid back in place.

arcbtc:
Giving bit of a refactor to make easier to use, progress can be tracked here. https://github.com/leesalminen/lnbits-legend/pull/1

lnbits/extensions/boltcards/lnurl.py
```diff
@@ -43,26 +43,26 @@
...
     # some wallets send everything as lower case, no bueno
     p = p.upper()
     c = c.upper()
     card = None
     counter = b""
     try:
-        card = await get_card_by_uid(card_uid)
-        card_uid, counter = decryptSUN(bytes.fromhex(p), bytes.fromhex(card.k1))
+        card = await get_card(card_id)
+        card_id, counter = decryptSUN(bytes.fromhex(p), bytes.fromhex(card.k1))
```
leesalminen, line 55: `decryptSUN` doesn't return a `card_id`, it returns a `card_uid`.

lnbits/extensions/boltcards/lnurl.py
```diff
@@ -43,26 +43,26 @@
...
 
-        if card.uid.upper() != card_uid.hex().upper():
+        if card.uid.upper() != card_id.hex().upper():
```
leesalminen, line 57: If we're dropping card_uids entirely, we can't verify the card_uid through the encryption.

lnbits/extensions/boltcards/lnurl.py
```diff
@@ -43,26 +43,26 @@
...
             return {"status": "ERROR", "reason": "Card UID mis-match."}
     except:
         return {"status": "ERROR", "reason": "Error decrypting card."}
 
     if card == None:
         return {"status": "ERROR", "reason": "Unknown card."}
 
-    if c != getSunMAC(card_uid, counter, bytes.fromhex(card.k2)).hex().upper():
+    if c != getSunMAC(card_id, counter, bytes.fromhex(card.k2)).hex().upper():
```
leesalminen, line 65: This will fail entirely, as the NTAG 424 devices are using card_uid. You can't make the cards do crypto with a user defined card_id.

lnbits/extensions/boltcards/crud.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+    card = await get_card(card_id)
+    assert card, "Newly created card couldn't be retrieved"
+    return card
+
+
+async def update_card(card_id: str, **kwargs) -> Optional[Card]:
+    if "is_unique" in kwargs:
+        kwargs["is_unique"] = int(kwargs["is_unique"])
+    if "uid" in kwargs:
```
leesalminen, line 47: If we really are dropping card_uid entirely, then we should get rid of this.

lnbits/extensions/boltcards/crud.py
```diff
@@ -0,0 +1,177 @@
...
+    row = await db.fetchone("SELECT * FROM boltcards.cards WHERE id = ?", (card_id,))
+    if not row:
+        return None
+
+    card = dict(**row)
+
+    return Card.parse_obj(card)
+
+
+async def get_card_by_uid(card_uid: str) -> Optional[Card]:
```
leesalminen, line 86: If we really are dropping card_uid we should drop this function

arcbtc:
When deleting a card the other records like refunds and hits become inaccessible, so I added a disable card function, for users who want the card not to work, but also want to keep all their records

lnbits/extensions/boltcards/lnurl.py
```diff
@@ -0,0 +1,204 @@
...
+    update_card_counter,
+    update_card_otp,
+)
+from .models import CreateCardData
+from .nxp424 import decryptSUN, getSunMAC
+
+###############LNURLWITHDRAW#################
+
+# /boltcards/api/v1/scan?p=00000000000000000000000000000000&c=0000000000000000
+@boltcards_ext.get("/api/v1/scan/{card_uid}")
```
titusz, line 44: The `p` query parameter is the encrypted `card_uid` together with a counter. Is it a good idea to expose the unencrypted `{card_uid}` in the path parameter (in the static part of lnurlw)? This could be a security issue. I would recommend to use another (database internal) id as path parameter instead of the card_uid!

PeterRounce:
I would recommend against having any static card identifier in the lnurlw:// link.

It is a privacy issue for card users.

PeterRounce:
Reference: https://github.com/boltcard

In the bolt card service code, we now have a mechanism for discovering uid and counter values on the first time the card is presented. This is designed to be a good user experience (no need to enter a UID or counter value) and computationally efficient.


The sequence of provisioning a new card will be:

- `createboltcard` is run on the server and presents a QR for the provisioning link including a one time use code
- boltcard app scans the QR and retrieves k0-k4, the lnurlw link base(, protocol name and version)
- boltcard app programs up the card
- boltcard app reads the card and hits the service endpoint
  - this functions as a test of the card and the service endpoint
  - this allows the service to find the card record using the CMAC key and update the uid and counter value fields

I hope this is also useful here to preserve user privacy and optimise the UX and server processes.

titusz:
> I would recommend against having any static card identifier in the lnurlw:// link.
> 
> It is a privacy issue for card users.

Agreed. But doesn´t that require a single global constant `AES_DECRYPT_KEY` on the service? And that `AES_DECRYPT_KEY` also has to be shared with every user that wants to provision a card?

lnbits/extensions/boltcards/lnurl.py
leesalminen: Added.

Issue #872, opened 2022-08-13
ERROR during `DELETE /lnurlp/api/v1/links/xxx`


callebtc:
```sh
2022-08-13 13:28:18.04 | INFO | 185.220.101.33:0 - "DELETE /lnurlp/api/v1/links/xxxx HTTP/1.1" 204
2022-08-13 13:28:18.04 | ERROR | lnbits.app:basic_error:194 | handled error
2022-08-13 13:28:18.04 | ERROR | lnbits.app:basic_error:195 | ERROR:
Traceback (most recent call last):
  File "/home/.cache/pypoetry/virtualenvs/lnbits-3PdPdHve-py3.8/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
    await self.app(scope, receive, sender)
  File "/home/.cache/pypoetry/virtualenvs/lnbits-3PdPdHve-py3.8/lib/python3.8/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/home/.cache/pypoetry/virtualenvs/lnbits-3PdPdHve-py3.8/lib/python3.8/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/home/.cache/pypoetry/virtualenvs/lnbits-3PdPdHve-py3.8/lib/python3.8/site-packages/starlette/routing.py", line 670, in __call__
    await route.handle(scope, receive, send)
  File "/home/.cache/pypoetry/virtualenvs/lnbits-3PdPdHve-py3.8/lib/python3.8/site-packages/starlette/routing.py", line 266, in handle
    await self.app(scope, receive, send)
  File "/home/.cache/pypoetry/virtualenvs/lnbits-3PdPdHve-py3.8/lib/python3.8/site-packages/starlette/routing.py", line 65, in app
    response = await func(request)
  File "/home/.cache/pypoetry/virtualenvs/lnbits-3PdPdHve-py3.8/lib/python3.8/site-packages/fastapi/routing.py", line 227, in app
    raw_response = await run_endpoint_function(
  File "/home/.cache/pypoetry/virtualenvs/lnbits-3PdPdHve-py3.8/lib/python3.8/site-packages/fastapi/routing.py", line 160, in run_endpoint_function
    return await dependant.call(**values)
  File "/lnbits-legend/lnbits/extensions/lnurlp/views_api.py", line 139, in api_link_delete
[...]

Issue #815, opened 2022-07-31
LNURLw: Exception when deleting a withdrawal link


callebtc:
Reproduce: 
* Create LNURLw
* Delete it

Stack trace:
```sh
...
File "/lnbits/extensions/withdraw/views_api.py", line 131, in api_link_delete
starlette.exceptions.HTTPException
During handling of the above exception, another exception occurred:
...
  File "/pypoetry/virtualenvs/lnbits-3PdPdHve-py3.9/lib/python3.9/site-packages/uvicorn/protocols/http/httptools_impl.py", line 533, in send
    raise RuntimeError("Response content longer than Content-Length")
RuntimeError: Response content longer than Content-Length
2022-07-31 11:13:54.27 | ERROR | Exception in ASGI application
ERROR:uvicorn.error:Exception in ASGI application
...
  File "/usr/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type RuntimeError is not JSON serializable
2022-07-31 11:13:56.12 | ERROR | lnbits.app:basic_error:182 | handled error
2022-07-31 11:13:56.12 | ERROR | lnbits.app:basic_error:183 | ERROR:
```

leesalminen:
I pushed a fix for this issue in #877 . Looks like it was copy-pasta from the lnurlp extension, same issue, same fix.

PR #877, opened 2022-08-14
resolves issue #872 & #815, error deleting lnurlp/lnurlw links

leesalminen:
This PR resolves the issue reported in #872 and #815, error deleting lnurlp/lnurlw links.


callebtc:
LGTM. Good find, thank you!

PR #884, opened 2022-08-15
Update installation.md

joedavison:
I needed to run this additional command to install python3.9-specific distutils. Without this, installation of several poetry dependencies (idna, pyparsing, sniffio) failed with message: ModuleNotFoundError: No module named 'distutils.cmd'


callebtc:
Thank you, very useful, I've ran into the same problem once. 

One request: Could you add the package just to the line above where we install python3.9? Just a bit more clean that way. I'd merge this immediately after.

joedavison:
Done, thank you.

callebtc:
LGTM

PR #889, opened 2022-08-16
add pending column to CSV export on wallet page, issue #887

### LINKED ISSUE #887: Wallet CSV export: add a column / field to report on "payment status"
### Situation
The wallet UI shows `payment status, memo, date, amount (sats), fee (sats)`.
The CSV export does not have a `payment status`, hence in accounting, a critical factor (paid or pending) is missing

### Proposal
Add `payment status` to the CSV export

dni:
Closes #887

### DIFF
--- a/lnbits/core/static/js/wallet.js
+++ b/lnbits/core/static/js/wallet.js
@@ -668,6 +668,9 @@ new Vue({
       })
     },
     exportCSV: function () {
+      // status is important for export but it is not in paymentsTable
+      // because it is manually added with payment detail link and icons
+      // and would cause duplication in the list
       let columns = this.paymentsTable.columns
       columns.unshift({
         name: 'pending',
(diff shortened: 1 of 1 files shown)


callebtc:
Please `make format` before we can merge!

callebtc:
Two questions: What's the difference between pending and isPaid? Why isn't it part of `this.paymentsTable.columns` already?

dni:
_What's the difference between pending and isPaid?_
they are equivalent, i remove isPaid.

_Why isn't it part of this.paymentsTable.columns already?_
 my guess is, they did not put it into paymentsTable, because they needed more control of the rendering of the column / payment details link / icon.

lnbits/core/static/js/wallet.js
```diff
@@ -668,7 +668,14 @@ new Vue({
       })
     },
     exportCSV: function () {
-      LNbits.utils.exportCSV(this.paymentsTable.columns, this.payments)
+      let columns = this.paymentsTable.columns
```
callebtc, line 671: Please add a short comment here why you're doing this, nobody will understand when looking at it in two years.

PR #899, opened 2022-08-17
fix migrations, add documentation, hopefully fix workflow

dni:
* make `make migration` usable for devs and also use it in github workflow
* added better documentation for the postgres migration port
* added the hint to also update `databases.sqlite3`
* changed datafolder and database name to `migration-data` and `migration`, for extra safety so you dont accidently overrite the wrong data
* fixed migration for `ext_invoices` and `ext_watchonly`


callebtc:
LGTM

PR #908, opened 2022-08-18
WIP: Diagon alley Extension

talvasconcelos:
WIP extension for a marketplace, and individual stores (stalls)! User can create a few stores, with different products and the have them aggregated on a "Marketplace"!

At this point it should allow for the creation of all the workflow, except for the marketplace (that aggregates all stores). Creating shipping zones, stores, products should be working. The store frontend should allow to place an order and pay it!  

Will come back to this description for some better explanation and a To Do list!


talvasconcelos (commented):
Have you tested if keys get generated for costumer with these changes?

Maybe the generate keys function could be refactored, and use query params instead but the way it was set, it needed the order payment_hash (identifier) to add the pubkey to order entry in DB when costumer calls it!

arcbtc:
ready to merge? This looks ready. When I tested everything works. LGTM

PR #914, opened 2022-08-19
Allow more than 2-nd level domain in incoming ln-addresses

akovalenko:
Make /api/v1/lnurlscan  no longer require exactly 2 parts for name's domain split by "."

Payment to addresses like anton@bits.my-ns.me is now possible.


callebtc:
Thank you, excellent find!

PR #915, opened 2022-08-19
Add docs for hosting LNBits on Fly.io (for free)

leesalminen:
Fly.io is a neat hosting provider that takes a docker container and runs it for you. They have a generous free tier that is enough to run LNBits w/ sqlite. This PR is just some docs on how to do it.


arcbtc:
Clever stuff!
I had to do `/home/ben/.fly/bin/flyctl`

leesalminen:
Oh right, after the Fly CLI install, it spits out some text you're supposed to copy/paste to get it into your $PATH. I'll note that in the docs.

Issue #897, opened 2022-08-17
Fix main wallet view layout for mobile screen sizes


callebtc:
No comment.

<img src="https://user-images.githubusercontent.com/93376500/185116072-f6741bf2-29c5-456d-8a4d-22b0edcf3837.jpg" width="400" height="790">

callebtc:
in LNURLp view, move pink buttons (NFC and printing)

<img src="https://user-images.githubusercontent.com/93376500/185368901-5ba9425f-fadb-4a05-aaa7-f5844c41e078.jpg" width="400" height="790">

dni:
> in LNURLp view, move pink buttons (NFC and printing)
> 
> <img alt="" width="400" height="790" src="https://user-images.githubusercontent.com/93376500/185368901-5ba9425f-fadb-4a05-aaa7-f5844c41e078.jpg">

this one got fixed by you right?

dni:
> No comment.
> 
> <img alt="" width="400" height="790" src="https://user-images.githubusercontent.com/93376500/185116072-f6741bf2-29c5-456d-8a4d-22b0edcf3837.jpg">

my solution for that would be do not let memo break and give it the space it needs for 1 line, that means you just need to scroll to the right more, but it is 10 times more clear.

![screenshot-1660985475](https://user-images.githubusercontent.com/1743657/185737176-0dcc1d5a-fc19-4281-bc82-f69fc00afd60.jpg)

dni:
another addition, don' t use columns for payment-details

![screenshot-1660988731](https://user-images.githubusercontent.com/1743657/185740073-292adad3-821a-41de-98e3-450d5712fdc1.jpg)
![screenshot-1660988722](https://user-images.githubusercontent.com/1743657/185740074-3817fbec-d34b-4919-8cad-efd396747ffa.jpg)

i leave this pr as is, i think its subjectively better than before, but i do not want get into any design discussion :).

dni:
https://github.com/lnbits/lnbits/pull/1732

PR #918, opened 2022-08-20
fix issue: #897, ugly memo in payments list

dni:
Fixes: #897

### DIFF
--- a/lnbits/core/templates/core/wallet.html
+++ b/lnbits/core/templates/core/wallet.html
@@ -165,11 +165,7 @@
                   <q-tooltip>Pending</q-tooltip>
                 </q-icon>
               </q-td>
-              <q-td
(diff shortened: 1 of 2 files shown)


dni:
screenshot-1660985475

memo is now a 1 liner

dni:
another addition, don' t use columns for payment-details
### before
screenshot-1660987304

### after
screenshot-1660988731
screenshot-1660988722

i leave this pr as is, i think its subjectively better than before, but i do not want get into any design discussion :).

arcbtc:
thats a long phone you have there!

arcbtc:
Be good to get this merged, but it needs compacting for a smaller screen Add some `v-if="gt-md"`, or something similar to some padding elements to tuck it all in 👍.

PR #928, opened 2022-08-25
Allow creation of invoices with msat precision

zebra-lucky:
Hi!

These changes allow create invoices with msat precision.
No gui changes, except allow wallet balance show fractional part of sats.
Intended for use via API calls.

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -61,6 +61,7 @@ async def create_invoice(
     webhook: Optional[str] = None,
     internal: Optional[bool] = False,
     conn: Optional[Connection] = None,
+    use_msat: bool = False,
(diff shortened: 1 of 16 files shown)


zebra-lucky:
Looking why regtest fails.

zebra-lucky:
rebased on current `main`

HamishMacEwan:
Given the many uncertainties even in sats denominated invoices (withholding thresholds, channel close reservations, etc, etc.) what is the benefit in permitting user specification of amounts as, currently admittedly, insignificant as millisats?

LNbits is not designed for either very large, or microscopically tiny, amounts.  It seems likely this will cause more trouble than it's worth.

zebra-lucky:
Rebased on current `main`

Issue #907, opened 2022-08-18
Frontend: Do not show NFC button if browser doesn't support it


callebtc:
JS should be able to check whether the browser supports NFC or not before the users presses the button. Applies to tpos, lnurlp, lnurlw extensions. Does this make sense @leesalminen?

<img width="400" alt="image" src="https://user-images.githubusercontent.com/93376500/185422250-5714ab2c-9cb5-4334-a2d8-d04a82f5481c.png">

leesalminen:
Yes.

leesalminen:
Can you assign this to me

leesalminen:
PR opened.

arcbtc:
This error is fine. Users know the thing does NFC, but the their device does not support it

arcbtc:
Its a good error. Getting rid of the button doesnt make sense

PR #934, opened 2022-08-30
Hide NFC button if device does not support NFC

leesalminen:
This PR resolves issue #907 . We want to hide the NFC button if the device does not support NFC in the lnurlp, lnurlw, and tpos extensions.


arcbtc:
Disable and a "This device does not support NFC" tooltip is better as people then still now its possible

arcbtc:
We had a big long discussion with some merchants and decided its better to have the nfc button present, so it helps guide people to a browser that does work

PR #943, opened 2022-09-05
Fix/tpos dismiss btn

talvasconcelos:
Removed the timeout on the green checkmark after payment and added a dismissable notification for the payment. Either dismiss the notification to remove the checkmark (and the notification), or even if the checkmark goes away (it closes if the screen is clicked) the notification stays and needs to be manually clicked!

pos


talvasconcelos:
Hotfix for TPoS failing to show QR invoice on previous created TPoS's without tip options!

arcbtc:
Can you just make the tick a button, and we press that?

PR #954, opened 2022-09-09
lndhub: remove delays

JssDWt:
Remove the delays in gettxs and getuserinvoices calls to support wallets that use the lndhub extension.


callebtc:
LGTM

PR #968, opened 2022-09-14
Holdinvoice

pseudozach:
* add holdinvoice endpoints for lndrest
* add webhooks for holdinvoice updates e.g. `state=ACCEPTED` or `state=CANCELED` as per LND REST API

This is useful when apps/extensions want to use hold invoice functionality instead of custodially accepting payments and then refunding them.

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -30,6 +30,7 @@ from .crud import (
     check_internal,
     create_payment,
     delete_payment,
+    get_standalone_payment,
     get_wallet,
     get_wallet_payment,
     update_payment_details,
@@ -155,6 +156,19 @@ async def cancel_hold_invoice(
     )
     if not ok:
         raise InvoiceFailure("Unexpected backend error.")
+
+    return True
+
+async def subscribe_hold_invoice(
+    *,
+    wallet_id: str,
+    payment_hash: bytes,
+    conn: Optional[Connection] = None,
+) -> bool:
+    payment = await get_standalone_payment(payment_hash, incoming=True)
+    asyncio.create_task(WALLET.hold_invoices_stream(
+        payment_hash=payment_hash, webhook=payment.webhook
+    ))
 
     return True
 
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -1,4 +1,5 @@
 import asyncio
+from distutils.log import error
 import json
 from binascii import unhexlify
 from io import BytesIO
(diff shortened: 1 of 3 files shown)


callebtc:
Nice! Concept ACK. Lot's of code, maybe I'd make sense to refactor it to a new file. Do you know how to add this for LND gRPC and CLN as well?

pseudozach:
> Nice! Concept ACK. Lot's of code, maybe I'd make sense to refactor it to a new file. Do you know how to add this for LND gRPC and CLN as well?

Yes, adding support for land grpc should be easy, CLN I need to check but I'm ok to keep adding support for other backends.

My biggest issue is that I was unable to send webhook from the subscription response outside of the lndrest code.
Let me know if you can think of a better structure. This is the only way I could make it work.
And btw it's used in production which is why I wrote it and open to improving.

callebtc:
Picked up this PR again. Did you have a chance to look at LND GRPC yet?

pseudozach:
@callebtc unfortunately no, couldn't get around to it yet. But it's been working like a champ in production for a while, very happy about that.

callebtc:
Hi! Looked at this again. Sorry for the slow response. Could you please run `make format`, check for `mypy` errors and resolve the merge conflict?

blakejakopovic:
I'm looking forward to having hold invoices supported for https://github.com/blakejakopovic/lightning_rs_webhook. Hold invoices can be useful for Nostr services, as they allow for failed requests to have payments cancelled without a refund access, or issues around high fee paths from suer supplied refund invoices, etc. I think we can also use them for split payment models as well.

@pseudozach is there something I can do to assist finalising this PR?

blckbx:
> @pseudozach is there something I can do to assist finalising this PR?

+1 we'd love to switch to hodl invoices as well

ziggie1984:
what is blocking this one ? Going to use it now for production at tunnesats.

lnbits/core/services.py
```diff
@@ -93,6 +95,82 @@ async def create_invoice(
 
     return invoice.payment_hash, payment_request
 
+async def create_hold_invoice(
+    *,
+    wallet_id: str,
+    amount: int,  # in satoshis
+    memo: str,
+    hash: Optional[bytes] = None,
```
ziggie1984, line 103: the hash is not allowed to be optional.

dni:
it would need some love :) another rebase and probably tests

dni:
closed in favor of https://github.com/lnbits/lnbits/pull/2869

PR #992, opened 2022-09-20
Move comments above variables in .env

JssDWt:
Some env variables had an inline comment next to them. The comment is included in the value though, not considered a comment. Some numbers and booleans are considered invalid and the ad space became a list of 2 values: `# csv ad image filepaths or urls` and `extensions can choose to honor`. Moving the comments above the variables fixes this.


fusion44:
Had this fix also included in PR #912.

dni:
thanks for the PR but closing in favor of https://github.com/lnbits/lnbits/pull/912

PR #553, opened 2022-03-07
Setup UI and admin extension

arcbtc:
Adds a setup UI/admin extension bool to .env, which if true transposes all .env variables into db and creates a UI experience for installing and managing the LNbits install

PR #997, opened 2022-09-21
Conv/final admin UI

talvasconcelos:
It should work now! Pulled main and fixed the conflicts!


dni:
use this PR instead now https://github.com/lnbits/lnbits/pull/553

PR #1002, opened 2022-09-23
API auditor

iWarpBTC:
Just a feature that helps a LNbits operator to keep track of the total balance of his public instance.

### DIFF
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -671,7 +671,10 @@ async def api_auditor(wallet: WalletTypeInfo = Depends(get_key_type)):
     total_balance = await get_total_balance()
     error_message, node_balance = await WALLET.status()
 
-    delta = node_balance - total_balance
+    if not error_message:
+        delta = node_balance - total_balance
+    else:
+        node_balance, delta = None, None
 
     return {
         "node_balance_msats": node_balance,
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -2,6 +2,7 @@ import asyncio
 import binascii
 import hashlib
 import json
+import time
 from http import HTTPStatus
 from io import BytesIO
 from typing import Dict, List, Optional, Tuple, Union
@@ -27,7 +28,7 @@ from lnbits.decorators import (
     require_invoice_key,
 )
 from lnbits.helpers import url_for, urlsafe_short_hash
-from lnbits.settings import LNBITS_ADMIN_USERS, LNBITS_SITE_TITLE
+from lnbits.settings import LNBITS_ADMIN_USERS, LNBITS_SITE_TITLE, WALLET
 from lnbits.utils.exchange_rates import (
     currencies,
     fiat_amount_as_satoshis,
@@ -39,6 +40,7 @@ from ..crud import (
     create_payment,
     get_payments,
     get_standalone_payment,
+    get_total_balance,
     get_wallet,
     get_wallet_for_key,
     save_balance_check,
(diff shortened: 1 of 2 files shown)


callebtc:
Condept ACK
Approach ACK but would like to hear more voices on this

dni:
i think it is a good idea, 

it would also very nice to ad it to the admin ext / AdminUI https://github.com/lnbits/lnbits/pull/553
so you can see it there

dni (approved):
looks good to me

lnbits/core/views/api.py
```diff
@@ -657,3 +659,23 @@ async def _generator(stream: BytesIO):
...
+
+@core_app.get("/api/v1/audit/")
+async def api_auditor(wallet: WalletTypeInfo = Depends(get_key_type)):
+    if wallet.wallet.user not in LNBITS_ADMIN_USERS:
+        raise HTTPException(
+            status_code=HTTPStatus.FORBIDDEN, detail="Not an admin user"
+        )
+
+    total_balance = await get_total_balance()
+    error_message, node_balance = await WALLET.status()
```
motorina0, line 672: - should check the value of `error_message`
- if the node cannot be accessed then (maybe) an exception should be raised

iWarpBTC: Now if node is down it just gives zero in node_balance_msats. Hesitating if it's better to see an exception or at least lnbits_balance_msats. I'm open for opinions.

lnbits/core/views/api.py
```diff
@@ -657,3 +659,23 @@ async def _generator(stream: BytesIO):
...
+
+
+@core_app.get("/api/v1/audit/")
+async def api_auditor(wallet: WalletTypeInfo = Depends(get_key_type)):
+    if wallet.wallet.user not in LNBITS_ADMIN_USERS:
+        raise HTTPException(
+            status_code=HTTPStatus.FORBIDDEN, detail="Not an admin user"
+        )
+
+    total_balance = await get_total_balance()
```
motorina0, line 671: If there are no balances, then `total_balance` is `None`. It should be `zero`.
Can be reproduced with `LNBITS_BACKEND_WALLET_CLASS=FakeWallet`

```
delta = node_balance - total_balance
TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'
```

motorina0:
LGTM

callebtc:
LGTM

lnbits/core/views/api.py
callebtc: I think this is fine as is and edge cases are out of scope for this PR. Eventually this can be used to build a security kill switch for LNbits which I'm very much looking forward to.

callebtc (approved):

lnbits/core/views/api.py
```diff
@@ -657,3 +659,23 @@ async def _generator(stream: BytesIO):
...
+    error_message, node_balance = await WALLET.status()
+
+    delta = node_balance - total_balance
```
callebtc, line 674:
```suggestion
    delta = node_balance - total_balance if not errror_message else None
```

lnbits/core/views/api.py
```diff
@@ -657,3 +659,23 @@ async def _generator(stream: BytesIO):
...
+
+    return {
+        "node_balance_msats": node_balance,
```
callebtc, line 677:
```suggestion
        "node_balance_msats": node_balance if not errror_message else None,
```

lnbits/core/views/api.py
callebtc, line 674:
```suggestion
    if not error_message:
        delta = node_balance - total_balance
    else:
        node_balance, delta = None, None
```

callebtc (changes requested):
See suggestion for the error case

Issue #949, opened 2022-09-08
all keys should be set on the NXP NTAG 424 DNA card


PeterRounce:
[the bolt card extension readme](https://github.com/lnbits/lnbits-legend/blob/b3093cafdbc20256abd884c2b495c34218cd313a/lnbits/extensions/boltcards/README.md) states "For this extension only two keys need to be set" and "Card Auth key (K0) can be omitted" however this is not compliant with the recommendations by the chip manufacturer NXP

[NTAG 424 DNA – Secure NFC T4T compliant IC](https://www.nxp.com/docs/en/data-sheet/NT4H2421Gx.pdf)

```
8.2.4.2 AppKey

Remark: It is highly recommended to change all 5 keys at personalization, 
even if not all keys are used in the application.
```

PeterRounce:
this is fixed in 
https://github.com/lnbits/lnbits/pull/1005/files#diff-c50047d4b10cf734a8f32f2f3fff7194fe093b26beefdb71f07b382b2b9e8aa9

PR #1005, opened 2022-09-23
Boltcard ext upgrade

iWarpBTC:
- updated readme (also responsing to issue #949)

- added webhook for notifications


arcbtc:
I wonder if it makes sense to add a webhook option to `pay_invoice` in `core.services` then this extension and any other could make use of it?

iWarpBTC:
I believe that's a good idea. Could it look like something like in last commit?

iWarpBTC:
> I wonder if it makes sense to add a webhook option to `pay_invoice` in `core.services` then this extension, and any other, could make use of it?

I believe that's a good idea. Could it look like something like in the last commit?

iWarpBTC:
It should be ready for a review. I'm wondering what you say to passing the webhook either as a string or as a tuple. Wouldn't be better to just split it to two params?

iWarpBTC:
I've just reduced this PR which I had let to grown in a nasty way. 

Remain
-  the ability to wipe a card with Boltcard NFC Card Creator (since v0.1.2)
-  revised card delete flow (with force downloading keys in a wipe json)
-  reason of payment fail is returned to a charging service
-  some import clean up
-  readme update. 

I keep the webhook updates for another PR. 

Hope this help to move this forward.

PR #1007, opened 2022-09-24
Remove unused proxy_fix

AaronDewes:
With uvicorn, you can use [CLI flags](https://www.uvicorn.org/deployment/#running-behind-nginx) to achieve the same. These could probably also be added to `lnbits/server.py` (I'm still manually running Lnbits using the uvicorn command, so I just do that normally).

proxy_fix does not seem necessary anymore.


callebtc:
Thank you, code removal is beautiful!

PR #703, opened 2022-07-05
Docker compose

tomichec:
This PR integrates the docker-compose based installation. It includes lnbits, postgres and tor-node service (to serve over onion network).

Using this installation is a simple way to get the lnbits up and running publicly within minutes.

PR #1008, opened 2022-09-25
Brianoflondon/docker compose for lnbits

brianoflondon:
This adds a very simple `docker-compose.yml` file to the repo and a little bit more info in the `.env.example` to help get things set up especially connecting to a LND REST api on a service like Voltage.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,7 +13,7 @@ LNBITS_DEFAULT_WALLET_NAME="LNbits wallet"
 LNBITS_AD_SPACE=""
 
 # Hides wallet api, extensions can choose to honor
-LNBITS_HIDE_API=false 
+LNBITS_HIDE_API=false
 
 # Disable extensions for all users, use "all" to disable all extensions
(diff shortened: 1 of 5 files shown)


dni:
imho it does not make a lot of sense to add a docker-compose only to start lnbits itselfs, you could just run docker by itself no composing needed.

i would make a little bit more sense if you had added a postgres container and maybe ssl offloading with nginx.

can you PR https://github.com/lnbits/lnbits/pull/1008/commits/e931bc3f008b603bc22d535f54a1fc1d9bc7de47 only, this is a valid fix.

brianoflondon:
The more examples you give, the more people are likely to get this stuff working, yes I probably will add a postgress to this at some point but I have no idea when. In the meantime, perhaps just having this here will encourage someone else. Adding a DC doesn't take anything away and might just help someone along the way. Do what you want with my contributions.

tomichec:
This is possible duplication with https://github.com/lnbits/lnbits/pull/703

However, that was rejected on the basis that it requires other software and here you require other software to be included.

PR #1025, opened 2022-09-30
Extension for Standard Sats Hedge Service

somes731:
Ready and working with Eclair backend.

I am going to update documentation for https://github.com/standardsats/kollider-hedge and may be prepare docker-compose for LNBits, Postgres and Kollider-hedge.

Added incoming payment listener for Eclair. Sorry, other clients need more work.

### DIFF
--- a/lnbits/core/templates/core/extensions.html
+++ b/lnbits/core/templates/core/extensions.html
@@ -26,7 +26,7 @@
         <q-icon
           :name="extension.icon"
           color="grey-5"
-          style="font-size: 4rem;"
+          style="font-size: 4rem"
         ></q-icon>
         {% raw %}
         <h5 class="q-mt-lg q-mb-xs">{{ extension.name }}</h5>
--- a/lnbits/core/templates/core/wallet.html
+++ b/lnbits/core/templates/core/wallet.html
@@ -168,7 +168,7 @@
               <q-td
                 key="memo"
                 :props="props"
-                style="white-space: normal; word-break: break-all;"
+                style="white-space: normal; word-break: break-all"
               >
                 <q-badge v-if="props.row.tag" color="yellow" text-color="black">
                   <a
@@ -341,7 +341,7 @@
             <q-expansion-item group="extras" icon="edit" label="Rename wallet">
               <q-card>
                 <q-card-section>
-                  <div class="" style="max-width: 320px;">
+                  <div class="" style="max-width: 320px">
                     <q-input
                       filled
                       v-model.trim="newName"
@@ -683,7 +683,7 @@
   </q-dialog>
 
   <q-dialog v-model="paymentsChart.show">
-    <q-card class="q-pa-sm" style="width: 800px; max-width: unset;">
+    <q-card class="q-pa-sm" style="width: 800px; max-width: unset">
       <q-card-section>
         <canvas ref="canvas" width="600" height="400"></canvas>
       </q-card-section>
--- a/lnbits/extensions/boltcards/templates/boltcards/index.html
+++ b/lnbits/extensions/boltcards/templates/boltcards/index.html
@@ -7,7 +7,7 @@
       <q-card-section>
         <div class="row items-center no-wrap q-mb-md">
           <div class="col">
-            <div class="row justify-start" style="width: 150px;">
+            <div class="row justify-start" style="width: 150px">
               <div class="col">
                 <h5 class="text-subtitle1 q-my-none">Cards</h5>
               </div>
@@ -365,7 +365,7 @@
           class="rounded-borders"
         ></qrcode>
       </q-responsive>
-      <p style="word-break: break-all;" class="text-center">
+      <p style="word-break: break-all" class="text-center">
         (Keys for
         <a
           href="https://play.google.com/store/apps/details?id=com.lightningnfcapp"
@@ -373,7 +373,7 @@
           >bolt-nfc-android-app</a
         >)
       </p>
-      <p style="word-break: break-all;">
+      <p style="word-break: break-all">
         <strong>Name:</strong> {{ qrCodeDialog.data.name }}<br />
         <strong>UID:</strong> {{ qrCodeDialog.data.uid }}<br />
         <strong>External ID:</strong> {{ qrCodeDialog.data.external_id }}<br />
--- a/lnbits/extensions/boltz/templates/boltz/_api_docs.html
+++ b/lnbits/extensions/boltz/templates/boltz/_api_docs.html
@@ -12,7 +12,7 @@
       />
       <img
         src="https://boltz.exchange/static/media/Boltz.02fb7acb.svg"
-        style="padding: 5px 9px;"
+        style="padding: 5px 9px"
         alt=""
       />
       <h5 class="text-subtitle1 q-my-none">
--- a/lnbits/extensions/boltz/templates/boltz/index.html
+++ b/lnbits/extensions/boltz/templates/boltz/index.html
@@ -68,7 +68,7 @@
           </template>
           <template v-slot:body="props">
             <q-tr :props="props">
-              <q-td style="width: 10%;">
+              <q-td style="width: 10%">
                 <q-btn
                   unelevated
                   dense
@@ -155,7 +155,7 @@
           </template>
           <template v-slot:body="props">
             <q-tr :props="props">
-              <q-td style="width: 10%;">
+              <q-td style="width: 10%">
                 <q-btn
                   unelevated
                   dense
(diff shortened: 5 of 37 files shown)


arcbtc:
🥳 Excellent!
I notice some core changes, that's something beyond the realm of extension commits usually. I imagined the extension would use the funds in an LNbits wallet to buy/sell to Kolider to keep the peg, but here it looks like that task is done by the Eclair node?

somes731:
@arcbtc no LNBits admin must take care about exchange funds. Core changes were made only for letting extension know about settled outgoing payment. Not just invoice listener. 

It seems I have to make some fixes here and there. Do you use Black for code formatting?

dni:
run ```poetry run black .``` or  ```make black```

somes731:
Thanks @dni.
It might be stupid question. 

Also, do you have developer guide somewhere? Tried quick search.

somes731:
@dni any updates about what to do with this PR to be merged?

arcbtc:
Sorry. We have been busy, and things that change core take a little longer

arcbtc:
You're also right on developer guide. We need to make one for extension builders

callebtc:
Hey, this is a very cool PR, congrats! 

Can you please merge or rebase this into main? We have updated the listeners as you can see from the merge conflicts. I'll go through the PR as soon as its in there!

somes731:
@callebtc 

I've merged parent main into my project and tested with couple of payments. Let me ensure that I have correct formatting for imports.

somes731:
Sorry for annoying you guys, I've applied JS prettier, please trigger workflow.

lnbits/wallets/eclair.py
```diff
@@ -227,3 +227,23 @@ async def paid_invoices_stream(self) -> AsyncGenerator[str, None]:
...
+                        message = json.loads(message)
+
+                        if message and message["type"] == "payment-sent":
+                            yield message["paymentHash"]
+
+            except Exception as exc:
+                logger.error(
+                    f"lost connection to eclair invoices stream: '{exc}', retrying in 5 seconds"
+                )
+                await asyncio.sleep(5)
```
callebtc, line 249: Why do you need a sent payment stream for this extension to work?

lnbits/tasks.py
```diff
@@ -167,15 +184,22 @@ async def perform_balance_checks():
...
+async def payment_callback_dispatcher(checking_id: str):
     """
     Takes incoming payments, sets pending=False, and dispatches them to
     invoice_listeners from core and extensions.
     """
-    payment = await get_standalone_payment(checking_id, incoming=True)
+    payment = await get_standalone_payment(checking_id)
     if payment and payment.is_in:
-        logger.trace(f"sse sending invoice callback for payment {checking_id}")
+        logger.trace(f"sse sending callback for incoming payment {checking_id}")
```
callebtc, line 194: did you test this with internal payments?

lnbits/wallets/eclair.py
somes731: I need this because when user withdraws from the wallet, hedge position has to be adjusted accordingly to the sats balance in the wallet.

lnbits/tasks.py
somes731: I am sorry does it mean some payments between LNBits wallets?

somes731: Bump @callebtc

lnbits/wallets/eclair.py
somes731: Bump @callebtc

lnbits/tasks.py
callebtc: Sorry I was a bit slow to respond. Yes, I meant internal payments between LNbits users / wallets.

callebtc:
> Should I run prettier with some additional options to fix formatting pipeline?

Yes, you can run `make format` for that.

lnbits/wallets/eclair.py
callebtc: Ah ok I get the idea. Question: why would you need a payment stream listener for the node backend? Wouldn't it be enough to listen for outgoing payments from an LNbits wallet and adjust the hedge accordingly?

I think the payment stream might be backend-dependent which may imply that this extension can only be used by a subset of backends. Would be nice if it were usable for all of them!

somes731: >  Wouldn't it be enough to listen for outgoing payments from an LNbits wallet and adjust the hedge accordingly?

I wasn't aware of that. It worked for Eclair backend easily. Could you point on specific LBits events?

Also now it kinda works similarly for incoming and outgoing payments. Maybe it makes sense to keep implementation uniform?

lnbits/tasks.py
somes731: I will look into it.

somes731: Tested it. Indeed it is not really working with internal payments now. But also it seems like there is another problem with Payments objects generally.

https://github.com/lnbits/lnbits/issues/1066

somes731: What i see for internal wallets is that only `is_in` payment events are somehow get into `payment_callback_dispatcher`. It is not entirely clear to me how payments settled in the core because I did even run on critical failure when tried to explore one solution.

I think that outgoing payments are somehow missed inside this function. May be you have some ideas @callebtc otherwise I will try to figure out what exactly happens with regard to internal payments.

callebtc:
I really like this PR but I think the implementation needs adjustment. This should all be implemented on a wallet-level, not on a node-level and should be independent of the backend used. Is there something specific making this impossible?

somes731:
> I really like this PR but I think the implementation needs adjustment. This should all be implemented on a wallet-level, not on a node-level and should be independent of the backend used. Is there something specific making this impossible?

What do you mean wallet-level as opposed to the node level? There is nothing about node-level. Only sent-payment events which I think LNBits Core misses. It is strange not to have them while invoice-paid events present. I agree that not all nodes may have such callbacks or event stream available but this is another point.

callebtc:
> What do you mean wallet-level as opposed to the node level? There is nothing about node-level IMO. Only sent-payment events which I think LNBits Core misses. It is strange not to have them while invoice-paid events present. I agree that not all nodes may have such callbacks or event stream available but this is another point.

I see your point. How about we implement a `Queue` that is triggered when a wallet makes a payment and you hook into that? That way, we wouldn't have to touch any of the wallet backends.

somes731:
> I see your point. How about we implement a Queue that is triggered when a wallet makes a payment and you hook into that? That way, we wouldn't have to touch any of the wallet backends.

I will be fine with whatever solution you will decide to implement in core functionalities. I just need receive and sent events for extension. I expected that my solution might be not the best so I implemented it only for Eclair to see and to test in the first place.

callebtc:
> I will be fine with whatever solution you will decide to implement in core functionalities. I just need receive and sent events for extension. I expected that my solution might be not the best so I implemented it only for Eclair to see and to test in the first place.

Cool, I think we can make this work! Will have to dive in deeper into the extension.

> > I would also kindly ask you to remove all changes that touch other extensions so we can keep this PR smaller.
> 
> But hedge plugin wont' work without events!

I mean all changes that have nothing to do with the PR: 
image

somes731:
> I mean all changes that have nothing to do with the PR:

I mean all changes that have nothing to do with the PR: sorry this is because my formatting tool has different versions. I will figure it out somehow.

somes731:
Ok. I fixed PR @callebtc. Sorry for that.

arcbtc:
Man, Its a real struggle getting this extension merged with the core changes. Is there no way funds can be sent from an lnbits wallet? So I ask for the extension to keep the fiat peg to 75% of an LNbits wallets funds (I imagine afloat is needed), then it could work without the dependence on Eclair, and the core changes, and will get merged. The extension would be much more useful for more people as well.

evd0kim:
Hi @callebtc, did you implement it?

> I see your point. How about we implement a Queue that is triggered when a wallet makes a payment and you hook into that? That way, we wouldn't have to touch any of the wallet backends.

callebtc:
> Hi @callebtc, did you implement it?

No. I've been trying to understand how this extension is supposed to work but the person I'm communicating with is not willing to give me an explanation but seems to just want "the thing done". 

The first attempt of implementing the outgoing payment stream was a failure due to a lack of communication. I can't come up with a better solution if I don't know what you are trying to do. You want to track outgoing payments? There are many ways of doing that. I'm not going to put time in building something that I don't know what it will be used for (because it might be the wrong solution).

So I suggest you explain what exactly the flow of this extension is supposed to be. Let's start with sats going into an LNbits wallet. What happens next?

dni:
this extension should be its own repo @somes731

dni:
closed in favor of https://github.com/lnbits/pegging

Issue #956, opened 2022-09-11
type integer for field "amount" of table "apipayments" for postgres database is not enough


as-schug:
While I try to migrate from sqlite3 to postgres, field amount from table apipayments is of type integer, but, value 3490661000 msats 
is above 2147483648 limit value for integer types

You should consider change that type to BIGINT...

callebtc:
I have encountered this problem myself. My workaround was to use lnbits to create the tables in PG but don't migrate the data yet. I then changed the type of all amount columns to BIGINT manually and did the migration.

I think the problem is that we use a single schema for all db backends and they only share  the INT type (SQLite doesn't know BIGINT). 

Maybe sqlalchemy has a workaround for this but the one possibilty that i can think of would be to change the schema in migrations.py according to the backend used (BIGINT if PG, INT if not).

as-schug:
> I have encountered this problem myself. My workaround was to use lnbits to create the tables in PG but don't migrate the data yet. I then changed the type of all amount columns to BIGINT manually and did the migration.
> 
> I think the problem is that we use a single schema for all db backends and they only share the INT type (SQLite doesn't know BIGINT).
> 
> Maybe sqlalchemy has a workaround for this but the one possibilty that i can think of would be to change the schema in migrations.py according to the backend used (BIGINT if PG, INT if not).

That´s what I did also.
Thanks for you explanation. I don´t know how it works inside lnbits source code...

PR #1030, opened 2022-10-03
fix: use bigint for amounts in postgres

motorina0:
The amount in `sats` can be higher than the the `INT` type in postgres (2^32)
**Solution**:
 - add the `big_int` property (simialr with `timestamp_now`)

**Note**:
 - in Postgres `int8` is an alias for `BIGINT`
 - expected_amount: `BIGINT`
 - timeout_block_height: `INT`

image

### DIFF
--- a/lnbits/db.py
+++ b/lnbits/db.py
@@ -52,6 +52,12 @@ class Compat:
             return ""
         return "<nothing>"
 
+    @property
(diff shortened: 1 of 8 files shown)


AryanJ-NYC:
Closes #956

PR #1033, opened 2022-10-04
On-the-fly invoice expiry check

callebtc:
The purpose of this PR is to greatly speed up startup time for a large lnbits instance by moving the invoice expiry calculation (using `bolt11.decode`) and storage of that information to right after to when the invoice is created instead of parsing through all pending incoming invoices every time during startup and deleting them if the expiry has passed.

**Logic:**
* When invoice is created, calculate and set `expiry`
* Upon startup, delete all payments with `expiry < now()`
* Upon every payment status check (when reloading the wallet for example), check expiry and delete if in the past.

**Migration** (`core` to `6, 7`):
* `6`: Add new column `expiry` to `apipayments`
* `7`: Loop through all `amount > 0 AND pending = true` invoices and calculate expiry, write to `expiry` column

Screenshot 2022-12-02 at 19 05 59


lnbits/core/crud.py
```diff
@@ -339,37 +339,14 @@ async def delete_expired_invoices(
...
+    await (conn or db).execute(
         f"""
-        SELECT bolt11
-        FROM apipayments
-        WHERE pending = true
-          AND bolt11 IS NOT NULL
-          AND amount > 0 AND time < {db.timestamp_now} - {db.interval_seconds(86400)}
+        DELETE FROM apipayments
+        WHERE pending = true AND amount > 0
+          AND expiry < {db.timestamp_now}
```
callebtc, line 347: Delete task operates on database column `expiry` instead of looping through all invoices and calculating their expiry date one at a time.

lnbits/core/crud.py
```diff
@@ -396,12 +373,18 @@ async def create_payment(
     # previous_payment = await get_wallet_payment(wallet_id, payment_hash, conn=conn)
     # assert previous_payment is None, "Payment already exists"
 
+    try:
+        invoice = bolt11.decode(payment_request)
+        expiration_date = datetime.datetime.fromtimestamp(invoice.date + invoice.expiry)
```
callebtc, line 378: Calculate and store invoice expiry date upon creation.

lnbits/core/crud.py
```diff
@@ -396,12 +373,18 @@ async def create_payment(
...
+    except:
+        expiration_date = datetime.datetime.now() + datetime.timedelta(days=31)
```
callebtc, line 380: In case of an error, default to 31 days.

lnbits/db.py
```diff
@@ -29,6 +29,13 @@ def interval_seconds(self, seconds: int) -> str:
             return f"{seconds}"
         return "<nothing>"
 
+    def datetime_to_timestamp(self, date: datetime.datetime):
+        if self.type in {POSTGRES, COCKROACH}:
+            return date.strftime("%Y-%m-%d %H:%M:%S")
+        elif self.type == SQLITE:
+            return time.mktime(date.timetuple())
+        return "<nothing>"
```
callebtc, line 37: Helper function that converts `datetime.datetime` objects to a suitable format for database storage. Uses `Timestamp` for Postgres and Unix timestamps for SQLite.

lnbits/core/models.py
```diff
@@ -170,6 +178,13 @@ async def check_status(
 
         logger.debug(f"Status: {status}")
 
+        if self.is_in and status.pending and self.is_expired:
+            expiration_date = datetime.datetime.fromtimestamp(self.expiry)
+            logger.debug(
+                f"Deleting expired incoming payment {self.checking_id}: expired {expiration_date}"
+            )
+            await self.delete(conn)
```
callebtc, line 186: Check the expiry every time an invoice is checked (usually called by the frontend).

lnbits/core/crud.py
dni: why the amount required here in that query?

lnbits/core/crud.py
dni: 31 days? strikes me as very long time for an invoice expiry. i would go for 1 day.
or maybe even move it to an env variable?

lnbits/core/models.py
dni: just asking, because we had this issue recently, what happens to stuck outgoing invoices?

dni:
very nice!

lnbits/core/models.py
callebtc: Incoming invoices with a passed expiry will not (should not) be accepted by the receiver node so it is (should be) safe to assume that we can delete these payments immediately when the expiry is reached.

lnbits/core/crud.py
callebtc: So we look at only invoices (incoming) and not payments (outgoing).

lnbits/core/crud.py
callebtc: 31 days is the maximum expiry for bolt11 invoices afaik

lnbits/core/migrations.py
```diff
@@ -188,3 +192,64 @@ async def m005_balance_check_balance_notify(db):
...
+
+async def m006_add_invoice_expiry_to_apipayments(db):
+    """
+    Adds invoice expiry field to apipayments and precomputes them for
+    existing entries
+    """
+    try:
+        await db.execute("ALTER TABLE apipayments ADD COLUMN expiry TIMESTAMP")
+    except OperationalError:
+        pass
```
motorina0, line 205: - should `raise` instead of `pass`, migration should not continue if `expiry` column was not added

lnbits/core/migrations.py
```diff
@@ -188,3 +192,64 @@ async def m005_balance_check_balance_notify(db):
...
+                FROM apipayments
+                WHERE pending = true
+                AND bolt11 IS NOT NULL
+                AND expiry IS NULL
+                AND amount > 0 AND time < {db.timestamp_now}
+                """
+            )
+        ).fetchall()
+        # then we delete all expired invoices, checking one by one
+        print(f"Checking expiry of {len(rows)} invoices")
```
motorina0, line 220: -replace `print` with `logger.info` (several occurrences in this file)

lnbits/db.py
```diff
@@ -149,14 +156,7 @@ def _parse_timestamp(value, _):
...
                 psycopg2.extensions.new_type(
-                    (1184, 1114),
-                    "TIMESTAMP2INT",
-                    _parse_timestamp
-                    # lambda value, curs: time.mktime(
-                    #     datetime.datetime.strptime(
-                    #         value, "%Y-%m-%d %H:%M:%S.%f"
-                    #     ).timetuple()
-                    # ),
+                    (1184, 1114), "TIMESTAMP2INT", _parse_timestamp
```
motorina0, line 159: Regarding:
```py
def _parse_timestamp(value, _):
    f = "%Y-%m-%d %H:%M:%S.%f"
    if not "." in value:
        f = "%Y-%m-%d %H:%M:%S"
    return time.mktime(datetime.datetime.strptime(value, f).timetuple())
```
If `expiry` is `None` then it throws:
```py
lnbits/lnbits/db.py", line 136, in _parse_timestamp
    if not "." in value:
TypeError: argument of type 'NoneType' is not iterable
````

File: lnbits/core/migrations.py (hunk 6 above)
Comment (motorina0) [reply lnbits/core/migrations.py:205]: - maybe `ALTER TABLE apipayments` and `UPDATE apipayments` should be two separate migrations (`m006_` and `m007`)

- `ALTER TABLE apipayments` migration can do a check if `expiry` column is present and skip executing the command if true

File: lnbits/core/migrations.py [hunk 9]
```diff
@@ -188,3 +192,64 @@ async def m005_balance_check_balance_notify(db):
...
+                SELECT bolt11, checking_id
+                FROM apipayments
+                WHERE pending = true
+                AND bolt11 IS NOT NULL
+                AND expiry IS NULL
+                AND amount > 0 AND time < {db.timestamp_now}
+                """
+            )
+        ).fetchall()
+        # then we delete all expired invoices, checking one by one
```
motorina0, line 219: - at this point it just adds the `expiry` value (does not execute `DELETE`)

lnbits/core/migrations.py
```diff
@@ -188,3 +192,64 @@ async def m005_balance_check_balance_notify(db):
...
+        ) in enumerate(rows):
+            print(f"Checking invoice {i}/{len(rows)}")
+            try:
+                invoice = bolt11.decode(payment_request)
+            except:
+                continue
+            if payment_hash != invoice.payment_hash:
+                print("Error: {payment_hash} != {invoice.payment_hash}")
+                continue
+
```
motorina0, line 233: - the `continue` statement, means some invoices will have None for `expiry`
- this means the `Payment` model must make `expiry` optional (`expiry: Optional[int]`)

lnbits/db.py
callebtc: Is a typing hint enough?

lnbits/core/migrations.py
```diff
@@ -188,3 +192,64 @@ async def m005_balance_check_balance_notify(db):
...
+        for i, (
+            payment_request,
+            payment_hash,
+        ) in enumerate(rows):
+            print(f"Checking invoice {i}/{len(rows)}")
+            try:
+                invoice = bolt11.decode(payment_request)
+            except:
+                continue
+            if payment_hash != invoice.payment_hash:
```
motorina0, line 230: - since `payment_hash` is the `checking_id` this condition excludes `internal` invoices. Why is this?

lnbits/core/models.py
```diff
@@ -170,6 +178,13 @@ async def check_status(
...
+
         if self.is_out and status.failed:
```
motorina0, line 188: nitpick: use `elif`

motorina0:
Suggestion:
 - add `Expiry Date` to the `Transactions` table
image

motorina0:
tACK (see comments)

callebtc:
> Suggestion:
> 
> * add `Expiry Date` to the `Transactions` table
>   image

Appreciate but I would not put it in, overcrowds the table a bit too much with not relevant information.

lnbits/core/migrations.py
callebtc: Good catch, I should have either used hash or no hash at all (which I will try now)

lnbits/core/migrations.py
callebtc: Separated migrations to `m006` and `m007`.

PR #1034, opened 2022-10-05
updated poetry install notes

TrezorHannes:
- referenced option for python3.10 and newer
- poetry install with --only main, instead of --no-dev (deprecated)

### DIFF
--- a/docs/guide/installation.md
+++ b/docs/guide/installation.md
@@ -26,8 +26,8 @@ sudo apt install python3.9 python3.9-distutils
 
 curl -sSL https://install.python-poetry.org | python3 -
 export PATH="/home/ubuntu/.local/bin:$PATH" # or whatever is suggested in the poetry install notes printed to terminal
-poetry env use python3.9
-poetry install --no-dev
+poetry env use python3.9 # you can exchange with python3.10 or newer versions. Identify your version with python3 --version and specify here
+poetry install --only main
 poetry run python build.py
 
 mkdir data


docs/guide/installation.md
```diff
@@ -26,8 +26,8 @@ sudo apt install python3.9 python3.9-distutils
 
 curl -sSL https://install.python-poetry.org | python3 -
 export PATH="/home/ubuntu/.local/bin:$PATH" # or whatever is suggested in the poetry install notes printed to terminal
-poetry env use python3.9
-poetry install --no-dev
+poetry env use python3.9 # you can exchange with python3.10 or newer versions. Identify your version with python3 --version and specify here
+poetry install --only main
 poetry run python build.py
```
dni, line 31: can you remove this line too? with the new poetry, this should not be needed anymore...

TrezorHannes: > 

Which line is not needed any more specifically?

dni: like that
```
# you can exchange with python3.10 or newer versions. Identify your version with python3 --version and specify here
poetry env use python3.9 
```

TrezorHannes: And on the comment entry, I'll change for both, line 28 `export` and for 29 `env use` to be consistent

dni: by env use can be ommited, if your default version is 3.9 or 3.10, it is just a fix for old ubuntu python versions

PR #1054, opened 2022-10-14
Extension: Pushnotifications

schneimi:
Hey, hope this is the right way to contribute. I don't know much about python and vuejs, anyhow I tried to implement push notifications with copying from other extensions.

Because it worked out quiet good and this code already rests for a while, I tried my best to give it a proper finish and wanted to hand it over to the community.

This was only tested with Umbrel 0.5.1 and modified docker image lnbitsdocker/lnbits-legend:0.9.4 with pywebpush package.

I documented how it works in the README.


callebtc:
Excellent way to contribute indeed, thank you for the PR! This is going to be a useful extension.

arcbtc:
I feel like this should be pulled down to core, and other extensions could make use of it if they wish

arcbtc:
Lets merge this as an extension, then later get it pulled own into core.

arcbtc:
What does this email address do?
image

schneimi:
> What does this email address do? image

Good question, at least it should not be this email address here! I think I just used it, because it's a required field. 

I found some explanation here: https://pypi.org/project/py-vapid/

As far as I understand, this is used by the push servers to contact the subscription providers in case of problems. So I would say this should be the contact email of the lnbits admin. I didn't find any env var of this kind, maybe something like LNBITS_ADMIN_EMAIL would be an idea.

It should be no problem to use a fake email such as admin@example.com, but I think it might be of interest for the admin to get to know about problems with sent push notifications.

Great, you consider merging this. Maybe some feedback gives direction to what is really needed besides this basic functionality.

arcbtc:
Gotcha. Hmm perhaps we do need admin email, what do you think @dni ?

arcbtc:
This works great. After discussing among other lnbits devs, we all agreed this should be in core, with an option to turn on/off in the top right corner via bell Material icon, alongside the color themes option.
Would you be interested in making a new PR adding this to core @schneimi ? You've done all the heavy lifting already. UI would be https://fonts.google.com/icons?icon.query=bell With a `<q-tooltip>Subscribe to notifications</q-tooltip>`

schneimi:
Thx for reviewing and your feedback. Yes, I like the idea and will try to come up with a new PR.

schneimi:
Here is the new PR: #1393

PR #1056, opened 2022-10-17
Add Rounding as tipping option

talvasconcelos:
Adds the option for a costumer to round the value to pay, as a tip.

Ex: amount is 1.25, costumer pays 2. 
0.75 is set as the tip and goes to selected tip wallet!

### DIFF
--- a/lnbits/extensions/tpos/templates/tpos/tpos.html
+++ b/lnbits/extensions/tpos/templates/tpos/tpos.html
@@ -230,7 +230,7 @@
             :outline="!($q.dark.isActive)"
             rounded
             color="primary"
-            label="Round"
+            label="Custom"
           ></q-btn>
           <q-input
             class="q-my-lg"
@@ -239,10 +239,10 @@
             filled
             v-model.number="tipRounding"
             :placeholder="roundToSugestion"
-            hint="Rounded amount"
+            hint="Total amount including tip"
           >
             <template v-slot:append>
-              <q-icon name="send" @click="calculatePercent" />
+              <q-icon name="check" @click="calculatePercent" />
             </template>
           </q-input>
         </div>
@@ -449,7 +449,7 @@
         if (change < 0) {
           this.$q.notify({
             type: 'warning',
-            message: 'Value must be greater than initial amount'
+            message: 'Amount with tip must be greater than initial amount.'
           })
           this.tipRounding = this.roundToSugestion
           return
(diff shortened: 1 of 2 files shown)


lnbits/extensions/tpos/templates/tpos/tpos.html
```diff
@@ -333,9 +358,48 @@ <h5 class="q-mt-none">
       },
       fsat: function () {
         return LNbits.utils.formatSat(this.sat)
+      },
+      isRoundValid() {
+        return this.tipRounding > this.amount
+      },
+      roundToSugestion() {
```
dni, line 365: typo suggestion

talvasconcelos (commented):
looks ok to me!! whatever you bosses think it's best!

motorina0:
- it is not clear that the `Fiat` amount is rounded and not the `sats` (the rounded amount should have a currency suffix)
- problem 1:
  - enter `12.00 USD` and click `OK`
  - on the dialog click `Custom`
  - an input field  with value `15` is displayed
  - click the `checkmark` on the right-side of the input (not very obvious, it should be a button)
  - invoice generated: OK
  - on the screen keyboard enter `9.00 USD` for amount and click `OK`  (do not refresh the window)
  - the `tip` dialog is open, the `custom` input field is visible and has the old value (15)
  - click the `Custom` button again
  - the value is set to `10`, as it should
 - problem 2:
  - on mobile/tablet mode there is no keyboard for the `Custom` input field (only works on desktop?)
image

talvasconcelos:
> * it is not clear that the `Fiat` amount is rounded and not the `sats` (the rounded amount should have a currency suffix)
> * problem 1:
>   
>   * enter `12.00 USD` and click `OK`
>   * on the dialog click `Custom`
>   * an input field  with value `15` is displayed
>   * click the `checkmark` on the right-side of the input (not very obvious, it should be a button)
>   * invoice generated: OK
>   * on the screen keyboard enter `9.00 USD` for amount and click `OK`  (do not refresh the window)
>   * the `tip` dialog is open, the `custom` input field is visible and has the old value (15)
>   * click the `Custom` button again
>   * the value is set to `10`, as it should
> * problem 2:
> * on mobile/tablet mode there is no keyboard for the `Custom` input field (only works on desktop?)
>   image

fixed

lnbits/extensions/tpos/templates/tpos/tpos.html
```diff
@@ -204,19 +204,51 @@ <h5 class="q-mt-none">
...
+            <q-input
+              class="col"
+              ref="inputRounding"
+              v-model.number="tipRounding"
+              :placeholder="roundToSugestion"
+              type="number"
+              hint="Total amount including tip"
+              :prefix="currency"
+            >
+              <!-- <template v-slot:append>
```
motorina0, line 235: dead code

PR #1065, opened 2022-10-21
Add webhook_api_key and webhook_custom_data to lnurlp webhook

dni:
The intention of this PR is to add proper authentication to webhook request of lnurlp via the X-Api-Key header and also make it possible to add a custom_data JSON which will get posted along with the other data.

* added db migration
* added to formfields to frontend
screenshot-1666337347

### HEADERS
```python
Headers({'host': 'localhost:8001', 'accept': '*/*', 'accept-encoding': 'gzip, deflate', 'connection': 'keep-alive', 'user-agent': 'python-httpx/0.23.0', 'content-type': 'application/json; charset=utf-8', 'x-api-key': 'KEYKEY', 'content-length': '423'})
```
### JSON
```json
{"payment_hash": "ede342f2a904cd56d7b644bb009a6c57b352cae71392f8116bc593235bdeda22", "payment_request": "lnbc780n1p34yjp2pp5ah359u4fqnx4d4akgjaspxnv27e49jh8zwf0syttckfjxk77mg3qhp5xme546df47p9f7spgp3mekf6r2sfx2wxz0s3ntgx0hhudkw5ggsqkh9wlmn7pstyaqkel8n7mwj9hhyqyp3k9fezzsywfzadnl86pu78kz2c7q8wmgtg8fj5h6d8983uadap7ewhtv3n3ndcmktu6y3x7mcqrrvylr", "amount": 78000, "comment": "453", "lnurlp": 3, "custom_data": {"lol": "lol2"}}
```


callebtc:
CACK

1) Why do you need to send Header *and* Data? It seems a bit specific to allow for only an X-Api-Key to be set (which is not universally used). Can't you just send authentication data inside the Data?

2) You should do input validation when the form is submitted. This way, the user will enter data that is broken and only later when the webhook is sent see some error when the json is parsed.

dni:
1.  i get your point, we could just put it into custom_data, but the proper place for api keys is the header. it is not that much more of an overhead so i would prefer to do like that.
2. yep, fixing this now

motorina0:
utACK

Some suggestions:
 - as mentioned above, the header names will vary across applications (`x-api-key`, `Authorization`, etc). One simple solution is to have a `JSON` field for headers:
 ```
 {
   header-name-1: "header-value-1"
   header-name-2 "header-value-2"
   ...
}
 ```
- the `headers` and `body` fields should only be displayed if there is an `Webhook URL` value. An `Advanced` toggle would be useful.
- the larger fields (like `body`) should be `textarea` (on multiple fields)

motorina0:
LGTM
Nit pick:
I would still add an `Advanced` toggle, there are quite a few fields now in that pop-up (`headers` and `body` only make sense if there is an URL present)

dni:
fields are only visigble when webhook url is set though

PR #1071, opened 2022-10-23
Add cliche wallet in guide

Bashy:


callebtc:
Thank you, good catch!

PR #1074, opened 2022-10-24
testing workflow to only run once

dni:
@callebtc i think it should be fine like that?

now it should only run on pull_request to main branch and also only on push to main branch? this is what we want right?

```yaml
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```


callebtc:
Non-issue. Closing!

PR #1110, opened 2022-11-07
API DOCS button in footer toolbar

dni:
added a API DOCS button linking to /docs swagger api docs into the footer toolbar.

screenshot-1667824218
screenshot-1667824392


motorina0:
tACK

PR #1116, opened 2022-11-08
Add the ability to check for last (5) transactions

talvasconcelos:
See a list of last TPOS transactions!

image

### DIFF
--- a/lnbits/extensions/tpos/templates/tpos/tpos.html
+++ b/lnbits/extensions/tpos/templates/tpos/tpos.html
@@ -272,6 +272,13 @@
           <q-btn icon="close" flat round dense v-close-popup />
         </q-card-section>
         <q-list separator class="q-mb-lg">
+          <q-item v-if="!lastPaymentsDialog.data.length">
+            <q-item-section>
+              <q-item-label class="text-bold"
+                >No paid invoices</q-item-label
+              >
+            </q-item-section>
+          </q-item>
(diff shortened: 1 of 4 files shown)


lnbits/extensions/tpos/templates/tpos/tpos.html
```diff
@@ -520,6 +559,25 @@ <h5 class="q-mt-none">
...
             response.data.data['BTC' + self.currency][self.currency]
         })
+      },
+      getLastPayments() {
+        return axios
+          .get(`/tpos/api/v1/tposs/${this.tposId}/invoices`)
+          .then(res => {
+            if (res.data && res.data.length) {
+              let last = [...res.data]
+              //last.length = Math.min(last.length, 5)
```
motorina0, line 569: - dead code

lnbits/extensions/tpos/views_api.py
```diff
@@ -81,6 +82,26 @@ async def api_tpos_create_invoice(
     return {"payment_hash": payment_hash, "payment_request": payment_request}
 
 
+@tpos_ext.get("/api/v1/tposs/{tpos_id}/invoices")
+async def api_tpos_get_latest_invoices(tpos_id: str = None):
+    payments = [
```
motorina0, line 87: - should wrap the logic in a `try-except`, we do not want SQL errors to pop-up at the api level

PR #1117, opened 2022-11-08
Fix usermanager API keys

### LINKED ISSUE #344: API Method to create accounts
Hi,

I want create user accounts over the API, I know this is possible with the user management extension. But the impotent part is that I just wana give my end users the keys and then delete everything except the invoice key on my side. But the user management extension keep all that information. Do I need now to write my own API / Extension? I don't want create the account over the regular html and parse the information out of it.

### LINKED ISSUE #871: Usermanager: shows wrong key
Usermanager API sidebar shows invoice API keys for most operations but it should be the admin key.

### LINKED ISSUE #1113: [BUG] problem getting wallets in user manager
Issue getting wallets with User Manager

Running MyNode build, most recent update.
 
When using the API call to get wallets in user manager, it is returning Admin Key Required. I have tried re-making users and it seems to make no difference. I cannot seem to find any way to pull the wallets info. All other features seem to work that I've tried thus far. 

 - LNbits version: 0.9.4
 - Database default

talvasconcelos:
Fixed API keys on example curl in UI!

Added hability for user to keep wallet after user deleted from extension, only through API.

Fixes #344 
Fixes #1113 
Fixes #871

### DIFF
--- a/lnbits/extensions/usermanager/crud.py
+++ b/lnbits/extensions/usermanager/crud.py
@@ -63,10 +63,11 @@ async def get_usermanager_users(user_id: str) -> List[Users]:
     return [Users(**row) for row in rows]
 
 
-async def delete_usermanager_user(user_id: str) -> None:
-    wallets = await get_usermanager_wallets(user_id)
-    for wallet in wallets:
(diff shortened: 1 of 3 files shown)


callebtc:
LGTM!

PR #1123, opened 2022-11-10
Install Extensions

motorina0:
PR moved here: https://github.com/lnbits/lnbits/pull/1231

### Summary
This PR allows extensions to be installed/uninstalled at runtime by an admin user.
 - there is no need for a server restart after an extension has been installed
   - the db is created/updated dynamically (`migrations.py`)
   - the endpoints and static resources are mapped dynamically
 - when an extension is uninstalled all endpoints to that extension will be blocked (including static resources like `html` files)
 - extensions can be installed/uninstalled only by admin users (listed in `LNBITS_ADMIN_USERS` env var)
 - after installation, an extension is not `active`. It must be explicitly activated by the admin user.
 - regular users will only see installed extensions that have been `activated` by an admin user
 - when an extension is uninstalled the `db` for it is NOT removed

### Config
 - a LNbits instance can connect to multiple extension repositories by setting the `LNBITS_EXTENSIONS_MANIFESTS` env variable. Eg:
```
LNBITS_EXTENSIONS_MANIFESTS="https://raw.githubusercontent.com/lnbits/lnbits-extensions/main/extensions.json,https://raw.githubusercontent.com/lnbits/lnbits-extensions/main/extensions-trial.json"
```

 - the repository manifest file (`extensions.json` in this case) is a JSON containing a list of extensions of this form:
 ```
{
      "id": "satspay",
      "name": "SatsPay Server",
      "shortDescription": "Create onchain and LN charges",
      "icon": "payment",
      "archive": "https://github.com/motorina0/lnbits-extensions/raw/main/extensions/satspay.zip",
      "dependencies": [
          "watchonly"
      ],
      "hash": ""
},
```
 - the most important attribute is the "archive" field (an url to the extension archive)
   - for an exact version use the commit link (eg: `https://github.com/lnbits/lnbits-extensions/raw/e28ce6d54e1ab02b05fe42d97a4e788343a7b1d0/extensions/scrub.zip`)
   - it is not necessary for the url to be github related

**Note**:
 - this PR removes several extensions, that is what gives it the high number of `Files changed`
 - only `15 files` are relevant, everything under `lnbits/extensions` can be ignored

### Test
 - use both `poetry` and `uvicorn`
 - only admin users can install/uninstall extensions

**Install&Activate**
 - install an extension (with an admin user)
 - check that regular users cannot yet access the extension
 - `Activate` the extensions
 -  users should be able to `Enable` and use the extension
 - `Deactivate` the extension
 - users should not see the extension in the list or access any UI or endpoints (`HTTP 401`)

**Dependencies**
 - an extension can depend on other extensions. This must be specified in the `dependencies: []` fields of the repository manifest file
 - if extension `A` depends on extension `B` then 
    - `A` cannot be installed until `B` is installed
    - `B` cannot be uninstalled before `A` is uninstalled

**Container Install**
- extension installation adds new files (at `lnbits/extensions`) to the filesystem. If the app runs in a docker `container` then the container filesystem state will be recreated if the container is destroyed and recreated.
- the `zips` for installed extensions are stored on the `LNBITS_DATA_FOLDER` (which must be mapped to a volume)
   ```
    volumes:
       - ./data/extensions:/app/lnbits/data/extensions
     ```
- on restart, the app checks if all the `zips` are installed. If not it will re-install them
- to test:
  - start app in container
  - install extensions
  - destroy container
  - start app again using the same image (with correct volume mapping)
  - expected result: the previously installed extensions are present

**Manual Install**
 - copy the `zip` file to the LNBITS_DATA_FOLDER/extensions` folder
 - restart LNbits
 - the extensions should be automatically installed

### Extension Coding Conventions
There are some rules that extensions must follow in order to be installable

**Imports**
 - no relative import of other extensions
    - **NOK**: `from ..watchonly.crud import get_config, get_fresh_address`
    - **OK**:    `from lnbits.extensions.watchonly.crud import get_config, get_fresh_address`
    - in the future extensions should communicate EXCLUSIVELY via REST API calls (no direct imports)

### Screenshots
- admin users have the "Install New Extension" button
image

- install extension page
image

### CODE REVIEW

File: tests/core/views/test_api.py [hunk 1]
```diff
@@ -162,7 +162,9 @@ async def test_pay_invoice_invoicekey(client, invoice, inkey_headers_from):
 
 # check POST /api/v1/payments: payment with admin key [should pass]
 @pytest.mark.asyncio
-@pytest.mark.skipif(is_regtest, reason="this only works in fakewallet")
+@pytest.mark.skipif(
+    True, reason="this only works in fakewallet"
+)  # is_regtest - must clarify with @calle the purpose of this test. Overlaps with `test_pay_invoice`
```
motorina0, line 167: @callebtc need help to clarify this test

lnbits/app.py
```diff
@@ -130,34 +141,62 @@ def signal_handler(signal, frame):
...
 
+def check_installed_extensions():
+    """
+    Check extensions that have been installed, but for some reason no longer present in the 'lnbits/extensions' directory.
+    One reason might be a docker-container that was re-created.
+    The 'data' directory (where the '.zip' files live) is expected to persist state.
+    """
+    extensions_data_dir = os.path.join(LNBITS_DATA_FOLDER, "extensions")
+
+    zip_files = glob.glob(f"{extensions_data_dir}/*.zip")
```
callebtc, line 152: Needs `os.path.join`?

lnbits/app.py
```diff
@@ -130,34 +141,62 @@ def signal_handler(signal, frame):
...
+    for zip_file in zip_files:
+        ext_name = Path(zip_file).stem
+        if not Path(f"lnbits/extensions/{ext_name}").is_dir():
```
callebtc, line 155: Same here, needs `os.path.join` and/or could use `os.path.exists`

lnbits/app.py
```diff
@@ -130,34 +141,62 @@ def signal_handler(signal, frame):
...
+            with zipfile.ZipFile(zip_file, "r") as zip_ref:
+                zip_ref.extractall("lnbits/extensions/")
```
callebtc, line 157: `os.path.join`

PR #1124, opened 2022-11-11
Docker file fix for missing directory

bitkarrot:
Currently Docker will error and exit because `data` directory is not created. 

This commit adds automatic `data` directory creation to Dockerfile so that the following docker command will not error. 

```
docker run --detach --publish 5000:5000 --name lnbits-legend --volume ${PWD}/.env:/app/.env --volume ${PWD}/data/:/app/data lnbits-legend
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,7 @@ RUN curl -sSL https://install.python-poetry.org | python3 -
 ENV PATH="/root/.local/bin:$PATH"
 
 WORKDIR /app
+RUN mkdir -p /app/lnbits/data
 
 COPY . .
 


bitkarrot:
the above Dockerfile works for deploying to local Docker app and also plebnet playground.

however for fly.io, recommend the mkdir be modified to: 

```
WORKDIR /app
RUN mkdir -p /app/lnbits/data
RUN mkdir /data
```

motorina0:
> 
> ```
> WORKDIR /app
> RUN mkdir -p /app/lnbits/data
> RUN mkdir /data
> ```

Since `WORKDIR` is already `/app` can we use relative path for `mkdir` like so?:
`RUN mkdir -p lnbits/data`

In case `WORKDIR` changes, then `RUN mkdir -p lnbits/data` does not have to be updated

bitkarrot:
> Since `WORKDIR` is already `/app` can we use relative path for `mkdir` like so?: `RUN mkdir -p lnbits/data`

Ok, Updated. Also tested on Docker, runs fine localhost without errors.

motorina0:
Merging: 
 - the `regtest` check failures are unrelated to this PR
 - will fix formatting in a different PR

bitkarrot:
Thank you @motorina0 !

PR #1139, opened 2022-11-21
feat: make boltcard message customizable

gimme:
This adds a field to the card create/edit screen that allows you to customize the message that gets sent with the scan response.

The current default message which includes a long URL for refunds can be jarring to look at. There needs to be a way to opt out from that.

### DIFF
--- a/lnbits/extensions/boltcards/lnurl.py
+++ b/lnbits/extensions/boltcards/lnurl.py
@@ -94,7 +94,9 @@ async def api_scan(p, c, request: Request, external_id: str = None):
         "k1": hit.id,
         "minWithdrawable": 1 * 1000,
         "maxWithdrawable": card.tx_limit * 1000,
-        "defaultDescription": f"Boltcard (refund address lnurl://{lnurlpay})",
+        "defaultDescription": card.message
+            .replace("{refund_url}", f"lnurl://{lnurlpay}")
+            .replace("{card_name}", card.card_name),
     }
 
 
--- a/lnbits/extensions/boltcards/migrations.py
+++ b/lnbits/extensions/boltcards/migrations.py
@@ -58,3 +58,10 @@ async def m001_initial(db):
         );
     """
     )
+
+
+async def m002_add_message(db):
+    """
+    Add column for card message that gets shown on scan.
+    """
+    await db.execute("ALTER TABLE boltcards.cards ADD COLUMN message TEXT NOT NULL DEFAULT 'Boltcard (refund address {refund_url})'")
--- a/lnbits/extensions/boltcards/models.py
+++ b/lnbits/extensions/boltcards/models.py
@@ -17,6 +17,7 @@ class Card(BaseModel):
     id: str
     wallet: str
     card_name: str
+    message: str
     uid: str
     external_id: str
     counter: int
@@ -45,6 +46,7 @@ class Card(BaseModel):
 
 class CreateCardData(BaseModel):
     card_name: str = Query(...)
+    message: str = Query("Boltcard (refund address {refund_url})")
     uid: str = Query(...)
     counter: int = Query(0)
     tx_limit: int = Query(0)
--- a/lnbits/extensions/boltcards/crud.py
+++ b/lnbits/extensions/boltcards/crud.py
@@ -20,6 +20,7 @@ async def create_card(data: CreateCardData, wallet_id: str) -> Card:
             external_id,
             wallet,
             card_name,
+            message,
             counter,
             tx_limit,
             daily_limit,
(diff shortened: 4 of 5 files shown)


lnbits/extensions/boltcards/models.py
```diff
@@ -45,6 +46,7 @@ async def lnurlpay_metadata(self) -> LnurlPayMetadata:
 
 class CreateCardData(BaseModel):
     card_name: str = Query(...)
+    message: str = Query("Boltcard (refund address {refund_url})")
```
callebtc, line 49:
```suggestion
    message: str = Query("")
```

lnbits/extensions/boltcards/lnurl.py
```diff
@@ -88,13 +88,22 @@ async def api_scan(p, c, request: Request, external_id: str = None):
...
+        message = message.replace("{" + key + "}", value)
+
     return {
         "tag": "withdrawRequest",
         "callback": request.url_for("boltcards.lnurl_callback", hitid=hit.id),
         "k1": hit.id,
         "minWithdrawable": 1 * 1000,
         "maxWithdrawable": card.tx_limit * 1000,
-        "defaultDescription": f"Boltcard (refund address lnurl://{lnurlpay})",
+        "defaultDescription": message,
```
callebtc, line 106:
```suggestion
        "defaultDescription": card.message or f"Boltcard (refund address lnurl://{lnurlpay})",
```

lnbits/extensions/boltcards/lnurl.py
```diff
@@ -88,13 +88,22 @@ async def api_scan(p, c, request: Request, external_id: str = None):
...
     lnurlpay = lnurl_encode(request.url_for("boltcards.lnurlp_response", hit_id=hit.id))
+
+    replacements = {
+        "refund_url": f"lnurl://{lnurlpay}",
+        "card_name": card.card_name,
+    }
+    message = card.message
+    for key, value in replacements.items():
+        message = message.replace("{" + key + "}", value)
+
```
callebtc, line 99:
```suggestion
```

callebtc (changes requested):
cACK.

Thank you for this PR! Few comments: I think the default value should not be in `model.py` but in `lnurl.py` or `views_api.py` or something (see my suggestion).

The replacement logic is redundant and probably broken, you should just use the Python functionality (check [this](https://realpython.com/python-f-strings/) if you don't know about it)

lnbits/extensions/boltcards/models.py
gimme: I think it's more logical to have the message property always be representative of the actual message that gets sent. It's clearer and easier to inspect.

With the suggested changes, if someone wanted to actually make the message blank, they'd probably be surprised by the default message being used.

lnbits/extensions/boltcards/lnurl.py
gimme: See reply above.

callebtc: Hey! Default values should not be hard coded into the migrations and in the model definition. Can you find an appropriate way to initialize those values?

lnbits/extensions/boltcards/models.py
callebtc: Strings should not be initialized here like that. Would be better in lnurl.py

lnbits/extensions/boltcards/lnurl.py
callebtc: It should be possible to use f-stings, we can't use this replacement logic.

gimme:
Reverted the replacement logic as it is not essential for this feature.

lnbits/extensions/boltcards/lnurl.py
gimme: Maybe I'm misunderstanding something but I don't see how you can use f-strings on input. Not relevant for this PR anymore though.

lnbits/extensions/boltcards/lnurl.py
gimme: Accepted your suggestion.

lnbits/extensions/boltcards/migrations.py
```diff
@@ -58,3 +58,12 @@ async def m001_initial(db):
...
     """
     )
+
+
+async def m002_add_message(db):
+    """
+    Add column for card message that gets shown on scan.
+    """
+    await db.execute(
+        "ALTER TABLE boltcards.cards ADD COLUMN message TEXT NOT NULL"
```
callebtc, line 68: Message can be NULL or no?

callebtc:
Maybe I'm blind but where is the part where the message is actually shown?

lnbits/extensions/boltcards/lnurl.py
```diff
@@ -94,7 +94,7 @@ async def api_scan(p, c, request: Request, external_id: str = None):
         "k1": hit.id,
         "minWithdrawable": 1 * 1000,
         "maxWithdrawable": card.tx_limit * 1000,
-        "defaultDescription": f"Boltcard (refund address lnurl://{lnurlpay})",
+        "defaultDescription": card.message or f"Boltcard (refund address lnurl://{lnurlpay})",
```
gimme, line 97: > Maybe I'm blind but where is the part where the message is actually shown?

This is where the message is shown.

arcbtc:
We have moved Boltcards into its own repo, can you add here https://github.com/lnbits/boltcards

PR #1141, opened 2022-11-22
fix failing tests because it to low balance in wallet

dni:
test wallet actually had to few funds for paying the invoice in tests/extensions/invoices/test_invoices_api.py


motorina0:
LGTM

PR #1143, opened 2022-11-23
Extension: Cashu

callebtc:
This PR adds the Cashu mint extension to LNbits. The extension has two main logically separate parts:

**The Cashu mint**
The mint is imported as a library from the `cashu` package. All FastAPI endpoints are implemented in LNbits, but the mint operations, including database operations, are handled by the `cashu` lib. The `cashu` lib can accept an external database from LNbits to use for the token and key storage such that all data to operate the mint is kept in the same place as where the rest of LNbits database lives. (Note: Cashu also accepts custom cruds which might be useful in other operations but since the internals of cashu and lnbits are so similar, it is not necessary in this case.)

**The Cashu wallet**
With the help of @motorina0, we have created a whole JavaScript implementation of the Cashu wallet (see [cashu-js-wallet](https://github.com/cashubtc/cashu-js-wallet)). The wallet can be shared with users and runs in their browser. It uses the local storage to store all tokens and keys. It is also a progressive web app (and even a browser extension), which means that it can be easily "added to home screen" on Mobile Chrome and Safari and on Desktop to access it like a native app.


arcbtc:
That db mint error is very confusing

PR #1152, opened 2022-11-28
fix: force default value initialisation (tipjar fix)

motorina0:
Tipjar also has a `CreateCharge` class. The `extra` field has been added in `satspay` but not in `tipjar`.

This PR enforces the default vaues for the `satspay` `CreateCharge` class.


callebtc:
Nice! Could you please resolve the conflicts?

motorina0:
> Nice! Could you please resolve the conflicts?

Done!

PR #1154, opened 2022-11-29
TPOS identifies itself via user-agent

iWarpBTC:
When TPOS is getting a lnurl response, it identifies itself via user-agent header. This identification can be then displayed (not only) in the boltcard extension. 

So let me know if this is a good idea or a bad idea.


callebtc:
Concept isn't clear to me. Could you explain why you needed this?

iWarpBTC:
So I can display it in the User agent column within Boltcard extension. The blank spaces are lnbits/tpos without this. 
user-agent

PR #1169, opened 2022-12-02
Websocket for payments

arcbtc:
Pushes LNbits wallet payments to websocket `wss://<lnbits>/api/v1/ws/<invoice-key>`

You can test here using this websocket here https://codepen.io/benarc/pen/abKRXpq

### DIFF
--- a/lnbits/core/tasks.py
+++ b/lnbits/core/tasks.py
@@ -55,7 +55,9 @@ async def wait_for_paid_invoices(invoice_paid_queue: asyncio.Queue):
                     pass
 
         wallet = await get_wallet(payment.wallet_id)
-        await websocketUpdater(wallet.inkey, {"balance": wallet.balance, "payment": payment.dict()})
+        await websocketUpdater(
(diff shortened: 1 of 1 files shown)


lnbits/core/tasks.py
```diff
@@ -53,6 +54,11 @@ async def wait_for_paid_invoices(invoice_paid_queue: asyncio.Queue):
                 except (httpx.ConnectError, httpx.RequestError):
                     pass
 
+        wallet = await get_wallet(payment.wallet_id)
+        await websocketUpdater(
+            wallet.inkey, {"balance": wallet.balance, "payment": payment.dict()}
+        )
```
callebtc, line 60: mypy requires you to handle the case if `wallet == None` here.

```suggestion
        if wallet:
                await websocketUpdater(
                    wallet.inkey, {"balance": wallet.balance, "payment": payment.dict()}
                )
        else:
                ...
```

callebtc (changes requested):

arcbtc:
This would be better as a dev doc

PR #1218, opened 2022-12-19
LNURLp - save `webhook` call details

motorina0:
### Summary
- **fix**: catch any exception when a `webhook` call fails
  - for example adding a LNURLp callback invalid header (`{"h1": 1}`) was not caught 
  - headers can only have string values, not ints

- feat: store the response `code`, `message` and `body`
 - this can help later troubleshoot why the `webhook` call failed


callebtc:
LGTM

PR #1229, opened 2022-12-22
fix: handle `extra` updates for internal payments

motorina0:
For internal payments there are two entries with the same `hash` in `apipayments`.

This caused the updated of `extra` to loose the previous set values if the wrong payment came first for `fetchone`
image

The fix allows to specify if the `incoming` or `outgoing` payment should be updated.

### DIFF
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -452,7 +452,7 @@ async def update_payment_details(
 
 
 async def update_payment_extra(
-    hash: str,
+    payment_hash: str,
     extra: dict,
     outgoing: bool = False,
     incoming: bool = False,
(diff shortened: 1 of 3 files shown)


lnbits/core/crud.py
```diff
@@ -454,15 +454,24 @@ async def update_payment_details(
...
     Only update the `extra` field for the payment.
     Old values in the `extra` JSON object will be kept unless the new `extra` overwrites them.
     """
+    amount_clause = ""
+
+    if outgoing != incoming:
+        if outgoing:
+            amount_clause = "AND amount < 0"
+        else:
+            amount_clause = "AND amount > 0"
```
callebtc, line 471: I would go for default `outgoing=True` and drop the `incoming` flag altogether.

motorina0: done, defaults to `outgoing=False`

callebtc:
LGTM

PR #1231, opened 2022-12-22
Extension install 02

motorina0:
Follow up of this PR: https://github.com/lnbits/lnbits/pull/1123
Doc: [extension-install.md](https://github.com/lnbits/lnbits/blob/extension_install_02/docs/guide/extension-install.md)
Merge too complex, moved here

### Summary
This PR allows extensions to be installed/uninstalled at runtime by an admin user.
 - there is no need for a server restart after an extension has been installed
   - the db is created/updated dynamically (`migrations.py`)
   - the endpoints and static resources are mapped dynamically
 - when an extension is uninstalled all endpoints to that extension will be blocked (including static resources like `html` files)
 - extensions can be installed/uninstalled only by admin users (listed in `LNBITS_ADMIN_USERS` env var)
 - after installation, an extension is not `active`. It must be explicitly activated by the admin user.
 - regular users will only see installed extensions that have been `activated` by an admin user
 - when an extension is uninstalled the `db` for it is NOT removed
 - [x] install from GitHub releases (see extension-install.md](https://github.com/lnbits/lnbits/blob/extension_install_02/docs/guide/extension-install.md))
 - [x] show `star count` from GitHub stars (for per-repo extensions)
 - [x] allow non-admins to view install page (but not install)
 - [ ] show warning if extension has critical bugs
 - [x] allow older versions to be installed
   - the LNbits instance can have an old version that only supports older extensions
 - [ ] add dependencies in `config.json` (???)

### Limitations
 - **GitHub API**: For unauthenticated requests, the rate limit allows for up to 60 requests per hour.
**Solution**:
  - add the `LNBITS_EXT_GITHUB_TOKEN` env variable. Specifying this variable increases the rate-limit
    - this value is not visible for the super-admin therefore SaaS  can re-use the same token for all instances
    - **Note**: tokens expire after 1 year

### Test
 - use both `poetry` and `uvicorn`
 - only admin users can install/uninstall extensions

**Install&Activate**
 - install an extension (with an admin user)
 - check that regular users cannot yet access the extension
 - `Activate` the extensions
 -  users should be able to `Enable` and use the extension
 - `Deactivate` the extension
 - users should not see the extension in the list or access any UI or endpoints (`HTTP 401`)

**Dependencies**
 - an extension can depend on other extensions. This must be specified in the `dependencies: []` fields of the repository manifest file
 - if extension `A` depends on extension `B` then 
    - `A` cannot be installed until `B` is installed
    - `B` cannot be uninstalled before `A` is uninstalled
 - [ ] specify dependencies for GitHub releases

**Container Install**
- extension installation adds new files (at `lnbits/extensions`) to the filesystem. If the app runs in a docker `container` then the container filesystem state will be recreated if the container is destroyed and recreated.
- the `zips` for installed extensions are stored on the `LNBITS_DATA_FOLDER` (which must be mapped to a volume)
   ```
    volumes:
       - ./data/extensions:/app/lnbits/data/extensions
     ```
- on restart the app checks if:
    - the extension has been installed (in the DB)
    - the zip for that extension is present (if not it re-downloads it)
    - the `zip` is installed. If not it will re-install it
    -
- to test:
  - start app in container
  - install extensions
  - destroy container
  - start app again using the same image (with and without volume mapping)
  - expected result: the previously installed extensions are present

### Extension Coding Conventions
There are some rules that extensions must follow in order to be installable

**Imports**
 - no relative import of other extensions
    - **NOK**: `from ..watchonly.crud import get_config, get_fresh_address`
    - **OK**:    `from lnbits.extensions.watchonly.crud import get_config, get_fresh_address`
    - in the future extensions should communicate EXCLUSIVELY via REST API calls (no direct imports)


arcbtc:
I want this merged so bad. 
What about extensions sharing function names? 
What happens if I try and install two extensions with the same name?

motorina0:
> What about extensions sharing function names?
 - there is no conflict as each extension has its own namespace
 
> What happens if I try and install two extensions with the same name?
 - extension IDs must be unique
 - the ID is also the path in the URL: https://legend.lnbits.com/withdraw/
    - `withdraw` in this case
 - if two extensions have the same name then the last one to be installed will overwrite the previous one

lnbits/app.py
```diff
@@ -105,52 +119,120 @@ def signal_handler(signal, frame):
...
+
+
+def check_installed_extension(ext: InstallableExtension) -> bool:
+    zip_files = glob.glob(
+        os.path.join(settings.lnbits_data_folder, "extensions", "*.zip")
+    )
+
+    if ext.has_installed_version:
+        return True
+    if ext.zip_path in zip_files:
```
dni, line 150: you can make a ```not in``` here put download_archive here and remove the else branch

lnbits/commands.py
```diff
@@ -59,64 +55,37 @@ def bundle_vendored():
...
-
     async with core_db.connect() as conn:
         if conn.type == SQLITE:
             exists = await conn.fetchone(
                 "SELECT * FROM sqlite_master WHERE type='table' AND name='dbversions'"
             )
         elif conn.type in {POSTGRES, COCKROACH}:
             exists = await conn.fetchone(
-                "SELECT * FROM information_schema.tables WHERE table_name = 'dbversions'"
+                "SELECT * FROM information_schema.tables WHERE table_schema= 'public' AND table_name = 'dbversions'"
```
dni, line 65: space missing ```table_schema=```

lnbits/core/crud.py
```diff
@@ -66,6 +67,103 @@ async def get_user(user_id: str, conn: Optional[Connection] = None) -> Optional[
...
+
+async def get_installed_extension(ext_id: str, conn: Optional[Connection] = None):
+    row = await (conn or db).fetchone(
+        "SELECT * FROM installed_extensions WHERE id = ?",
+        (ext_id,),
+    )
+    if not row:
+        return None
+
+    return dict(row)
```
dni, line 144: ```return dict(row) if row else None```

lnbits/core/crud.py
```diff
@@ -66,6 +67,103 @@ async def get_user(user_id: str, conn: Optional[Connection] = None) -> Optional[
...
+        "SELECT * FROM installed_extensions",
+        (),
+    )
+    return [InstallableExtension.from_row(row) for row in rows]
+
+
+async def get_inactive_extensions(*, conn: Optional[Connection] = None) -> List[str]:
+    inactive_extensions = await (conn or db).fetchall(
+        """SELECT id FROM installed_extensions WHERE NOT active""",
+        (),
```
dni, line 160: is this needed if its empty?

lnbits/core/helpers.py
```diff
@@ -0,0 +1,42 @@
...
+
+    async with ext_db.connect() as ext_conn:
+        await run_migration(ext_conn, ext_migrations, current_version)
+
+
+async def run_migration(db, migrations_module, current_version):
+    matcher = re.compile(r"^m(\d\d\d)_")
+    db_name = migrations_module.__name__.split(".")[-2]
+    for key, migrate in migrations_module.__dict__.items():
+        match = match = matcher.match(key)
```
dni, line 30: is this an error? ```match = match = matcher.match(key)```too much matching for me :)

lnbits/core/helpers.py
```diff
@@ -0,0 +1,42 @@
...
+        logger.error(e)
+        raise ImportError(
+            f"Please make sure that the extension `{ext.code}` has a migrations file."
+        )
+
+    async with ext_db.connect() as ext_conn:
+        await run_migration(ext_conn, ext_migrations, current_version)
+
+
+async def run_migration(db, migrations_module, current_version):
```
dni, line 26: is it possible to type the arguments? i can not really follow what is happening here in the next lines

lnbits/core/views/api.py
```diff
@@ -22,21 +22,30 @@
     WebSocketDisconnect,
 )
 from fastapi.exceptions import HTTPException
+from fastapi.params import Body
```
dni, line 25: this should be importet from `from fastapi import Body` doesnt it throw an mypy error?

lnbits/core/views/generic.py
```diff
@@ -52,42 +55,105 @@ async def extensions(
...
+
+        installable_exts: List[
+            InstallableExtension
+        ] = await InstallableExtension.get_installable_extensions()
+        installable_exts += [
+            e for e in installed_exts if e.id not in installed_exts_ids
+        ]
+
+        for e in installable_exts:
+            installed_ext = [ie for ie in installed_exts if e.id == ie.id]
```
dni, line 90: ```installed_ext = next((ie for ie in installed_exts if e.id == ie.i), None)```
get single element

lnbits/core/crud.py
```diff
@@ -66,6 +67,103 @@ async def get_user(user_id: str, conn: Optional[Connection] = None) -> Optional[
...
+        (active, ext_id),
+    )
+
+
+async def delete_installed_extension(
+    *, ext_id: str, conn: Optional[Connection] = None
+) -> None:
+    await (conn or db).execute(
+        """
+        DELETE from installed_extensions  WHERE id = ?
```
dni, line 130: 1 space to much ``ions  WHERE```

lnbits/app.py
```diff
@@ -65,13 +76,16 @@ def create_app() -> FastAPI:
...
     app.add_middleware(GZipMiddleware, minimum_size=1000)
+    app.add_middleware(InstalledExtensionMiddleware)
 
     register_startup(app)
     register_assets(app)
     register_routes(app)
     register_async_tasks(app)
     register_exception_handlers(app)
 
+    setattr(core_app_extra, "register_new_ext_routes", register_new_ext_routes(app))
```
callebtc, line 87: Can you add a comment here?

lnbits/app.py
```diff
@@ -105,52 +119,120 @@ def signal_handler(signal, frame):
...
+        ext_start_func()
+
+    if hasattr(ext_module, f"{ext.code}_static_files"):
+        ext_statics = getattr(ext_module, f"{ext.code}_static_files")
+        for s in ext_statics:
+            app.mount(s["path"], s["app"], s["name"])
+
+    logger.trace(f"adding route for extension {ext_module}")
+
+    prefix = f"/upgrades/{ext.hash}" if ext.hash != "" else ""
```
callebtc, line 210:
```suggestion
    prefix = f"/upgrades/{ext.hash}"
```

should do the same :)

lnbits/commands.py
```diff
@@ -59,64 +55,37 @@ def bundle_vendored():
...
+async def db_versions():
+    async with core_db.connect() as conn:
+        current_versions = await get_dbversions(conn)
+        return current_versions
+
+
+async def load_disabled_extension_list() -> None:
+    """Update list of extensions that have been explicitly disabled"""
+    inactive_extensions = await get_inactive_extensions()
+    settings.lnbits_deactivated_extensions += inactive_extensions
```
callebtc, line 91: Might have unforeseen consequences to overwrite the setting here. Maybe a separate variable (or setting?) would be more appropriate.

lnbits/core/helpers.py
motorina0: yeah, looks like it
this code has just been moved here, the same issue is on `main`

callebtc: lol this works? didn't know

lnbits/extension_manager.py
```diff
@@ -0,0 +1,513 @@
...
+            r
+            for r in all_releases
+            if r.archive == archive and r.source_repo == source_repo
+        ]
+
+        return selected_release[0] if len(selected_release) != 0 else None
+
+
+class InstalledExtensionMiddleware:
+    def __init__(self, app: ASGIApp) -> None:
```
callebtc, line 405: Could you add comments to this middleware and why we need it?

lnbits/extension_manager.py
```diff
@@ -0,0 +1,513 @@
...
+            _, path_name = path_elements
+            path_type = None
+
+        # block path for all users if the extension is disabled
+        if path_name in settings.lnbits_deactivated_extensions:
+            response = JSONResponse(
+                status_code=HTTPStatus.NOT_FOUND,
+                content={"detail": f"Extension '{path_name}' disabled"},
+            )
+            await response(scope, receive, send)
```
callebtc, line 426: What does this do? You await a `JSONResponse` object? I don't get this....

lnbits/extension_manager.py
```diff
@@ -0,0 +1,513 @@
...
+
+    @classmethod
+    async def get_extension_releases(cls, ext_id: str) -> List["ExtensionRelease"]:
+        extension_releases: List[ExtensionRelease] = []
+
+        for url in settings.lnbits_extensions_manifests:
+            try:
+                error_msg = "Cannot fetch extensions manifest"
+                manifest = await gihub_api_get(url, error_msg)
+                if "repos" in manifest:
```
callebtc, line 359: See below

lnbits/extension_manager.py
```diff
@@ -0,0 +1,513 @@
...
+        return extension_list
+
+    @classmethod
+    async def get_extension_releases(cls, ext_id: str) -> List["ExtensionRelease"]:
+        extension_releases: List[ExtensionRelease] = []
+
+        for url in settings.lnbits_extensions_manifests:
+            try:
+                error_msg = "Cannot fetch extensions manifest"
+                manifest = await gihub_api_get(url, error_msg)
```
callebtc, line 358: This could use Pydantic validation! How about making a `Manifest(BaseModel)` and using `Manifest.parse_obj(response_dict)` here?

Then you can do `r.id` instead of `r["id"]` etc. and you'll get an error early enough if the response is malformed.

lnbits/extension_manager.py
```diff
@@ -0,0 +1,513 @@
...
+
+    lates_release_url = (
+        f"https://api.github.com/repos/{org}/{repository}/releases/latest"
+    )
+    error_msg = "Cannot fetch extension releases"
+    latest_release = await gihub_api_get(lates_release_url, error_msg)
+
+    config_url = f"""https://raw.githubusercontent.com/{org}/{repository}/{repo["default_branch"]}/config.json"""
+    error_msg = "Cannot fetch config for extension"
+    config = await gihub_api_get(config_url, error_msg)
```
callebtc, line 494: Could use pydantic validation as well!

lnbits/extension_manager.py
```diff
@@ -0,0 +1,513 @@
...
+                )
+
+        shutil.rmtree(self.ext_dir, True)
+        shutil.copytree(
+            os.path.join(self.ext_upgrade_dir, self.id),
+            os.path.join("lnbits", "extensions", self.id),
+        )
+
+    def nofiy_upgrade(self) -> None:
+        """Update the the list of upgraded extensions. The middleware will perform redirects based on this"""
```
callebtc, line 252:
```suggestion
        """Update the list of upgraded extensions. The middleware will perform redirects based on this"""
```

lnbits/extension_manager.py
```diff
@@ -0,0 +1,513 @@
...
+    version: str
+    archive: str
+    source_repo: str
+    hash: Optional[str]
+    html_url: Optional[str]
+    description: Optional[str]
+    details_html: Optional[str] = None
+
+    @classmethod
+    def from_github_release(cls, source_repo: str, r: dict) -> "ExtensionRelease":
```
callebtc, line 115: `ExtensionGithubRelease` pydantic object?

lnbits/core/helpers.py
motorina0: fixed!
It works was there from the start of time.

lnbits/core/helpers.py
motorina0: Done!
However, I've used `migrations_module: Any`. I could not figure out what the type for all modules is.
`typeof(migrations_module)` simply returns `module`

lnbits/extension_manager.py
```diff
@@ -0,0 +1,513 @@
...
+                latest_release=ExtensionRelease.from_github_release(
+                    repo["html_url"], latest_release
+                ),
+            )
+        except Exception as e:
+            logger.warning(e)
+        return None
+
+    @classmethod
+    def from_manifest(cls, e: dict) -> "InstallableExtension":
```
callebtc, line 307: Pydantic object instead of `e : dict` :)

lnbits/app.py
motorina0: Not really, the new code would produce `"/upgrades/"` for prefix when the hash is empty.

lnbits/commands.py
motorina0: There was indeed an issue that the values of `settings` would be overwritten when an admin user would save the `Settings` from the UI. That was fixed.

In order to make matters more explicit the `Settings` class is now an aggregate of 3 classes:

```py
class Settings(EditableSettings, ReadOnlySettings, TransientSettings):
```

- `EditableSettings` - what the admin user sees in the UI and what gets stored in the DB
- `ReadOnlySettings` - read from the `.env` file, cannot be changed from the code or from the UI
- `TransientSettings`:
```py
class TransientSettings(InstalledExtensionsSettings):
    # Transient Settings:
    #  - are initialized, updated and used at runtime
    #  - are not read from a file or from the `setings` table
    #  - are not persisted in the `settings` table when the settings are updated
    #  - are cleared on server restart
```

lnbits/extension_manager.py
motorina0: yes, done

lnbits/extension_manager.py
callebtc: Thank you for this detailed answer!

callebtc:
LGTM

callebtc (approved):

arcbtc:
Couple of notes:
* Please make star theme the same as what we have in extensions.html
* Remove JS from template file into a separate js file, as we do with other template files in core

motorina0:
> Remove JS from template file into a separate js file, as we do with other template files in core
  - not so straightforward, can be addressed in a different PR
> Have you removed ability to role back extension versions? if so can we have an `install` or `update` button on the `manage` dialogue, without clicking through cards.
  - will keep it like this for now, later can be improved

arcbtc:
Hazaaah! LGTM (after I have made a video and composed a tweet)

PR #1232, opened 2022-12-22
BUG: fix auth issue for superusers  when allowed users are set

dni:
this should fix the issue a @AxelHamburch had in telegram:

"""
Somehow I seem to have locked out my super user account. I had set LNBITS_ALLOWED_USERS and probably forgot to specify the super user. Now I can't get in anymore. Also a LNBITS_ADMIN_UI=false, reboot and again to "true" brings the same super user which does not work.
Can I do a "reset to defaults" from outside the UI?
"""

lnbits/core/views/generic.py not excluding admin users and superuser for allowed_users check

### DIFF
--- a/lnbits/core/views/generic.py
+++ b/lnbits/core/views/generic.py
@@ -38,7 +38,7 @@ async def favicon():
 
 
 @core_html_routes.get("/", response_class=HTMLResponse)
-async def home(request: Request, lightning: str = None):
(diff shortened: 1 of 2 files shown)


AxelHamburch:
Nice, its working now! 🤜🤛
I can no longer lock myself out if I forgot to add the admin to Allowed Users. 😁

PR #1235, opened 2022-12-23
fix: onchain wallet deleted for charge

motorina0:
**Steps**:
 - go to `satspay` extension
 - open `New Charge` dialog
 - select `onchain` wallet (do NOT click `Create Charge` yet)
 - in new tab:
    - go to `watchonly` extension
    - delete the account that the charge was pointing to
 - back in the `satspay`  tab click `Create Charge`

**Actual Result**
 - a technical error is shown: `NoneType' object has no attribute 'address'`
 - there is an error stacktrace

**Expected Result**
 - it fails with a user-friendly message: `Wallet '{onchainwallet}' can no longer be accessed.`
 - no stacktrace

### DIFF
--- a/lnbits/extensions/satspay/views_api.py
+++ b/lnbits/extensions/satspay/views_api.py
@@ -46,6 +46,7 @@ async def api_charge_create(
             **{"paid": charge.paid},
         }
     except Exception as ex:
+        logger.debug(ex)
         raise HTTPException(
             status_code=HTTPStatus.INTERNAL_SERVER_ERROR, detail=str(ex)
         )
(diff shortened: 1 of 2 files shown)


lnbits/extensions/satspay/views_api.py
```diff
@@ -37,13 +37,19 @@
...
+    try:
+        charge = await create_charge(user=wallet.wallet.user, data=data)
+        return {
+            **charge.dict(),
+            **{"time_elapsed": charge.time_elapsed},
+            **{"time_left": charge.time_left},
+            **{"paid": charge.paid},
+        }
+    except Exception as ex:
+        logger.debug(ex)
```
callebtc, line 49:
```suggestion
        logger.debug(f"Satspay error: {str}")
```

PR #1241, opened 2022-12-27
Update BoltCard Extension README

arbadacarbaYK:


lnbits/extensions/boltcards/README.md
```diff
@@ -42,18 +42,23 @@ Updated for v0.1.3
     - Or you can Click the "KEYS / AUTH LINK" button to copy the auth URL to the clipboard. Then paste it into the Android app (Create Bolt Card -> PASTE AUTH URL).
 - Click WRITE CARD NOW and approach the NFC card to set it up. DO NOT REMOVE THE CARD PREMATURELY! 
 
-## Erasing the card - Boltcard NFC Card Creator
-Updated for v0.1.3
+## Rewriting / Erasing the card - Boltcard NFC Card Creator
+
+It is possible not only to reset the keys but also to disable the SUN function and completely erasing the card so it can be used again as a static tag or set up as a new Bolt Card.
```
callebtc, line 47:
```suggestion
It is possible not only to reset the keys but also to disable the SUN function and completely erase the card so it can be used again as a static tag or set up as a new Bolt Card. 
```

lnbits/extensions/boltcards/README.md
```diff
@@ -42,18 +42,23 @@ Updated for v0.1.3
...
+IMPORTANT: 
+* It is immanent that you have access to your old keys so do not erase this card in LNbits before you copied those strings!
+* If you tried to write to them and failed you will need the same amount of positive writing sessions to unlock the card.
+
+- in the BoltCard-Extension click the QR code button next to your old card and copy Key0
+- in the BoltApp click Advanced - Reset keys and paste the Key0 into the first field named Key0
+- repeat with Key1/Key2/Key3/Key0
+- when done pasting all 4 keys scan your card with the BoltApp
+- Thats it 🥳
+- Now if there is all success the card can be safely deleted from LNbits (but keep the keys backuped anyway; batter safe than brick).
```
callebtc, line 58:
```suggestion
- If everything was successful the card can be safely deleted from LNbits (but keep the keys backed up anyway; batter safe than brick).
```

iWarpBTC:
Why was the readme updated from wiping using a json (QRed or copy pasted) to copy paste manualy every single key?

arbadacarbaYK:
Hey Gene, If cards are written wrong fromon the beginning (e.g. only key3+4) it didnt work for Darth. This works always. If you feel it should be in as well we can add the former text ?

iWarpBTC:
Maybe I'm missing something, but in the present readme there are described the same steps that happens when the key json is scanned/pasted. Presuming there's a typo in the last number in `repeat with Key1/Key2/Key3/Key0` where should be 4? So I just don't see how would there be a different outcome.

arbadacarbaYK:
With the json all are updated at once, maybe thats why ? You´re right shoult be 4 in the end, lets then just add both ways and everyone is fine :)

PR #1242, opened 2022-12-27
update lock file and workflows to poetry 1.3.1

dni:
update lock file and workflows to poetry 1.3.1
lock files created with 1.3 are incombatible with our old lock file


callebtc:
How does this interact with our current install guide? Does it install the newest version of poetry (1.3.1) and is that still compatible with our old (1.2.1) lock file?

dni:
our current install guide just mentions how to install latest poetry, which will work fine with this lock file, the only issue is that users with an already install poetry version of 1.2 will have to upgrade `poetry self upgrade` if they want to use the new lockfile, or run `poetry install`

motorina0:
✅ LGTM
nitpick: is there a way to specify the `poetry` (and `python`) version globally (env var?) so we only need to change in one place when an upgrade takes place?

PR #1246, opened 2022-12-28
use bytes.fromhex/hex builtins instead of binascii.hexlify/unhexlify

prusnak:
Since LNBits does not support Python 3.5 and older (which are EOL since 2020 anyway), we can use the builtins instead of binascii.

### DIFF
--- a/lnbits/bolt11.py
+++ b/lnbits/bolt11.py
@@ -1,7 +1,6 @@
 import hashlib
 import re
 import time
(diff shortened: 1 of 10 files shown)


callebtc:
Fantastic PR! 

Could you run `make format` and push again?

callebtc (approved):

prusnak:
> Could you run `make format` and push again?

Done

Issue #1245, opened 2022-12-28
[BUG] QR codes with invoices are missing lightning: prefix


prusnak:
**Describe the bug**

- QR codes that encode LN invoices are missing "lightning:" prefix
- this happens in Wallet, Events, TPoS, Cashu, ..., literally everywhere
- when QR code is missing the prefix, the generic QR code scanner (in iOS camera) does not now what app should it open and by default it tries to open Google Search

**To Reproduce**
1. Generate invoice QR code
2. Try to scan with iOS Camera
3. Observe that the scanned link leads to browser search not to a lightning wallet

**Expected behavior**
I expect the link will lead to an app that can pay the invoice

**Screenshots**

<img src="https://user-images.githubusercontent.com/42201/209802771-735ba7f8-da73-4a8c-9ef2-4baae4bc8909.jpeg" width="400">

**Desktop (please complete the following information):**
 - LNbits version: https://legend.lnbits.com/ => 0df81e5aaf6ea74487356cd0c4b1b2774cefad0c

iWarpBTC:
Besides adding the prefix I thought it could be also nice to make the QR content upper case. It saves space for some reason.

prusnak:
> Besides adding the prefix I thought it could be also nice to make the QR content upper case. It saves space for some reason.

Yes, QR encoding is more efficient when only using upper-case. That said, this opens another can of worms, because lots of wallets do not understand/register the "LIGHTNING:" prefix, they only understand/register the "lightning:" prefix. So you either use "lightning:lnbc1..." or "lightning:LNBC1..." but not "LIGHTNING:LNBC1..."

prusnak:
I think this got fixed in https://github.com/lnbits/lnbits/pull/1253 - thanks @iWarpBTC!

PR #1253, opened 2022-12-28
QR invoices and lnurls got prefix and uppercased

iWarpBTC:
This PR addresses the issue #1245 and adds 'lightning:' prefix into QRs containing payment requests and lnurls. Above that it makes payment requests upper cased to save some space (about from 65x65 px to 57x57 in my testing).

### DIFF
--- a/lnbits/core/templates/core/wallet.html
+++ b/lnbits/core/templates/core/wallet.html
@@ -231,7 +231,7 @@
                           <a :href="'lightning:' + props.row.bolt11">
                             <q-responsive :ratio="1" class="q-mx-xl">
                               <qrcode
-                                :value="props.row.bolt11"
+                                :value="'lightning:' + props.row.bolt11.toUpperCase()"
                                 :options="{width: 340}"
                                 class="rounded-borders"
                               ></qrcode>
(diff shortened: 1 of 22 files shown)


talvasconcelos (approved):

prusnak:
There are quite a lot of places where you added prefix but not the call to `toUpperCase`()`. Is that intentional?

Also it seems the code is getting messier.

Would it be possible to override the behaviour of the `<qrcode>` component so instead of

``` patch
-<qrcode value="{{data}}">
+<qrcode :value="'lightning:' + data.toUpperCase()">
```

you could do something like

``` patch
-<qrcode value="{{data}}">
+<qrcode prefix="lightning" value="{{data}}" uppercase="true">
```

and the component would do the rest?

callebtc (approved):
Awesome PR and great grunting and great trick with the upper case to save QR code space! 

LGTM

callebtc:
> you could do something like
> 
> ```diff
> -<qrcode value="{{data}}">
> +<qrcode prefix="lightning" value="{{data}}" uppercase="value">
> ```
> 
> and the component would do the rest?

That would be great to have but I don't know how viable it is to override the the component itself. I this we can get away with the fix by @iWarpBTC for now!

iWarpBTC:
1. Missing `toUpperCase()` is intentional. The lnurls already are uppercased. 

2. Looks cool, but definitely something I am not capable of do so far. Maybe if somebody shill me with some examples or tutorials, I could take it as a homework for another PR.

PR #1255, opened 2022-12-29
add UUID as id instead of sequential integer

### LINKED ISSUE #1230: [Feature request] LNURLp links should have a random ID, not sequential
**Problem**
LNURLp links look like this right now: https://legend.lnbits.com/lnurlp/**123456** and the number goes up with every new link. 

**Solution**
Use a UUID instead of a counter.

talvasconcelos:
Fixes #1230

### DIFF
--- a/lnbits/extensions/lnurlp/crud.py
+++ b/lnbits/extensions/lnurlp/crud.py
@@ -1,6 +1,5 @@
 from typing import List, Optional, Union
 
-from lnbits.db import SQLITE
 from lnbits.helpers import urlsafe_short_hash
 
 from . import db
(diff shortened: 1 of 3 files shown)


callebtc:
ACK

callebtc:
LGTM

callebtc (approved):

arcbtc:
Wait, what, this is not a bug, did someone complain? https://github.com/lnbits/lnbits/issues/1230#issuecomment-1367605687

fiatjaf:
Looks good to me, but I think https://hashids.org/ would be a nicer solution here, the ids would be shorter, it is a much less disruptive change since there are no migrations, and also it would apply retroactively to all existing ids (although this may be an anti-feature).

Yeah, I don't know, maybe proceed with this change, the UUIDs are not that big.

arcbtc:
Aright, if you're happy 👍
LGTM

PR #1256, opened 2022-12-29
Revert "Update BoltCard Extension README"

iWarpBTC:
Reverts lnbits/lnbits#1241

Change was made based on a wrong assumption. In fact it doesn't matter if the keys are filled via key json or pasted manually.


callebtc:
LGTM thank you for picking up the detail

PR #1257, opened 2022-12-29
don't skip mypy for extensions: hivemind, lnurlpayout

prusnak:
these are already properly typed


prusnak:
The coverage report seems really spurious since I haven't touched `lnbits/db.py` or `lnbits/commands.py`

callebtc:
Rebased in #1272

PR #1259, opened 2022-12-29
remove duplicit functions 

prusnak:
probably merge errors?

### DIFF
--- a/lnbits/extensions/satspay/crud.py
+++ b/lnbits/extensions/satspay/crud.py
@@ -147,53 +147,6 @@ async def save_theme(data: SatsPayThemes, css_id: str = None):
             INSERT INTO satspay.themes (
                 css_id,
                 title,
-                user,
-                custom_css
-                )
-            VALUES (?, ?, ?, ?)
-            """,
-            (
-                css_id,
-                data.title,
-                data.user,
-                data.custom_css,
-            ),
-        )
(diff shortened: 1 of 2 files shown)


lnbits/extensions/satspay/crud.py
```diff
@@ -175,52 +175,5 @@ async def get_themes(user_id: str) -> List[SatsPayThemes]:
...
-    row = await db.fetchone("SELECT * FROM satspay.themes WHERE css_id = ?", (css_id,))
-    return SatsPayThemes.from_row(row) if row else None
-
-
-async def get_themes(user_id: str) -> List[SatsPayThemes]:
-    rows = await db.fetchall(
-        """SELECT * FROM satspay.themes WHERE "user" = ? ORDER BY "title" DESC """,
-        (user_id,),
-    )
-    return [SatsPayThemes.from_row(row) for row in rows]
```
arcbtc, line 222: This is different to the earlier function, are we sure this function is not replacing the previous function and actually the one being used?

prusnak: You are right, it seems the other set of functions should be used - fixed in 2e572106232553a131931aa1ecde2511425715d6

dni:
gerty bit is fixed here https://github.com/lnbits/lnbits/pull/1263/

PR #1260, opened 2022-12-29
let's use unique function names

prusnak:
so mypy is not confused

most of these were typos (copy-pasta) errors anyway


callebtc:
You rock so much it's hard to express.

PR #1262, opened 2022-12-30
fix tipjar mypy

dni:


lnbits/extensions/tipjar/crud.py
```diff
@@ -46,16 +42,11 @@ async def create_tipjar(data: createTipJar) -> TipJar:
...
         """,
         (data.name, data.wallet, data.webhook, data.onchain),
     )
-    if db.type == SQLITE:
-        tipjar_id = result._result_proxy.lastrowid
-    else:
-        tipjar_id = result[0]
-
-    tipjar = await get_tipjar(tipjar_id)
+    row = await db.fetchone("SELECT * FROM tipjar.TipJars LIMIT 1")
```
callebtc, line 48: Why not use `get_tipjar` here?

dni: i think it was an error on my part during refactoring

callebtc:
LGTM doing satoshi's work

PR #1266, opened 2022-12-31
Swap out material icons for extensions for actual logo tiles

arcbtc:
Because we're running out of material icons, and its a nice way for extensions to express themselves.
I'm just going ahead and making what I think the different extension developers would want, and they can update later.


callebtc:
Beautiful 😍

I noticed that the tiles have different heights in mobile. Can it be fixed?

arcbtc:
Thats the same currently with icons, and is outside of the scope of this pr. Its the headings of the extensions which are the issue in the above example.

arcbtc:
Hmmm. as we will have star reviews and the like soon, we should make some room for those elements. 

The below simulates the wordpress plugin layout (without star rating). I am sure they did the leg work on this design decision, so we should just copy.
image

PR #1269, opened 2023-01-02
fix mypy issues for events

dni:


callebtc:
Getting 404 after having paid for an event. 

image

dni:
fixed d371dac8
was unrelated to mypy changes though, it was a bug

PR #1270, opened 2023-01-02
fix scrub for mypy

dni:


lnbits/extensions/scrub/views_api.py
```diff
@@ -43,9 +40,7 @@ async def api_links(
 
 
 @scrub_ext.get("/api/v1/links/{link_id}", status_code=HTTPStatus.OK)
-async def api_link_retrieve(
-    r: Request, link_id, wallet: WalletTypeInfo = Depends(get_key_type)
-):
+async def api_link_retrieve(link_id, wallet: WalletTypeInfo = Depends(get_key_type)):
```
callebtc, line 43: Can you simply delete the argument of this function without breaking it?

dni: yes, `r` was not used and its das in dependency injection

lnbits/extensions/scrub/tasks.py
```diff
@@ -25,9 +25,9 @@ async def wait_for_paid_invoices():
         await on_invoice_paid(payment)
 
 
-async def on_invoice_paid(payment: Payment) -> None:
+async def on_invoice_paid(payment: Payment):
     # (avoid loops)
-    if payment.extra.get("tag") == "scrubed":
+    if payment.extra and payment.extra.get("tag") == "scrubed":
```
prusnak, line 30: I think this should read:

``` python
if payment.extra and payment.extra.get("tag") == "scrubed":
```

so the function bails out even when `payment.extra` is not set, no?

PR #1271, opened 2023-01-02
removing type: ignore from Query, Depends, Body and import them correctly

dni:
removing type: ignore from Query, Depends, Body and import them correctly

### DIFF
--- a/lnbits/core/views/generic.py
+++ b/lnbits/core/views/generic.py
@@ -2,9 +2,8 @@ import asyncio
 from http import HTTPStatus
 from typing import Optional
 
-from fastapi import Request, status
+from fastapi import Depends, Query, Request, status
 from fastapi.exceptions import HTTPException
-from fastapi.params import Depends, Query
 from fastapi.responses import FileResponse, RedirectResponse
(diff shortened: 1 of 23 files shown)


lnbits/core/views/generic.py
```diff
@@ -217,7 +216,7 @@ async def pay():
 
 
 @core_html_routes.get("/deletewallet", response_class=RedirectResponse)
-async def deletewallet(request: Request, wal: str = Query(...), usr: str = Query(...)):  # type: ignore
```
callebtc, line 220: Can you delete the argument without breaking things?

dni: same as in the other pr

lnbits/extensions/copilot/views_api.py
```diff
@@ -1,8 +1,7 @@
 from http import HTTPStatus
 
 from fastapi import Request
-from fastapi.param_functions import Query
-from fastapi.params import Depends
+from fastapi.param_functions import Depends, Query
```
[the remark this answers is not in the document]
dni: eac55c22

PR #1273, opened 2023-01-03
remove unnecessary argument encoding from bytes.decode() and str.encode()

prusnak:
the encoding parameter defaults to `'utf-8'`, so it's not needed

reference:
* https://docs.python.org/3/library/stdtypes.html#str.encode
* https://docs.python.org/3/library/stdtypes.html#bytes.decode


callebtc:
gotsent

PR #1280, opened 2023-01-04
make payment.extra non-optional

prusnak:
this makes handling in tasks.py so much easier

### DIFF
--- a/lnbits/extensions/cashu/tasks.py
+++ b/lnbits/extensions/cashu/tasks.py
@@ -28,6 +28,7 @@ async def wait_for_paid_invoices():
 
 
 async def on_invoice_paid(payment: Payment) -> None:
-    if payment.extra and not payment.extra.get("tag") == "cashu":
+    if payment.extra.get("tag") != "cashu":
         return
+
     return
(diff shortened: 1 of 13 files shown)


lnbits/extensions/cashu/tasks.py
```diff
@@ -28,6 +28,10 @@ async def wait_for_paid_invoices():
 
 
 async def on_invoice_paid(payment: Payment) -> None:
-    if payment.extra and not payment.extra.get("tag") == "cashu":
+    if not payment.extra:
```
prusnak, line 31: This is where the logic was changed, not just a simple refactor. I think the suggested behaviour is more correct than the old one.

dni: we could also try to evalute if there is a good reason for payment.extra to be Optional, maybe we can get rid of it altogether?

prusnak:
Rebased to include changes in recently merged https://github.com/lnbits/lnbits/pull/1286 and https://github.com/lnbits/lnbits/pull/1281

dni:
loving it, LGTM

dni (approved):
lgtm

prusnak:
rebased to cover https://github.com/lnbits/lnbits/pull/1293

prusnak:
if these are merged earlier than this PR, this PR would need to be reworked to address these as well: #1295 #1296 #1297 #1299

PR #1281, opened 2023-01-04
FIX: tpos mypy issues

dni:


lnbits/extensions/tpos/tasks.py
```diff
@@ -20,10 +20,9 @@ async def wait_for_paid_invoices():
 
 
 async def on_invoice_paid(payment: Payment) -> None:
-    if payment.extra.get("tag") != "tpos":
+    if not payment.extra or payment.extra.get("tag") != "tpos":
```
prusnak, line 23: Can you please put

``` python
if not payment.extra:
    return
```

as a separate block before the original condition?

I am trying to increase readability of these in https://github.com/lnbits/lnbits/pull/1280/files and having it as a separate block helps avoiding logical errors.

prusnak (changes requested):

lnbits/extensions/tpos/tasks.py
prusnak: I updated https://github.com/lnbits/lnbits/pull/1280 to remove Optional like you suggested.

After this change, the extra `if not payment.extra:` is not required.

PR #1321, opened 2023-01-06
Make Example Great Again

arcbtc:
resurrecting the Example extension, so we can add guides and examples for extension developers


callebtc:
MEGA

arcbtc:
this is actually mergable as is, but I will add some bits first

PR #1322, opened 2023-01-06
Fix: Boltcard refunds

iWarpBTC:
- fix loading of refund records
- getting rid of unnecessary Unions


dni:
LGTM

callebtc:
Thank you for the fix!

PR #1330, opened 2023-01-08
Example Caddy Configuration for NIP-05 Extension

joelklabo:
I had trouble setting this up for a while because I wasn't using nginx so I thought this might be helpful to others using Caddy. Just a suggestion.


callebtc:
That's great! Thank you, this will be useful for people!

PR #1335, opened 2023-01-09
Limit the size of images in Marketplace

talvasconcelos:
Makes images smaller when uploaded from the UI and checks the image size when adding a product from the API.


callebtc:
LGTM

PR #1336, opened 2023-01-09
fix: mypy - constrain `ignore_missing_imports = "True"`

motorina0:
### Summary
 - `pyproject.toml` had the flag `ignore_missing_imports = "True"`. It was too relaxed so all missing imports were ignored.
 - this PR narrows down what is ignored, enforces stricter rules and cleans-up redundant `# type: ignore`
 
### Changes

- [x] Explicitly specify the modules for which missing imports should be ignored:
  - there is a new config in `pyproject.toml` that specifies for which modules to ignore the imports
  - the global `mypy` `ignore_missing_imports = "True"` has been removed

- [x] use relative imports inside the extension module
  - this is enforced by `mypy` after the "explicit ignore" change
  - this is required for https://github.com/lnbits/lnbits/pull/1231

- [x] remove redundant `# type: ignore`
  - there are now fewer ignores in the code
  - many of the remaining ones can be a source of bugs and should be handled
     - passing an `Optional` to a mandatory input param
     - importing from one extension from another (hard dependency, should use API calls) 
  - the bulk of the changed files are caused by this change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,21 +97,7 @@ exclude = """(?x)(
 [[tool.mypy.overrides]]
 module = [
   "embit.*",
-  "secp256k1.*",
-  "uvicorn.*",
-  "sqlalchemy.*",
-  "sqlalchemy_aio.*",
-  "websocket.*",
-  "websockets.*",
-  "pyqrcode.*",
-  "cashu.*",
-  "shortuuid.*",
-  "grpc.*",
-  "lnurl.*",
-  "bitstring.*",
-  "ecdsa.*",
-  "psycopg2.*",
-  "pyngrok.*"
+  "cashu.*"
 ]
 ignore_missing_imports = "True"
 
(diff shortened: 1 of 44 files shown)


pyproject.toml
```diff
@@ -96,6 +95,10 @@ exclude = """(?x)(
     | ^lnbits/wallets/lnd_grpc_files.
 )"""
 
+[[tool.mypy.overrides]]
+module = "embit.*,secp256k1.*,uvicorn.*,sqlalchemy.*,sqlalchemy_aio.*,websocket.*,websockets.*,pyqrcode.*,cashu.*,shortuuid.*,grpc.*,lnurl.*,bitstring.*,ecdsa.*,psycopg2.*,pyngrok.*"
```
motorina0, line 99: @dni is there a nicer way to build this list (eg: each module on a new line)?

dni: https://mypy.readthedocs.io/en/stable/config_file.html?highlight=pyproject#example-pyproject-toml

docs say its possible

```
[[tool.mypy.overrides]]
module = [
    "somelibrary",
    "some_other_library"
]
```

motorina0: updated, thanks!

dni (approved):
lgtm

PR #1338, opened 2023-01-09
Don't Multiply Sat Price by 100 for NIP-5 Verification

joelklabo:
The price in sats is being multiplied by 100, should only be done for fiat currencies. 

I input a price of 999 sats and got an invoice for 99900:
image


arcbtc:
Yes, `npm install` in the lnbits route, then `make format`

PR #1361, opened 2023-01-12
Adds universal tinyurl redirect for lnurls and the like

arcbtc:
Makes a general tinyurl redirect function that could be used by extensions for making lnurls smaller 
Test here:
https://satoshigo.app/docs#/default/api_create_tinyurl_api_v1_tinyurl_post

Here is a redirected lnurlpay from this redirect https://satoshigo.app/t/76e7f06e
image


lnbits/core/views/api.py
```diff
@@ -706,3 +709,29 @@ async def websocket_update_get(item_id: str, data: str):
...
+
+############################TINYURL##################################
+
+
+@core_app.post("/api/v1/tinyurl")
+async def api_create_tinyurl(url: str, endless: bool = False):
+    tinyurl = await get_tinyurl_by_url(url)
+    if tinyurl:
+        return tinyurl
+    return await create_tinyurl(url, endless)
```
callebtc, line 722: This has no authentication, is that intended?

callebtc (commented):
cACK

Few comments: 

- No authentication: is that wanted? Anyone can create URLs on your instance?
- there is no function to delete tinyurls
- there is a bool called `endless` that isn't used

lnbits/core/models.py
```diff
@@ -213,3 +213,14 @@ class BalanceCheck(BaseModel):
     @classmethod
     def from_row(cls, row: Row):
         return cls(wallet=row["wallet"], service=row["service"], url=row["url"])
+
+
+class TinyURL(BaseModel):
+    id: str
+    url: str
+    endless: bool
+    time: float
```
motorina0, line 222: - I cannot find where `endless` and `time` are used. Can these fields be dropped?

lnbits/core/migrations.py
```diff
@@ -269,3 +269,16 @@ async def m008_create_admin_settings_table(db):
...
+
+
+async def m009_create_tinyurl_table(db):
+    await db.execute(
+        f"""
+        CREATE TABLE IF NOT EXISTS tiny_url (
+          id TEXT PRIMARY KEY,
+          url TEXT,
+          endless BOOL NOT NULL DEFAULT false,
+          time TIMESTAMP NOT NULL DEFAULT {db.timestamp_now}
```
motorina0, line 281: see https://github.com/lnbits/lnbits/pull/1361/files#r1085550024

lnbits/core/crud.py
```diff
@@ -620,3 +620,40 @@ async def create_admin_settings(super_user: str, new_settings: dict):
...
     await db.execute(sql, (super_user, json.dumps(new_settings)))
     return await get_super_settings()
+
+
+# tinyurl
+# -------
+
+
+async def create_tinyurl(domain: str, endless: bool, conn: Optional[Connection] = None):
+    tinyurl_id = uuid4().hex[:8]
```
motorina0, line 630: - `uuid4().hex` returns the `hex` (base16) value. This means only characters in the range `0-1` and `a-f`

- `[:8]` only the first 8 characters (eg: `1ede84bc`)
  - one character represents 2 `bytes`
  - this means the max value range is `2^16` (`65,536`). This is not such a big number, collisions might occur

### Suggestion
This already exists:
```py
def urlsafe_short_hash() -> str:
    return shortuuid.uuid()
```

[shortuuid doc:](https://pypi.org/project/shortuuid/)
> shortuuid solves this problem by generating uuids using Python's built-in uuid module and then translating them to **base57** using lowercase and uppercase letters and digits, and removing similar-looking characters such as l, 1, I, O and 0.

- the **base57** encoding from `shortuuid` would significantly increase the values space

lnbits/core/models.py
arcbtc: https://github.com/lnbits/lnbits/pull/1361#issuecomment-1401954793

lnbits/core/crud.py
arcbtc: Great explanation. Thanks! Done

lnbits/core/views/api.py
arcbtc: yes, I thought it would be more useful that way. Do you think its needed?

arcbtc: Yep. Done for create, delete, and get tinyurls

arcbtc:
> cACK
> 
> Few comments:
> 
>     * No authentication: is that wanted? Anyone can create URLs on your instance?
> 
>     * there is no function to delete tinyurls
> 
>     * there is a bool called `endless` that isn't used

Thanks for review both! Done first two points here. regarding `endless` https://github.com/lnbits/lnbits/pull/1361#issuecomment-1401954793

PR #1363, opened 2023-01-13
[Moved] Wallets: add custom invoice expiry

### LINKED ISSUE #1358: [Feature request] Custom invoice expiration time
Concerned API: `api/v1/payments`

I am currently creating invoices using lnbits, and I would like to have the ability to set a custom expiration time when creating an invoice. I haven't been able to find this feature in the current version of lnbits.

I would like to have an option to set a custom expiration time when creating an invoice through lnbits. This would allow me to set the expiration time to a specific value that works best for my use case.

I have considered using the default expiration time offered by lnbits, but it does not work well for my use case.

**Additional context**
This feature would be especially useful for merchants who want to set a specific expiration time for their invoices. It would also give users more control over their invoicing experience.

callebtc:
This PR turned into an absolute nightmare after I merged main into it. Thank you GitHub. Now it has moved to: #1396

Adds option to change the invoice expiry. Only works for backends that support custom expiry times. Right now these are the following ones.

Replaces #753. Fixes #1358, #352

Tested:

 - [x] LND REST
 - [x] LND GRPC 
 - [x] CLN 
 - [x] Eclair
 - [ ] Sparko
 - [ ] LNbits
 - [ ] LightningTipBot?

### New settings entry: 

`LIGHTNING_INVOICE_EXPIRY` with default value `600` (10 minutes). If no expiry is specified during invoice creation (through the API), then this value will be used. 

### New API option: 

```
{"out": false, "amount": <int>, "expiry": <int>}
```

### Test
Tests invoice creation with 31 days of expiry and checks whether the result is as expected.


lnbits/wallets/eclair.py
```diff
@@ -73,13 +73,17 @@ async def create_invoice(
...
         data: Dict = {"amountMsat": amount * 1000}
+        if kwargs.get("expiry"):
+            data["expireIn"] = kwargs["expiry"]
+
         if description_hash:
-            data["description_hash"] = description_hash.hex()
+            data["descriptionHash"] = description_hash.hex()
         elif unhashed_description:
-            data["description_hash"] = hashlib.sha256(unhashed_description).hexdigest()
+            data["descriptionHash"] = hashlib.sha256(unhashed_description).hexdigest()
```
callebtc, line 86: Fix: wrong key according to https://acinq.github.io/eclair/#payments

callebtc:
I merged main into this PR, now it's a desaster.

PR #1364, opened 2023-01-13
add example configuration for connecting LND over grpc

kornpow:
I mostly just brought the example env file up to date with the documentation located at: https://github.com/lnbits/lnbits/blob/main/docs/guide/wallets.md

PR #1365, opened 2023-01-13
Update UserManager Api Docs

kornpow:
It would make my LNBits dev experience better if all of the apis showed what the response to the api call is, to help me know if what I am doing does what I want it to. Many of the extensions do not document their responses, but for this PR, I will focus only on the UserManager extension.

In addition to updating all/most of the response documentation, I also updates some of the schema information for the requests, for a few API, mostly to help explain the usage of the `admin_id` parameter.

# Before
image
image

# After
image
image

### DIFF
--- a/lnbits/extensions/usermanager/models.py
+++ b/lnbits/extensions/usermanager/models.py
@@ -2,21 +2,21 @@ from sqlite3 import Row
 from typing import Optional
 
 from fastapi.param_functions import Query
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 
 class CreateUserData(BaseModel):
-    user_name: str = Query(...)
-    wallet_name: str = Query(...)
-    admin_id: str = Query(...)
+    user_name: str = Query(..., description="Name of the user")
+    wallet_name: str = Query(..., description="Name of the user")
+    admin_id: str = Query(..., description="Id of the user which will administer this new user")
     email: str = Query("")
     password: str = Query("")
 
 
 class CreateUserWallet(BaseModel):
-    user_id: str = Query(...)
-    wallet_name: str = Query(...)
-    admin_id: str = Query(...)
+    user_id: str = Query(..., description="Target user for this new wallet")
+    wallet_name: str = Query(..., description="Name of the new wallet to create")
+    admin_id: str = Query(..., description="Id of the user which will administer this new wallet")
 
 
 class User(BaseModel):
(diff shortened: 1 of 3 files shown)


callebtc:
Thanks for this PR! 

This is cool but wonder if we should do all API endpoints or none at all for consistency. Might be quite some work though. Any way to automate this?

Nit: There are unrelated changes in `.env.example` in this PR.

kornpow:
> Thanks for this PR!
> 
> This is cool but wonder if we should do all API endpoints or none at all for consistency. Might be quite some work though. Any way to automate this?
> 
> Nit: There are unrelated changes in `.env.example` in this PR. Edit: Ah, just saw #1364

I will rebase now that #1364 was merged.

Would it be acceptable if I just finish implementing response documentation for the usermanager plugin? It will probably be too much of a task to do everything all at once. Also hopefully once people see the pattern for documenting responses, they can begin self-documenting.

arcbtc:
Are the responses not shown in legend.lnbits.com/docs?

callebtc:
Let's have it then! Please run `make format` once and push again. Changes in the .env can remain, since they seem to be missing!

arcbtc:
image
Oh I see! Yeah, that is cool

arcbtc:
Wait, I think this can be done with models, rather than declaring all this as a separate json https://fastapi.tiangolo.com/tutorial/response-model/

kornpow:
I will look into that! I’ve been delaying working on this but I think I’ll be able to push on it this evening.

kornpow:
@arcbtc Made a bit of progress using pydantic models. Still some issues.
image
For example `Usermanager Users Delete` api, I attempted to set the response as a type directly, and not using the `pydantic BaseModel`.
Though this is supposed to say: `[str, int ]` it says `[str, str]`.

For this API in particular, it should probably return something else anyways... I have no idea what this magic response list is referring too :) Pydantic definitely works better with dicts than lists.

Getting this going will definitely be a big endevour but its been fun getting to use LNBits!

dni:
i already did that for the boltz extension with response models, you can look into it

kornpow:
I made more progress on this, and I made better use of Pydantic models that already exist. Getting closer.
Thanks for the idea @dni that helped a bit.

kornpow:
@callebtc @arcbtc I think this is pretty much ready! The formatting check fails, but it passes locally. Looking by the logs it mightve been a one-off error.

arcbtc:
Sorry. Yes please, would you mind pring here https://github.com/lnbits/usermanager-extension

PR #1369, opened 2023-01-14
add flake8 + add pylint + fix reported issues

prusnak:
- add flake8 as dev dependency
- add pylint as dev dependency
- run autoflake once:
  - remove unused imports (from standard library)
  - remove unused variables
  - remove unnecessary pass statements
- fix all enabled pylint issues
- fix some of the flake8 issues (remaining ones will be fixed in future PRs)
- run pylint in the CI (new workflow)

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -1,0 +1,26 @@
+name: flake8
+
+on: [push, pull_request]
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9"]
+        poetry-version: ["1.3.1"]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Set up Poetry ${{ matrix.poetry-version }}
+        uses: abatilo/actions-poetry@v2
+        with:
+          poetry-version: ${{ matrix.poetry-version }}
+      - name: Install dependencies
+        run: |
+          poetry install
+      - name: Run tests
+        run: make flake8
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -21,6 +21,7 @@ jobs:
           poetry-version: ${{ matrix.poetry-version }}
       - name: Install dependencies
         run: |
+          poetry config virtualenvs.create false
           poetry install
       - name: Run tests
         run: make flake8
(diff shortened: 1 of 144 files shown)


callebtc:
Massive! We should add this to `make format` and check in the CI for autoflake as well.

prusnak:
When all issues with flake8 are addressed (still quite a lot of work needed) it will be enough to just add flake8 in the CI. No need for autoflake IMO.

callebtc:
I like that it will strip away unused imports but that's right, it will be picked up by a flake run.

callebtc:
@prusnak we had an update in the boltz extension, could you maybe re-run this once more and then I'll merge this!

prusnak:
> @prusnak we had an update in the boltz extension, could you maybe re-run this once more and then I'll merge this!

rebased and re-run on top of main

prusnak:
the mypy failure is not related to this PR since the CI tests fail while mypy is being installed, not while mypy is being run

prusnak:
I added new commits which fix all enabled pylint issues.

The last commit adds pylint to CI.

prusnak:
Fixed all outstanding Pylint and Flake8 issues. Added Pylint and Flake8 to CI.

prusnak:
Rebased on top of current main

prusnak:
Rebased on top of current main

dni:
how is it going, need any help?

prusnak:
> how is it going, need any help?

It's all done on my part. All we need is a review and a merge. :-)

prusnak:
Rebased on top of current main

Btw. it seems you are removing extensions one by one from the repo and moving them to a separate repository. I don't feel this is the good way, because then we would need to replicate all mypy/pylint/flake8/pyright functionality in all these extension repositories. If this is not done, the extensions will become error-prone.

arcbtc:
Sorry for extracting extensions and making some of your commit redundant! I'll stop for now until we get your pr merged.

LNbits roadmap was always to extract extensions out and make them installable, and by doing so mean core can be the primary focus of testing and improving for this repo. Users can then install extensions they require, which may only be a few, reducing risks for their install significantly. 

LNbits will maintain a manifest of vetted extensions https://github.com/lnbits/lnbits-extensions. Extensions being "vetted", means we have looked at the code and tested to make sure its not doing anything malicious, and have encouraged best practices such as mypy and pylight. If the extension reaches the requirements, it can be added. WordPress have the same process, and I am sure there are some compromises finding that balances where they have the maximum amount of developers, without sacrificing too much quality. Testing clients will also be encouraged for all these separate extension repos, but personally don't think it should be demanded. With the right permissions system, and a clear list of core functions an extension can access we can limit risks, even from a messy extension.

prusnak:
Thanks @arcbtc for explanation, it makes sense. 👍 

We could create a tool (lnbits-extension-checker) which people developing extensions can run on their repos and it will report all found issues. This tool would be also run by people vetting the extension before it's added into the extension -repo.

PR #1392, opened 2023-01-22
fix #1391

### LINKED ISSUE #1391: lost connection to cliche's invoices stream cause 'await'
it appears 
It seems that the cliche backend has a problem in the paid_invoices_stream function. 

The following error is returned by LNBits
lost connection to cliche's invoices stream: 'object WebSocket cannot be used in 'await' expression', retry in 5 seconds

https://github.com/lnbits/lnbits/blob/ffb5a91f6747c0150e0bafe9e322e6f20964fbd0/lnbits/wallets/cliche.py#L154

Removing the 'await' removes the error.

johnongit:
Change the paid_invoices_stream logic.
This  the same logic than eclair.py


callebtc:
Thank you for this PR! You've decided to use another ws function instead of `create_connection`, why is that?

IMO, either we should use only `ws.create_connection` or only `ws.connect`

callebtc:
Please also make sure to run `make format` once before pushing.

johnongit:
> Thank you for this PR! You've decided to use another ws function instead of `create_connection`, why is that?
> 
> IMO, either we should use only `ws.create_connection` or only `ws.connect`

Hello.

Thx for your answer. 
Because I'm not a developer and because the current paid_invoices_stream function does not work properly (#1391).
I used the eclair.py method , but I'm not really skilled to update the whole cliche.py.
https://github.com/lnbits/lnbits/blob/ffb5a91f6747c0150e0bafe9e322e6f20964fbd0/lnbits/wallets/eclair.py#L215-L224

> Please also make sure to run make format once before pushing.

I will !

arcbtc:
Would this work?
```
async with create_connection(
                    self.endpoint
                ) as ws:
                    while True:
```

johnongit:
> create_connection

Sadly no @arcbtc . it returns this

> 2023-01-24 08:01:20.40 | ERROR | lnbits.wallets.cliche:paid_invoices_stream:176 | lost connection to cliche invoices stream: '__aenter__', retrying in 5 seconds

dni:
still relevant? there are merge conflicts now

Issue #1048, opened 2022-10-10
lnaddress error because description_hash


schneimi:
I get this error with v0.9.4 when I actually send sats to the address with WoS.

```
2022-10-10 14:09:45.99 | INFO | xxx.xxx.xxx.xxx:0 - "GET /.well-known/lnurlp/xxxxxxxx HTTP/1.1" 200
2022-10-10 14:10:05.33 | ERROR | handled error
2022-10-10 14:10:05.33 | ERROR | ERROR:
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/usr/local/lib/python3.9/site-packages/starlette/middleware/gzip.py", line 24, in __call__
    await responder(scope, receive, send)
  File "/usr/local/lib/python3.9/site-packages/starlette/middleware/gzip.py", line 43, in __call__
    await self.app(scope, receive, self.send_with_gzip)
  File "/usr/local/lib/python3.9/site-packages/starlette/middleware/cors.py", line 84, in __call__
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.9/site-packages/starlette/exceptions.py", line 93, in __call__
    raise exc
  File "/usr/local/lib/python3.9/site-packages/starlette/exceptions.py", line 82, in __call__
    await self.app(scope, receive, sender)
  File "/usr/local/lib/python3.9/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/usr/local/lib/python3.9/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.9/site-packages/starlette/routing.py", line 670, in __call__
    await route.handle(scope, receive, send)
  File "/usr/local/lib/python3.9/site-packages/starlette/routing.py", line 266, in handle
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.9/site-packages/starlette/routing.py", line 65, in app
    response = await func(request)
  File "/usr/local/lib/python3.9/site-packages/fastapi/routing.py", line 227, in app
    raw_response = await run_endpoint_function(
  File "/usr/local/lib/python3.9/site-packages/fastapi/routing.py", line 160, in run_endpoint_function
    return await dependant.call(**values)
[...]

PR #1393, opened 2023-01-22
[FEAT] Push notification integration into core

schneimi:
Hey, here comes my second PR with push notification integration into core. Went smoother than expected, and I could even make some improvements.

Some notes about the integration:
- The bell is wrapped in a Vue component, because it must also work on other pages in wallet context such as extensions.
- When it is disabled it shows the reason in tooltip (only shows with q-btn type='a') and details in browser console.
- For creating the VAPID key pair `lnbits_startup()` seemed like the right place.
- The VAPID pubkey is passed via component property, which is way better than inline script.
- I used `devnull@example.com` as webpush `sub` because I think there should not be one real email knowing about problems of all LNBits installations.
- Added `url_handlers` to web app manifest. When supported, this opens the installed app instead of the browser when clicking on the notification message.
- Subscription does not work in browser incognito mode. There is intentionally no reliable way to recognize this before the user clicks the bell.
- I removed the test notification message as I couldn't find out why it doesn't work with Chrome. There is a timing issue but service worker is ready. Anyhow I am not sure if users really expect this.

This may be out of scope and you may just skip it, but I also wanted the lnurl-pay `comment` parameter (e.g. https://widgets.twentyuno.net/) to show in the notification message, so I additionally patched `lnaddress/lnurl.py`.
There still was this issue I reported with #1048. Someone made a fix back then, but it never got merged.

Apologies for any formatting errors, I have no proper Python setup. At least I tried my best to take over all review changes from the last PR. ;-)

Someone should test this with iOS devices.


arcbtc:
Not working for me at all, in browser or mobile :/

schneimi:
Oh no.... what exactly doesn't work? Don't you even see the bell, or does the bell not activate? No error messages on server or browser console?

Did you test with incoming payment? Hm, maybe it's some issue with the service worker update, in case you refreshed an already visited app/page. Not sure how the browser handles this. I will look into it with a clean installation.

You could try to manually unregister the service worker to get the new one installed.

arcbtc:
Nothing worked, no bell, nothing. oooo, could be the funding source and some listener issue. I shall try a different one

arcbtc:
Working! No idea why is was not working to begin with

dni:
i would for sure remove data/* files and also package-lock.json

lnbits/core/tasks.py
```diff
@@ -88,3 +98,29 @@ async def mark_webhook_sent(payment: Payment, status: int) -> None:
...
+
+        if comment:
+            body += f"\r\n{comment}"
+
+        for subscription in subscriptions:
+            url = (
+                f"https://{subscription.host}/wallet?usr={wallet.user}&wal={wallet.id}"
+            )
+            await send_push_notification(subscription, title, body, url)
+    return
```
dni, line 126: return is not needed here

lnbits/core/views/api.py
```diff
@@ -677,6 +686,41 @@ async def api_auditor():
...
+
+    for wallet_id in wallet_ids:
+        new_subscription = await create_push_notification_subscription(
+            endpoint, wallet_id, data.subscription, host
+        )
+        assert (
+            new_subscription
+        ), "Newly created push notification subscription couldn't be retrieved"
+
+    return ""
```
dni, line 712: usually we return the model on a POST request, any reason not to do that here? if not also add some return type

lnbits/core/views/api.py
```diff
@@ -677,6 +686,41 @@ async def api_auditor():
...
+
+
+@core_app.delete("/api/v1/push_notification", status_code=HTTPStatus.NO_CONTENT)
+async def api_delete_push_notification_subscriptions(
+    data: DeletePushNotificationSubscription,
+    wallet: WalletTypeInfo = Depends(require_admin_key),
+):
+    await delete_push_notification_subscriptions(data.endpoint)
+    return ""
```
dni, line 721: this return is not needed

arcbtc:
Can we get that privatekey auto-generated and stored in db, so its different for each install?

lnbits/core/views/api.py
```diff
@@ -677,6 +686,41 @@ async def api_auditor():
...
+            endpoint, wallet_id, data.subscription, host
+        )
+        assert (
+            new_subscription
+        ), "Newly created push notification subscription couldn't be retrieved"
+
+    return ""
+
+
+@core_app.delete("/api/v1/push_notification", status_code=HTTPStatus.NO_CONTENT)
```
dni, line 715: the return status_code should be OK here for a successful delete.

arcbtc:
If anyone is interested you can test this pr on https://satoshigo.app, it works well!

lnbits/extensions/lnaddress/lnurl.py
```diff
@@ -1,3 +1,4 @@
+import hashlib
```
dni, line 1: is this file related to the PR? lnbits/extensions/lnaddress/lnurl.py

lnbits/helpers.py
```diff
@@ -211,3 +216,15 @@ def get_current_extension_name() -> str:
     except:
         ext_name = extension_director_name
     return ext_name
+
+
+def get_push_notification_pubkey():
+    push_notification_key_file = os.path.join(
```
dni, line 222: is there a reason why this is not inside the database? because if we use postgres that datadir may not even exist

lnbits/tasks.py
```diff
@@ -181,3 +188,41 @@ async def invoice_callback_dispatcher(checking_id: str):
...
+    Generates VAPID key pair if not already exist
+    """
+    privkey = Vapid().from_file(push_notification_key_file.format("pem"))
+
+    if not os.path.exists(push_notification_key_file.format("pub")):
+        pubkey = privkey.public_key.public_bytes(
+            serialization.Encoding.X962, serialization.PublicFormat.UncompressedPoint
+        )
+        f = open(push_notification_key_file.format("pub"), "w")
+        f.write(b64urlencode(pubkey))
```
dni, line 209: same here, i think database would be a better place for the keys

lnbits/tasks.py
```diff
@@ -181,3 +188,41 @@ async def invoice_callback_dispatcher(checking_id: str):
...
+
+
+async def send_push_notification(subscription, title, body, url=""):
+    try:
+        logger.debug(f"sending push notification")
+        webpush(
+            json.loads(subscription.data),
+            json.dumps({"title": title, "body": body, "url": url}),
+            Vapid().from_file(push_notification_key_file.format("pem")),
+            {"aud": "", "sub": "mailto:devnull@example.com"},
```
dni, line 220: why is this email link needed? is that required for push notification? and why example?

lnbits/tasks.py
```diff
@@ -181,3 +188,41 @@ async def invoice_callback_dispatcher(checking_id: str):
...
+        )
+    except WebPushException as e:
+        """
+        This should clean up the database over time from stale endpoints
+        """
+        logger.error(f"failed sending push notification: {str(e)}")
+        await delete_push_notification_subscriptions(subscription.endpoint)
+        return
```
dni, line 228: not needed

lnbits/extensions/lnaddress/lnurl.py
schneimi: > Can we get that privatekey auto-generated and stored in db, so its different for each install?

Yes, it's important that each installation gets it's own keypair! It already works this way, if you don't check-in the data files. ;-)

Sure, this can be put in db. Just used file system because the lib works this way. Privkey alone should do it.

lnbits/tasks.py
schneimi: Explained it here: https://github.com/lnbits/lnbits/pull/1054#issuecomment-1355900490

Yes, it's mandatory by spec and you get error otherwise.

lnbits/tasks.py
schneimi: Yes, in most use cases this should not pile up to dimensions causing problems. Just thought having a simple cleanup routine is better than none at all, as they are not long living by nature and users will resubscribe sooner or later, producing new ones.

But such thing can be implemented when problems are actually experienced.

schneimi:
I tried to integrate the webpush settings with the `settings` table, but it's coupled too tight to the admin-ui. Not sure if this table is rather meant to be `admin_settings` or one table for all settings. The key pair is now stored in `webpush_settings`.

Some other things that changed:
* Renamed push_notification to webpush (less clumsy)
* Subscriptions are now user based (wallets might be muteable later on)
* Works with multiple tabs, or app installations with different user accounts in one browser (there is just one PushAPI subscription for all)

lnbits/tasks.py
schneimi: Just found out about the WebPushException in case of unsubscribed or expired and could just catch the 410 Gone. There is no reason to hold on to them, they will be tried every time again and will always fail.
```
WebPushException: Push failed: 410 Gone
Response body:push subscription has unsubscribed or expired.
```

schneimi:
Brought the PR up to date with main and tried to fix all workflow errors, but I am out of skills on [this](https://github.com/schneimi/lnbits/actions/runs/4456574124/jobs/7827106142) last one.

Well, I hope I covered all your remarks. Let me know if I can do anything further or if you need some explanations.

dni:
can you contact me on telegram? i want to get this merged :)

arcbtc:
Is this in the new rate-limiter pr, shall I close?

dni:
hi @schneimi i made a huge rebase and fixed some stuff, unfortunatly i think i have broken it somehow. can you jump in again and help me fix it? i want to have to merged asap

schneimi:
> hi @schneimi i made a huge rebase and fixed some stuff, unfortunatly i think i have broken it somehow. can you jump in again and help me fix it? i want to have to merged asap

Hey, sorry for not reaching out earlier, this has been left undone for long. You have good timing, this is my first day off in long time. Happy to have a look into it.

dni:
very cool!  :) i made it all up to date improved some bits :) a lot has changed since the initial commit :)

schneimi:
Indeed a very lot has changed, thx for the rebasing work. :)

Found the problem I think. In helpers.py `t.env.globals["WEBPUSH_PUBKEY"] = settings.lnbits_webpush_pubkey` reads `None`, but I have no clue why. This leads to js error, because it's not a base64 value as expected.

Anyway the bigger problem is, that the VAPID key pair is not persistent anymore with the change to `update_cached_settings(` in services.py. This will change the key pair on each server restart, and lead to push messages getting rejected from already installed client service workers that initialized with another pubkey.

Basically the key pair is set once in a server lifetime and should never change. Not sure what the best solution for this sort of settings is. I also don't see this as editable admin settings, because it can break things. Originally I just put the key files in data folder on the very first start.

Did some cleanup, bugfixing and refactoring on the occasion.

dni:
ok @schneimi settings are now persistent, that took me longer than i want to admit, can you try again?

dni:
reactored inside `update_admin_settings`
https://github.com/lnbits/lnbits/pull/1903

schneimi:
> ok @schneimi settings are now persistent, that took me longer than i want to admit, can you try again?

Great you made it, settings look good! But there is still some kind of problem with the priv key, thats why no notification is sent.

I noticed in the old code I used `privkey.decode("utf-8")` to save, but when populating the value I used `vapid.from_pem(bytes(webpush_settings.lnbits_webpush_privkey, "utf-8"))`.

Hope this helps, unfortunately I cannot test it right now, and I am running out of time. Will be on vacation for two weeks.

Just saw, you inserted `{"aud": "", "sub": "mailto:alan@lnbits.com"}` again. I still would not recommend this, this should be dummy value or admin setting.

dni (approved):

arcbtc (approved):

PR #1402, opened 2023-01-23
TEST: LNbits as fundingsource in regtest

dni:
this pr adds a tool to create a user with predefined ID. so we can use it to test LNbits as fundingsource inside regtest.

-   github workflow
-   tools/create_fake_user.py
-   changes in Dockerfile so we can run the tools/\*.py
-   changes in .dockerignore so tools are available inside docker aswell

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,7 +4,6 @@ docker
 docs
 tests
 venv
-tools
 
 lnbits/static/css/*
 lnbits/static/bundle.js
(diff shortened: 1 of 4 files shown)


dni:
https://github.com/lnbits/lnbits/actions/runs/3989202776/jobs/6841343727#step:7:154

i am a little confused, i did not create the user with the correct adminkey to use in ```LNbitsWallet``` fundingsource, the test still passed which i expected to fail, are we doing no external transactions? only interal ones? because they really should not pass if the fundingsource is broken?

i am going to add the commit to create that user for now, but that is still really weird.

PR #1403, opened 2023-01-23
Changes https for http in super user link

arcbtc:
Rather than have `https://127` the link should be `http://127`


lnbits/core/services.py
```diff
@@ -434,7 +434,7 @@ async def check_admin_settings():
 
         http = "https" if settings.lnbits_force_https else "http"
```
dni, line 435: you could remove this line because its not needed anymore, i did this so it shows a https url when force_https ENV variable is true.

arcbtc: Aye, good point

PR #1404, opened 2023-01-24
update .gitignore to ignore .toml

bitkarrot:
Prevent accidental .toml checkins for fly.io deploys!


dni:
i am not sure about this ignore, maybe only add the fly.toml only, because what about ```pyproject.toml``` you could probably exclude that also? imho this first is better.

.gitignore
```diff
@@ -38,3 +38,6 @@ docker
 
 # Nix
 *result*
+
+# fly.io
+*.toml
```
callebtc, line 43:
```suggestion
fly.toml
```

callebtc: You ok with this @bitkarrot?

bitkarrot: yup @callebtc

.gitignore
```diff
@@ -40,6 +40,9 @@ docker
 # Nix
 *result*
 
+# fly.io
+fly.toml
+
 # Ignore extensions (post installable extension PR)
 extensions/
```
callebtc, line 47:
```suggestion
```

PR #1408, opened 2023-01-25
WIP: Nostr admin

arcbtc:
An always on admin client that connects to relays and can route data to extensions. The extension will also be able to ban/allow list relays


arcbtc:
Moved here https://github.com/lnbits/nostr-admin-extension

PR #1417, opened 2023-01-26
TEST: depend all tests on single regtest

dni:
run all the regtest sequentially instead of in parallel

pro:
* saves github action time
* fails faster
* cleaner workflow file
* less jobs that run

con:
* harder to see which wallet fails
* fails on the first issue


dni:
decided to no do it for now :)

PR #1422, opened 2023-01-26
Adjusted Gerty satoshi quote max length

blackcoffeexbt:
Adjusted Gerty satoshi quote max length now Gertys memory issues are fixed


arcbtc:
Having switched to installable extensions, please remove Gerty to its own repo and add changes there

PR #1426, opened 2023-01-28
Removes Usermanager

arcbtc:


callebtc:
Shouldn't we keep this as a core extension?

arcbtc:
No. There is no such thing. There is core and extensions

PR #1429, opened 2023-01-29
fix gitignore

callebtc:


arcbtc:
This must be ignored for extensions to be extracted. If you want to work on Cashu extract it first

prusnak:
Btw, you can still add gitignored files to commit via `git add -f file` (`-f` == `--force`)

PR #1446, opened 2023-02-03
Added Fee to Tpos Extension

KiddoKodes:
Hey,

  I have added a fee field to the tpos extension so now folks can charge their users for using their tpos service. The feature is optional and defaults to 0% fees. The user gets charged x% more in sats and the extra gets stored in the main wallet.


arcbtc:
Great! Could you please pr https://github.com/lnbits/tpos-extension
As we have pulled tpos out of lnbits now we are moving to extensions being actually installable

KiddoKodes:
So I can close this one?

KiddoKodes:
Created One https://github.com/lnbits/tpos-extension/pull/2

PR #1474, opened 2023-02-07
Serve `nostr` relay info

motorina0:
According to **NIP 11**: https://github.com/nostr-protocol/nips/blob/master/11.md

Some nostr clients do not use the URL `path` for relay info (only the domain name):

- If the relay WS is: 
  - `wss://lnbits.link/nostrrelay/client`
- Then the relay info should be served at 
  - `https://lnbits.link/nostrrelay/client`
- However, some clients only check `https://lnbits.link`

This PR serves generic relay info at `https://lnbits.link`

### DIFF
--- a/lnbits/core/views/generic.py
+++ b/lnbits/core/views/generic.py
@@ -42,7 +42,7 @@ async def favicon():
 @core_html_routes.get("/", response_class=HTMLResponse)
 async def home(request: Request, lightning: str = ""):
 
-    if request.headers.get("accept") == "application/json+nostr":
+    if request.headers.get("accept") == "application/nostr+json":
         return JSONResponse(
             content=settings.lnbits_relay_information,
             headers={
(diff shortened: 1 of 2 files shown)


dni:
LGTM

callebtc:
It still seems out of place and it's very specific to a single extension (nostr relay).

Have you considered using a middleware for patching `/`?

motorina0:
> Have you considered using a middleware for patching /?
- the middleware would still go in `core`
- the condition would be evaluated for each request (not only for `GET "/"`)

callebtc:
> * the middleware would still go in `core`

The middleware could be imported from an extension.

lnbits/core/views/generic.py
```diff
@@ -41,6 +41,17 @@ async def favicon():
 
 @core_html_routes.get("/", response_class=HTMLResponse)
 async def home(request: Request, lightning: str = ""):
+
+    if request.headers.get("accept") == "application/nostr+json":
```
callebtc, line 45: Should check whether `settings.lnbits_relay_information` is set.

arcbtc:
So this is just if they pass `application/nostr+json`? It seems like more an issue on the protocol side. I don't see an issue in adding for now and then we lobby nostr to not have such tight spec.

motorina0:
Changed my mind. Clients are wrong. They MUST use the `path` part.
Closing.

PR #1479, opened 2023-02-10
diagon alley suggestions.

fiatjaf:
Use replaceable events instead of "update" actions. "update" actions are not reliable on Nostr since there is no guarantee you'll ever get all. It's better to use NIP-33 and just publish the same thing over and over.

Also allow each user to have multiple stalls and make them independent, with a different (replaceable) event for each.

One good thing about having stalls be separate events is that one can attach, for example, a `g` (geohash) tag to them that says more-or-less where they are so that can be filtered naturally by relays (see NIP-12 suggestions) if someone wants to query stalls only near them.

### DIFF
--- a/lnbits/extensions/market/README.md
+++ b/lnbits/extensions/market/README.md
@@ -6,24 +6,23 @@
 
 In Diagon Alley, `merchant` and `customer` communicate via NOSTR relays, so loss of money, product information, and reputation become far less likely if attacked.
 
-A `merchant` and `customer` both have a NOSTR key-pair that are used to sign notes and subscribe to events. 
+A `merchant` and `customer` both have a NOSTR key-pair that are used to sign notes and subscribe to events.
 
 #### For further information about NOSTR, see https://github.com/nostr-protocol/nostr
 
-
 ## Terms
 
-* `merchant` - seller of products with NOSTR key-pair
-* `customer` - buyer of products with NOSTR key-pair
-* `product` - item for sale by the `merchant`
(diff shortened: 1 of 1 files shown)


fiatjaf:
I'm sending this because @arcbtc asked me to take a look.

I do have other suggestions not included above:

- since the checkout process depends on NIP-04 entirely, I don't think there should be an option to use email or phone or whatnot. Everything should be done via NIP-04, for simplicity. Of course users can still go out of their way and type: "please send me an email", but that shouldn't be in the spec -- because if it is in the spec that implies that the merchant MUST support phone calls and email messages.
- this should be submitted as a NIP to the NIPs repository.
- I don't think that shipping section helps. I think it is much better for the customer to send their order and the merchant replies with the shipping price along with the price for the total order. Is there a problem with this?
- there are many other small tweaks I would to make this more into the spirit of other Nostr subprotocols, for example: there is no need for a special `"message"` field inside any JSON since the clients can just treat every nonstandard NIP-04 message received as an arbitrary message that must be human-readable; there is no need for UUIDs everywhere; instead of CSV strings you should use normal JSON arrays; and so on.

fiatjaf:
Another idea is to make it so each individual product is its own event. I like this a lot because then "categories" can be set as natural `"t"` tags in their Nostr event and it is easy to build a client that queries products from different stalls in a single Nostr query, we could also put amounts in queryable tags -- and also updates are more granular.

But maybe we would have to experiment more.

arcbtc:
Took on your advise and put here https://github.com/nostr-protocol/nips/pull/330

PR #1510, opened 2023-02-16
Default ext install

motorina0:
###  Summary
- the plan is for all extensions to be pulled out of the main`lnbits` repo.
- however, some LNbits instances expect to have certain extensions installed on start-up
  - these extensions could be installed manually by the admin, but that would mean a certain downtime

**Solution**:
 - this PR introduces a new environment variable: `LNBITS_EXTENSIONS_DEFAULT_INSTALL` (comma separated extension ids)
 - all the extensions in this list will be installed when the server starts
 - the first found release from the `manifest.json` is picked by default
 - if no release found, the extension will not be installed and the server will start without it (and try to find it on the next restart)
 - if an extension from this list is uninstalled then it will be re-installed on the next restart
   - the extension must be removed from this list in order to not be re-installed.


dni:
LGTM

PR #1529, opened 2023-02-20
Added cyber theme

arcbtc:
Because we all liked the theme from this pr https://github.com/lnbits/lnbits/pull/622


arcbtc:
Sadly button font color is too tricky to make dark, so primary and secondary had to be made darker 
image

Marfusios:
@arcbtc thanks for the PR, I think we should really also [modify default CSS params](https://github.com/lnbits/lnbits/pull/622/files#diff-2ddb08a237300a38c4bcc9a07e6a87fe7097c3e44cf595700370ada8418c1abbR36-R55), otherwise there is this ugly UI defect: 

image

Additionally, in my previous PR the biggest impact had different font:

```css
@import url(https://fonts.googleapis.com/css2?family=Montserrat:wght@500;700&display=swap);
...
[data-theme='strike'] {
    font-family: Montserrat, Roboto, -apple-system, 'Helvetica Neue', Helvetica, sans-serif !important;
}
```

PR #1532, opened 2023-02-21
Add redirect paths for extensions

motorina0:
### Summary
- some extensions need to map routes outside their scope (and under the root path)
  - by default extension endpoints are under `/ext_id`

Examples
  - **lnurlp** must map `/.well-known/lnurlp/`
  - **nostrrelay** should serve relay info under `/` (if a particular header is set)

```curl
GET 'http://localhost:5000/' --header 'Accept: application/nostr+json'
```

### Solution
 - an extension can specify a `{ext_id}_redirect_paths` variable in `__init__.py` (similar to `{ext_id}_static_files`)
 - nostrrelay example (here)[https://github.com/lnbits/nostr-relay-extension/blob/main/__init__.py#L23]:

```py
nostrrelay_redirect_paths = [
  {
    "from_path": "/",
    "redirect_to_path": "/api/v1/relay-info",
    "header_filters": {
        "accept": "application/nostr+json"
    }
  }
]
```
 - this means: WHEN a request is made to `/` AND has the header `"accept": "application/nostr+json"` THEN redirect the call to the extension endpoint `nostrrelay/api/v1/relay-info` 
 - for this functionally a new middleware is added (`ExtensionsRedirectMiddleware`)

**Subpaths**
The sub-path will be appended to the redirect (`/abc` in the below example):
```curl
GET 'http://localhost:5000/abc' --header 'Accept: application/nostr+json'
```
will be redirected to
```
GET 'http://localhost:5000/nostrrelay/api/v1/relay-info/abc' --header 'Accept: application/nostr+json'
```

> **Note**:
 - the middleware classes have been extracted to their own file
 - only `ExtensionsRedirectMiddleware` is new (`InstalledExtensionMiddleware` was just moved)

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -29,6 +29,7 @@ from .core.services import check_admin_settings
 from .core.views.generic import core_html_routes
 from .extension_manager import (
     Extension,
+    ExtensionsRedirectMiddleware,
     InstallableExtension,
     InstalledExtensionMiddleware,
     get_valid_extensions,
@@ -76,7 +77,10 @@ def create_app() -> FastAPI:
     )
 
     app.add_middleware(GZipMiddleware, minimum_size=1000)
+
+    # order of these two middlewares is important
     app.add_middleware(InstalledExtensionMiddleware)
+    app.add_middleware(ExtensionsRedirectMiddleware)
 
     register_startup(app)
     register_assets(app)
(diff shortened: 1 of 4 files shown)


callebtc:
LGTM

callebtc:
It wasn't immediatelly clear to me that the redirect will only work with local paths of the extension, i.e.

`"redirect_to_path": "/api/v1/relay-info"` redirects to `nostrrelay/api/v1/relay-info`.

callebtc:
Adding a redirect that listens for a path that has the same name as the extension causes an infinite loop:

```

cashu_redirect_paths = [
    {
        "from_path": "/cashu",
        "redirect_to_path": "/wallet",
    }
]
```

callebtc:
LGTM, tested.

PR #1533, opened 2023-02-21
fix: activate default installed extensions

motorina0:
### Summary
 - extensions can be installed by default when the server starts
 - when an extension is auto-installed it is expected to be ready to use
 - this PR implicitly activates extensions that are auto-installed


callebtc:
LGTM

PR #1542, opened 2023-02-23
RIP Lntxbot, thank you for all your help building LN, you will be missed :(

arcbtc:
Removes Lntxbot from LNbits
Sad, sad day...


callebtc:
F 🫡

PR #1544, opened 2023-02-23
Extend bar for mobile

callebtc:
New bottom main action bar with only three buttons. Better height for mobile use. Added slack under most bottom UI object so the bar can not cover it when scrolling around.

Tester: make sure you can scroll all the way down and read the most bottom content of the page without the bar getting in your way.

telegram-cloud-photo-size-2-5370721535972722642-y


arcbtc:
Buttons are very big. We can also receive by pasting an lnurl. Mind if I have a little tweak?

arcbtc:
Can we go with something a phone wallety, like this?
I can add if you're busy
https://codepen.io/arcbtc/pen/mdzBXaP

arcbtc:
image
Working. Looks good 👍

PR #1545, opened 2023-02-23
Enable extensions to have their own dependencies

jackstar12:
This is an idea i had to tackle the issue of not being ablte to easily have extra dependencies within my extension (in my specific case the [discordbot](https://github.com/jackstar12/discordbot)

This approach uses an additional `pyproject.toml` file which keeps track of the installed extensions (automatically updated)
and is also included in the root `pyproject.toml`.

Whenever a new extension which contains a `pyproject.toml` file is added, it is included within the extensions and then a
`poetry lock`
`poetry install --sync`
from a subprocess.

The same happens when an extension is uninstalled (dependencies will be removed automatically.

You can try it out by cloning my fork and then setting
`LNBITS_EXTENSIONS_MANIFESTS=["https://raw.githubusercontent.com/jackstar12/discordbot/main/manifest.json"]`
in your env file

A feature like that would allow removing dependencies which are installed exclusively for certain extensions to me moved out of the main repository aswell

This is still incomplete (support for other installation methods?) and I'd like to gather feedback and your thoughts on this problem


dni:
i really like the idea, @callebtc @motorina0 what do you guys think ;) can we do something like that?

.gitignore
```diff
@@ -44,5 +44,7 @@ docker
 fly.toml
 
 # Ignore extensions (post installable extension PR)
-extensions/
-upgrades/
+lnbits/extensions/*
+!lnbits/extensions/pyproject.toml
+!lnbits/extensions/extensions/
```
dni, line 49: i do not like the fact that i uses extensions/extensions dir, cant we just but the first pyproject inside lnbits/ and use extensions dir as is?

lnbits/settings.py
```diff
@@ -52,6 +52,7 @@ class ExtensionsSettings(LNbitsSettings):
     lnbits_extensions_default_install: List[str] = Field(default=[])
     # required due to GitHUb rate-limit
     lnbits_ext_github_token: str = Field(default="")
+    lnbits_poetry_path: str = Field(default="poetry")
```
dni, line 55: is this path ever different than poetry? or we can you configure it?

.gitignore
jackstar12: This would work. I had it this way because poetry requires the directory to exist, but I think you're right. It could even be in the root folder aswell

lnbits/settings.py
jackstar12: It would be different if you don't have poetry in your path (can be changed through .env)

prusnak:
I don't really like that this PR reintroduces the extensions directory and I don't think we want to have a central repository of extensions dependencies in this repo.

How about including the required extra dependencies in manifest.json of each extension? Would that work?

jackstar12:
> It doesnt introduce a central extension repository again. It works just like the current one where u can download things. The only addition is a pyproject.toml file in the lnbits/extensions dir which keeps track of the currently installed extensions that depend on some things (those who have a pyproject.toml file).

callebtc:
Here is an alternative idea to this: extensions that need external libs should probably copy them into their own `lib/` directory and import them locally to not mess with the package manager state. 

Thoughts?

jackstar12:
> Here is an alternative idea to this: extensions that need external libs should probably copy them into their own `lib/` directory and import them locally to not mess with the package manager state.
> 
> Thoughts?

I tried this out aswell. However this approach only works as long as the library you want to include doesn't have any dependencies itself. I tried it out with discord.py. Doing  `from lib import discord` was no problem but then it failed because discord depends on aiohttp and it tries to import it through `import aiohttp`. So in order to get it to work one would have to mess with the `PYTHONPATH` which I don't think is a good idea because this could lead to packages being installed twice with different versions

jackstar12:
This is what the flow would look like:
image

When an extension is installed the `lnbits/extensions/pyproject.toml` is dynamically updated and when changes are detected, they are installed. When something goes wrong, it's just being rolled back

The same happens when the extension is uninstalled again (instead of adding removing)

pyproject.toml
```diff
@@ -39,6 +39,9 @@ async-timeout = "4.0.2"
 pyln-client = "0.11.1"
 boltz-client = "0.1.3"
 cashu = "0.9.0"
+toml = "^0.10.2"
+extensions = {path = 'lnbits/extensions', develop = true}
+types-toml = "^0.10.8.5"
```
dni, line 44: types-toml should be dev-dependencies

jackstar12:
I have added some more changes.

I've restructured some startup logic to handle more edge cases. 
- First, I moved `register_routes` into the startup event, so that it can run after the dependencies of extensions have been checked (including extensions that try to import not yet installed stuff would lead to errors)
- I did the same thing with `check_installed_extensions` in order to avoid a certain edge case.
If you were to delete an installed extension from the file system which has dependencies, and then run `check_extension_dependencies` it would remove these thinking it was uninstalled, but as soon as `check_installed_extensions` ran it installs it right back. 
- Because `check_installed_extensions` now runs before `register_routes` it does not have to register the routes anymore

Makefile
```diff
@@ -46,7 +46,21 @@ test:
...
 	DEBUG=true \
-	poetry run pytest
+	poetry run pytest -m "not extensions"
+
+# Due to being rather slow, these have their own command
+test-ext:
+	BOLTZ_NETWORK="regtest" \
+	BOLTZ_URL="http://127.0.0.1:9001" \
+	BOLTZ_MEMPOOL_SPACE_URL="http://127.0.0.1:8080" \
+	BOLTZ_MEMPOOL_SPACE_URL_WS="ws://127.0.0.1:8080" \
```
prusnak, line 56: These 4 lines have been removed in https://github.com/lnbits/lnbits/commit/f943ca7d5127c92bd7c796d76bcfb5f1e66abfbd from the `test` target. No need to add them here.

prusnak (changes requested):

jackstar12:
@dni @callebtc @prusnak 
If you want to try out the extensions test, you can do so by registering these manifests:
```
main: https://raw.githubusercontent.com/lnbits/lnbits-extensions/main/extensions.json
intentionally invalid extension: https://raw.githubusercontent.com/jackstar12/invalid/main/manifest.json
extension with dependencies: https://raw.githubusercontent.com/jackstar12/lnbits-dependencies/main/manifest.json
```

lnbits/core/helpers.py
```diff
@@ -63,3 +64,19 @@ async def stop_extension_background_work(ext_id: str, user: str):
...
+    """
+    Call a subprocess, waiting for it to finish. If it exits with a non-zero code, an exception is thrown.
+    """
+    process = await asyncio.create_subprocess_exec(*args, **kwargs)
+    # stdout, stderror = await process.communicate()
+    # if stdout:
+    #    logger.info(stdout)
+    #
+    # logger.error(stderror)
+    # code = process.returncode
```
callebtc, line 79: Some comments here.

callebtc:
Concept ACK

jackstar12:
> What happens to the poor lads who still use `requirements.txt` to install LNbits? Is there a way to detect that and let them know that installing an extension with deps will fail (at least)?

The way I currently do that is by checking id poetry is installed (just executing `poetry` or the specified poetry path in a subprocess). If  one tries to install am extension with deps it tells the user that this extension requires the usage of poetry (or nix?). 

The check could be a bit improved, because poetry being doesnt guarantee that its actually used to run lnbits, so an additional check with the path of the executing interpreter should be done.

dni:
after talking with @jackstar12 , this is closed

PR #1549, opened 2023-02-25
introduce baseurl and remove force_https

dni:
IDEA: introduce BASEURL variable like other webapplications also have

this PR introduces the settings variable `LNBITS_BASEURL`

this makes generating URLs on the serverside much easier for developers.
you do not need to rely on a request object anymore to create an absolute URL,
for example for all the lnurl stuff, server startup log/superuser urls, etc...

the obvious downside is that users need to introduce a new variable if they are behind a proxy.

but i think it is tradeoff worth taking, because this will reduce the overall complexity of the code
and also simplyfies generating proper url for other developers aswell.

also gets rid of `LNBITS_FORCE_HTTPS` variable, because it is not used anymore.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,9 +8,6 @@ PORT=5000
 
 DEBUG=false
 
-# Baseurl for creating urls via the backend
-LNBITS_BASEURL="http://0.0.0.0/"
-
 # Allow users and admins by user IDs (comma separated list)
 LNBITS_ALLOWED_USERS=""
 LNBITS_ADMIN_USERS=""
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,9 @@ PORT=5000
 
 DEBUG=false
 
+# Baseurl for creating urls via the backend
+LNBITS_BASEURL="http://0.0.0.0/"
+
 # Allow users and admins by user IDs (comma separated list)
 LNBITS_ALLOWED_USERS=""
 LNBITS_ADMIN_USERS=""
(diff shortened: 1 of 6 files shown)


.env.example
```diff
@@ -8,6 +8,9 @@ PORT=5000
 
 DEBUG=false
 
+# Baseurl for creating urls via the backend
+LNBITS_BASEURL="http://0.0.0.0/"
```
callebtc, line 12:
```suggestion
LNBITS_BASEURL="http://127.0.0.1/"
```

callebtc: Some OS'es can't connect to `0.0.0.0`, it's only safe to use it for _listening_ on an IP (not connecting).

lnbits/settings.py
```diff
@@ -88,7 +88,7 @@ class ThemesSettings(LNbitsSettings):
 
 
 class OpsSettings(LNbitsSettings):
-    lnbits_force_https: bool = Field(default=False)
+    lnbits_baseurl: str = Field(default="http://0.0.0.0/")
```
callebtc, line 91:
```suggestion
    lnbits_baseurl: str = Field(default="http://127.0.0.1/")
```

.env.example
dni: removed it from .env.template for now, as we talked

lnbits/settings.py
dni: 18ec8549

Issue #1442, opened 2023-02-02
[BUG] Invalid OpenAPI definition


titusz:
The generated OpenAPI definition at https://legend.lnbits.com/openapi.json is invalid.

To reproduce load the OpenAPI definition into an editor like https://editor.swagger.io/ or use a command line tool like https://pypi.org/project/openapi-spec-validator/

For example the `CreatePayLinkData` schema has the following property:

```json
"comment_chars": {
  "title": "Comment Chars",
  "exclusiveMaximum": 800.0,
  "minimum": 0.0,
  "type": "integer",
  "default": 0
}
```
`exclusiveMaximmum` must be a boolean and not a number.

The OpenAPI spec needs to be valid to support generating API client libraries from the spec. Due to the spec errors I failed to generate a python client library with https://github.com/openapi-generators/openapi-python-client

Best case would be to validate the OpenAPI spec during GitHub CI testing.

dni:
damn ther are more issues than this, i have no idea what is going on there? can you take i look? https://github.com/lnbits/lnbits/actions/runs/4280189161/jobs/7451668700

dni:
i think i finally figured it out 
https://github.com/lnbits/lnbits/pull/1551

PR #1551, opened 2023-02-27
add github workflow for checking openapi

dni:
adds workflow that validates openapi.json

relating to issue: https://github.com/lnbits/lnbits/issues/1442

### DIFF
--- a/lnbits/settings.py
+++ b/lnbits/settings.py
@@ -245,6 +245,13 @@ class EditableSettings(
             **{k: v for k, v in d.items() if k in inspect.signature(cls).parameters}
         )
 
+    # fixes openapi.json validation, remove field env_names
+    class Config:
+        @staticmethod
+        def schema_extra(schema: dict[str, Any]) -> None:
+            for prop in schema.get("properties", {}).values():
+                prop.pop("env_names", None)
+
 
 class EnvSettings(LNbitsSettings):
     debug: bool = Field(default=False)
--- a/.github/workflows/openapi.yml
+++ b/.github/workflows/openapi.yml
@@ -1,0 +1,27 @@
+name: openapi
+
+on: [push, pull_request]
+
+jobs:
+  check:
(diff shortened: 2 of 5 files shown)


jackstar12:
Fix for current error:
```patch
diff --git a/lnbits/settings.py b/lnbits/settings.py
index adec5951..ae373383 100644
--- a/lnbits/settings.py
+++ b/lnbits/settings.py
@@ -228,6 +228,12 @@ class EditableSettings(
             **{k: v for k, v in d.items() if k in inspect.signature(cls).parameters}
         )
 
+    class Config:
+        @staticmethod
+        def schema_extra(schema: dict[str, Any]) -> None:
+            for prop in schema.get("properties", {}).values():
+                prop.pop("env_names", None)
+
 
 class EnvSettings(LNbitsSettings):
     debug: bool = Field(default=False)
```

dni:
omg give this man a bounty!

lnbits/settings.py
```diff
@@ -245,6 +245,13 @@ def from_dict(cls, d: dict):
             **{k: v for k, v in d.items() if k in inspect.signature(cls).parameters}
         )
 
+    # fixes openapi.json validation, remove field env_names
```
motorina0, line 248: - is there no side effect from removing `env_names`?
- when was `env_names` added and why?

dni: `env_names` is added by BaseSettings class https://github.com/pydantic/pydantic/blob/3703caa3725a2d9190d37f7af7aab0fa2d096fae/pydantic/v1/env_settings.py#L118

it seems to me its just additional metadata, and not uses anywhere else. i did not notice anything testing.

motorina0: got it!

motorina0 (approved):

dni:
closes https://github.com/lnbits/lnbits/issues/1442

PR #1557, opened 2023-03-02
FEAT: Filters for GET requests, add it to GET /payments

dni:
This is still a draft / proof of concept PR based on @jackstar12 PR on usermanager.
https://github.com/lnbits/usermanager/pull/2

- introduce Filters class
- introduce decorator parse_filters
- LIMIT / OFFSET


arcbtc:
This is great, but scary. Can you take a look @callebtc @motorina0, i defer to experts on this

jackstar12:
SQL injection should not be possible (haven't tried) because of pydantic. 
Limit and offset are `int` and the other field values are validated according to the filter model.

callebtc:
Does this break current limit and offset implementations?

jackstar12:
> Does this break current limit and offset implementations?

No, the limit and offset params are still there, just parsed differently. 

I'd also like to add an utility function for generating openapi docs (declare all filger model fields as optional parameters)

dni:
screenshot-1678259188

now it supports filtering inside openapi aswell :muscle:

callebtc:
Congrats on this amazingly well-done PR.

Issue #1547, opened 2023-02-24
settings.lnbtis_disabled_extensions use case


jackstar12:
I was wondering about the usage of `settings.lnbits_disabled_extensions` in the `ExtensionManager`

I assume the use to be manually disabling an extension in the extensions folder (because it is never assigned in the code, just through env)

https://github.com/lnbits/lnbits/blob/f943ca7d5127c92bd7c796d76bcfb5f1e66abfbd/lnbits/extension_manager.py#L55-L99

However, the iteration here seems wrong, because it is iterating through the paths and checking it against extension codes.
https://github.com/lnbits/lnbits/blob/f943ca7d5127c92bd7c796d76bcfb5f1e66abfbd/lnbits/extension_manager.py#L70-L72

A fix for this would be to instead add an if statement before the line
https://github.com/lnbits/lnbits/blob/f943ca7d5127c92bd7c796d76bcfb5f1e66abfbd/lnbits/extension_manager.py#L84
Something like:
`if extension_code not in self._disabled:`

Maybe also rename it in the process? Somethign like `lnbits_ignored_extensions` could suit better as disabled might be misleading with the per user active state of the extension

dni:
LGTM

jackstar12:
> Good catch!
> 
> * `settings.lnbits_disabled_extensions` is now redundant
> * admin users can disable (deactivate) any extension from the `Manage Extensions` UI (the install screen)
> * I think this setting can be removed (@dni please confirm)

I've actually found it useful for disabling an extension that is not installed through a repository.

dni:
:thinking: hm, is it really useful though? i would remove it aswell, because as a non-developer you should never need it, and a a developer its simply enought to mv or remove the extensions from the directory.

can you make a PR removing that setting :)?

PR #1564, opened 2023-03-05
remove lnbits_disabled_extensions

jackstar12:
closes https://github.com/lnbits/lnbits/issues/1547 and corrects behaviour of `is_admin_only` (was previously checking with the folder path and not the extension code)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,8 +30,6 @@ LNBITS_DEFAULT_WALLET_NAME="LNbits wallet"
 # Hides wallet api, extensions can choose to honor
 LNBITS_HIDE_API=false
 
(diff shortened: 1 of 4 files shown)


dni:
can you remove it from .env.template and lnbits/settings.py aswell?

jackstar12:
@dni its also used in the admin ui, should i remove it there aswell?
https://github.com/lnbits/lnbits/blob/c81d3e06c8c4217c1c38a71e24964df0e7b80d2c/lnbits/core/templates/admin/_tab_server.html#L70-L81

motorina0 (approved):
utACK ([untested ACK](https://github.com/bitcoin/bitcoin/issues/6100#issuecomment-98471846))

PR #1566, opened 2023-03-08
update installation.md

bitkarrot:
lnbits-legend -> lnbits for systemd, because cloning repo would create 'lnbits' directory not 'lnbits-legend'


dni (approved):
LGTM

PR #1568, opened 2023-03-08
add version to fastapi app and startup log

dni:
make versions more visible on multiple places.

screenshot-1678261590
screenshot-1678261583


callebtc:
The version should be available to all parts of LNbits. How making it a read-only settings value? @dni

dni:
> The version should be available to all parts of LNbits. How making it a read-only settings value? @dni

done!

Issue #1088, opened 2022-10-27
Nix flake error with `pyScss`


afilini:
**Describe the bug**

Trying to use lnbits with the nix flake file I'm getting the error:

```
building the system configuration...
error: unsupported build system requirement pyScss
```

**To Reproduce**
1. Run `nix build github:lnbits/lnbits` and you should get the error

I'm not sure why `pyScss` was added, but maybe it's not perfectly standard-compliant since [poetry2nix](https://github.com/nix-community/poetry2nix) fails on it.

arcbtc:
Our nix build is broken. The issue is dependencies installed in poetry not being used in the poetry build file, seems to be a poetry update. Our plan is to get the application built as we want then use nix as our quality standard as we come out of beta. I'll make a note in the install guide.

prusnak:
This should have been left open, since the issue was not fixed and I spent 1 hour today trying to figure out what's wrong with my setup. Please reopen ...

arcbtc:
Reopened. Sorry about that 😬. I owe you 1hr.

arcbtc:
Chill. This is an issue for later, as noted in the install guide. Post-beta we will some nix whizz a bounty for fixing our nix build
![image](https://user-images.githubusercontent.com/33088785/222207776-231bff64-05b7-450a-bb1b-05595e3258aa.png)

afilini:
I think the pyScss error is caused by the fact that in nixpkgs it is published under `pythonPackages.pyscss` (note the un-capitalized `s`). Changing `pyproject.toml` to use `pyscss` instead fixes that error, but now I'm getting a different issue about infinite recursion caused by setuptools...

prusnak:
> Changing `pyproject.toml` to use `pyscss` instead fixes that error, but now I'm getting a different issue about infinite recursion caused by setuptools...

You can try removing `pyScss` from `pyproject.toml:57`:

``` patch
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ flake8 = { version = "^6.0.0", python = ">=3.8.1" }
 pylint = { version = "^2.15.10", python = ">=3.7.2" }
 
 [build-system]
-requires = ["poetry-core>=1.0.0", "pyScss"]
+requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.scripts]
```

prusnak:
[...]

PR #1574, opened 2023-03-09
fix flake error: #1088

dni:
fixes flake install issue thanks to https://github.com/lnbits/lnbits/issues/1088

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ flake8 = { version = "^6.0.0", python = ">=3.8.1" }
 pylint = { version = "^2.15.10", python = ">=3.7.2" }
 
 [build-system]
-requires = ["poetry-core>=1.0.0", "pyScss"]
+requires = ["poetry-core>=1.0.0", "pyscss"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.scripts]


afilini:
That's not enough to fix the build, in my tests I also had to override some packages due to the setuptools "bug" (see https://github.com/nix-community/poetry2nix/blob/master/docs/edgecases.md).

But after fixing those, the latest 0.10.2 cannot inherently work with nix, it tries to download the extensions in the folder of the source code, which is in the /nix/store and it's read only. So even if you get it to build it won't run.

I tried hacking something together to store the extensions in a different writable directory (like `/var/lib/lnbits/extensions`) but for some reason even when setting the `PYTHONPATH` lnbits would not pick them up, so I ultimately gave up.

I think in order to make this work the first step would be to add an option to install extensions in a separate path and load them from there, then it can be made to work with nix. I personally don't have time to work on it at the moment.

prusnak:
> I also had to override some packages due to the setuptools "bug"

Sent PR to poetry2nix here: https://github.com/nix-community/poetry2nix/pull/1046

afilini:
I switched back to an older version of lnbits, this is what I have:

```
embit = super.embit.overridePythonAttrs (old: {
  buildInputs = old.buildInputs or [ ] ++ (with pkgs.python39.pkgs; [ setuptools ]);
});
pylightning = super.pylightning.overridePythonAttrs (old: {
  buildInputs = old.buildInputs or [ ] ++ (with pkgs.python39.pkgs; [ setuptools ]);
});
represent = super.represent.overridePythonAttrs (old: {
  buildInputs = old.buildInputs or [ ] ++ (with pkgs.python39.pkgs; [ setuptools ]);
});
lnurl = super.lnurl.overridePythonAttrs (old: {
  buildInputs = old.buildInputs or [ ] ++ (with pkgs.python39.pkgs; [ setuptools ]);
});
sqlalchemy-aio = super.sqlalchemy-aio.overridePythonAttrs (old: {
  buildInputs = old.buildInputs or [ ] ++ (with pkgs.python39.pkgs; [ setuptools ]);
});
sse-starlette = super.sse-starlette.overridePythonAttrs (old: {
  buildInputs = old.buildInputs or [ ] ++ (with pkgs.python39.pkgs; [ setuptools ]);
});
```

prusnak:
> I switched back to an older version of lnbits, this is what I have:

Thanks, created PR out of these too: https://github.com/nix-community/poetry2nix/pull/1049

prusnak (approved):
I think we should merge this.

I have couple of other fixes in my pipeline, but they are waiting for external PRs to be merged, so let's not block this change and I'll send another PR when I am ready.

PR #1575, opened 2023-03-09
Doc: Fix dev guide for getting started with extensions

wazint:
Mostly changing the instructions to clone the example repo instead of copying it.


dni:
LGTM

PR #1576, opened 2023-03-10
Bump shortuuid dependency to fix build on Python 3.10

prusnak:
Old version of shortuuid does `assert sys.version >= "3.5"` which fails on Python 3.10 🤦 

Newer version does not do it, so let's bump the dependency.


dni (approved):
lgtm

PR #1578, opened 2023-03-11
flake: update to latest nixpkgs

prusnak:
After bumping shortuuid in https://github.com/lnbits/lnbits/pull/1576 (allowing us to use the default Python 3.10) and overriding build-systems in poetry2nix in https://github.com/nix-community/poetry2nix/pull/1046 and https://github.com/nix-community/poetry2nix/pull/1049 we can now successfully build lnbits via nix flake. All we need to do is to update the flake.lock file which this PR does.

Caveat: `nix run` still is not able to successfully launch lnbits, because the new approach to extensions want to install files in read-only location. I opened an issue for this: https://github.com/lnbits/lnbits/issues/1579

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "flake-utils": {
       "locked": {
-        "lastModified": 1656928814,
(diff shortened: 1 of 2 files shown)


prusnak:
Can you please @afilini confirm this does build LNbits for you? (Running is another issue, see https://github.com/lnbits/lnbits/issues/1579)

Issue #1579, opened 2023-03-11
Make it possible to override the `extensions` path


prusnak:
Currently, the path is hardcoded as `Path(settings.lnbits_path, "extensions")` but I think we should make it possible to override this path the same we we allow to override the data folder via `LNBITS_DATA_FOLDER`.

I propose to introduce `LNBITS_EXTENSIONS_FOLDER` setting which does that (can default to `./extensions`).
This will allow a system administrator to install LNbits into a read-only location (maybe they are using Nix which installs programs to immutable locations) and still allow installation of extensions in `/var/lib/lnbits/extensions` or some other mutable location.

prusnak:
I think @afilini mentioned in https://github.com/lnbits/lnbits/pull/1574#issuecomment-1462149910 that he has some proof of concept for this? Would you please share this with us?

motorina0:
Thanks for the request, it makes sense!
I will try to implement it this week.

prusnak:
Another approach would be to store unpacked extensions in `{LNBITS_DATA_FOLDER}/extensions/` where we already store the downloaded ZIP files.

We would need to update `PYTHON_PATH` to reflect that location for finding the sources (but the same if true even if we had a separate `LNBITS_EXTENSIONS_FOLDER` variable).

dni:
1 more thing to consider ```{LNBITS_DATA_FOLDER}``` is also currently zipped and downloaded when you perform the backup i recently introduced.

i think its fine, to put them in there, they will get a backup aswell and wont need to be redownloaded ;) after restoring.

motorina0:
- I have started by using the `settings.lnbits_data_folder` as a default location for extensions
- the changes are simple, see: https://github.com/lnbits/lnbits/compare/overriede_extensions_path
- however, there is a problem with hardcoded paths in `__init__.py` (`StaticFiles` and `template_renderer`)
```py
tpos_static_files = [
    {
        "path": "/tpos/static",
        "app": StaticFiles(directory="lnbits/extensions/tpos/static"),
        "name": "tpos_static",
    }
]


def tpos_renderer():
    return template_renderer(["lnbits/extensions/tpos/templates"])
```
- making this change means that all extensions have to be touched in order to support it

dni:
is it possible now?

motorina0:
No, not yet.
I have not found a way for loading modules from outside the path of the python process.

HaosGames:
[...]

PR #1580, opened 2023-03-12
refactor extension_manager to use `pathlib.Path` instead of `os.path.join`

prusnak:
While digging around issue https://github.com/lnbits/lnbits/issues/1579 I found out that the extension manager mixes usage of `os.path.join` and `pathlib.Path`

This PR tries to use only the latter.

Draft PR, since this is untested yet.

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -202,7 +202,7 @@ def check_installed_extension(ext: InstallableExtension) -> bool:
         os.path.join(settings.lnbits_data_folder, "extensions", "*.zip")
     )
 
-    if ext.zip_path not in zip_files:
+    if f"./{str(ext.zip_path)}" not in zip_files:
         ext.download_archive()
     ext.extract_archive()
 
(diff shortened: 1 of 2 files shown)


prusnak:
Btw, it seems this code is not covered in the coverage tests (judging by looking at https://app.codecov.io/gh/lnbits/lnbits/pull/1580/blob/lnbits/extension_manager.py)

Is there a way how to do this? This is critical code which is super important to have it covered by tests.

motorina0:
> Btw, it seems this code is not covered in the coverage tests (judging by looking at https://app.codecov.io/gh/lnbits/lnbits/pull/1580/blob/lnbits/extension_manager.py)
> 
> Is there a way how to do this? This is critical code which is super important to have it covered by tests.

I agree. Added: https://github.com/lnbits/lnbits/issues/1584#issue-1632187712

motorina0:
Merged, thank you!

PR #1583, opened 2023-03-17
simplify exclude syntax for mypy in pyproject.toml

prusnak:
we can use the toml syntax and don't have to resort to crazy cfg syntax


dni (approved):
LGTM

PR #1589, opened 2023-03-23
Draft: Fix connection to cliche invoices stream

Bashy:
Currently cliché's invoice stream is using `websocket` module, which is synchronous, and can't be use with `await`.

`2023-03-23 17:25:46.31 | ERROR | lnbits.wallets.cliche:paid_invoices_stream:165 | lost connection to cliche's invoices stream: 'object WebSocket can't be used in 'await' expression', retrying in 5 seconds
`
raised by `ws = await create_connection(self.endpoint)`

I suggest this fix using `websockets` module, which provides async coroutine, thus, it's possible to wait (`await`) for new messages from the cliche's websocket connection passively, without impact.

### DIFF
--- a/lnbits/wallets/cliche.py
+++ b/lnbits/wallets/cliche.py
@@ -5,6 +5,7 @@ from typing import AsyncGenerator, Optional
 
 from loguru import logger
 from websocket import create_connection
+from websockets import connect
 
 from lnbits.settings import settings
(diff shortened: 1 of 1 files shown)


dni:
this will be closed by https://github.com/lnbits/lnbits/pull/1619

PR #1590, opened 2023-03-28
FEAT: use versioning for frontend (npm) and copy it to lnbits/static/vendor for easier updating

dni:
## SIDE EFFECTS
- <strike>the install process now involves 1 more step `npm install`</strike>
- added command in Makefile to copy files from node_modules into static/vendor
- the bundling step for vendor assets is now removed from the runtime and only run with `poetry install`

## PROS
- <strike>cleaner git repository / history</strike>
- easier dependency management, updating and adding
- <strike>faster search inside IDEs</strike>

## CHANGES
-  <strike>remove static/vendor</strike>
- add npm dependencies
- add bolt11-decoder
- <strike>run npm install inside docker image</strike>
- <strike>only use bundle.js and bundle.css</strike>
- <strike>build.py, use node_modules for bundling vendor assets</strike>

## TODO
- <strike>update documentation</strike>
- <strike>update nix flake</strike>


dni:
@prusnak we probably have to change the flake.nix aswell, right? it is now necassary to run `npm install before` we are running `poetry install`

prusnak:
> @prusnak we probably have to change the flake.nix aswell, right? it is now necassary to run `npm install` before we are running `poetry install`

Yes. I can do it in another PR when this is merged.

motorina0 (approved):
I would have not put this on the top of the list, but if its already done we can as well merge it

lnbits/static/vendor/vue-router.js
```diff
@@ -1,13 +1,13 @@
 /*!
-  * vue-router v3.4.3
```
motorina0, line 2: I was expecting for the version to be the same during the upgrade phase.

callebtc:
Post-merge comment: We have revised our approach: 

- We will get rid of the entire python vendor build process from `pyproject.toml` (and `build.py`)
- Instead, we will use `npm` for bundling and minifying, it has a state of the art bundler and an actually working minifier (next PR)
- We will add the steps for npm bundling and minifiying and copying everything to `/vendor` in the the `Makefile`
- LNbits will still serve unbundled, unminified files in `/vendor` if `DEBUG=true`, else it will serve the minified bundle created previously by npm

This is good because nobody needs to touch npm if you 1) install lnbits, 2) work on lnbits. 

Only 3) if you update/add/remove packages you would need to do so with npm (which is already the case today), and then use the `Makefile` to create the bundle once and commit changes to the repo before we ship the update.

This is a very good compromise and I think also the optimal solution.

Issue #1425, opened 2023-01-27
[BUG] Trying to pay already used LN invoice: 500 Internal Server Error


andrejpekarcik:
**Describe the bug**
I paste the LN invoice (outbound payment), click Pay and invoice is correctly processed and payed. I paste once more the same invoice and I get 500 Internal Server Error.

**To Reproduce**
1. Go to your lnbits wallet.
2. Click on PASTE REQUEST and paste your LN invoice
3. Click READ then PAY
4. LN invoice is payed
5. Click PASTE REQUEST and paste the same LN invoice
6. Click READ then PAY
7. 500 Internal Server Error

**Expected behavior**
I expected message 'LN invoice is already paid'

**Screenshots**
![Captura desde 2023-01-27 14-04-44](https://user-images.githubusercontent.com/11619221/215097189-a32060b2-423c-4f18-b671-5b98c814f8c1.png)

**Additional context**
https://legend.lnbits.com exhibits same bug. Commit version: 45a0657828f2dc2bd772c57a20af4df5670c4ee6

dni:
can confirm the issue, paying an external invoice twice.
![screenshot-1680037635](https://user-images.githubusercontent.com/1743657/228367162-3a1521ae-cc9c-4322-b1bd-d956f23b19ce.jpg)

dni:
i think the issue is resolved in those to PRs i have open linked to this issue, hopefully merged soon :) thanks for creating an issue!

PR #1593, opened 2023-03-28
BUG: proper exception for paid interal invoices

dni:
add better exception when internal invoice is paid twice.

debugging this issue https://github.com/lnbits/lnbits/issues/1425

i found that also the exception for internal payment was wrong:
before trying to pay twice:

`HTTPException 520: Payment failed: Only internal invoices can be used!`

now if you try to pay twice:
`HTTPException 520: Internal invoice already paid.`

### DIFF
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -590,18 +590,18 @@ async def delete_wallet_payment(
 
 async def check_internal(
     payment_hash: str, conn: Optional[Connection] = None
-) -> Optional[str]:
+) -> Optional[tuple[str, bool]]:
     row = await (conn or db).fetchone(
         """
-        SELECT checking_id FROM apipayments
-        WHERE hash = ? AND pending AND amount > 0
+        SELECT checking_id, pending FROM apipayments
+        WHERE hash = ? AND amount > 0
         """,
         (payment_hash,),
     )
     if not row:
         return None
     else:
-        return row["checking_id"]
+        return row["checking_id"], row["pending"]
 
 
 # balance_check
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -42,10 +42,7 @@ from .crud import (
 )
 from .models import Payment
 
-try:
-    from typing import TypedDict
-except ImportError:  # pragma: nocover
-    from typing_extensions import TypedDict
+from typing import TypedDict
 
 
 class PaymentFailure(Exception):
@@ -153,9 +150,12 @@ async def pay_invoice(
             extra=extra,
         )
 
-        # check_internal() returns the checking_id of the invoice we're waiting for
-        internal_checking_id = await check_internal(invoice.payment_hash, conn=conn)
-        if internal_checking_id:
+        internal_check = await check_internal(invoice.payment_hash, conn=conn)
+        if internal_check:
+
+            if not internal_check[1]:
+                raise PaymentFailure("Internal invoice already paid.")
+
             logger.debug(f"creating temporary internal payment with id {internal_id}")
             # create a new payment from this wallet
             await create_payment(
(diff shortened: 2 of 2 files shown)


lnbits/core/crud.py
```diff
@@ -590,18 +590,18 @@ async def delete_wallet_payment(
...
-) -> Optional[str]:
+) -> Optional[tuple[str, bool]]:
     row = await (conn or db).fetchone(
         """
-        SELECT checking_id FROM apipayments
-        WHERE hash = ? AND pending AND amount > 0
+        SELECT checking_id, pending FROM apipayments
+        WHERE hash = ? AND amount > 0
         """,
         (payment_hash,),
```
callebtc, line 599: Do not change an existing crud, add a new one that checks pending here.

lnbits/core/services.py
```diff
@@ -153,9 +148,12 @@ class PaymentKwargs(TypedDict):
             extra=extra,
         )
 
-        # check_internal() returns the checking_id of the invoice we're waiting for
-        internal_checking_id = await check_internal(invoice.payment_hash, conn=conn)
-        if internal_checking_id:
+        internal_check = await check_internal(invoice.payment_hash, conn=conn)
```
callebtc, line 151: unpackgin in python: `x, y = f()`

lnbits/core/services.py
```diff
@@ -153,9 +148,12 @@ class PaymentKwargs(TypedDict):
...
+        if internal_check:
+
+            if not internal_check[1]:
```
callebtc: Very bad style

internal check is treaded as a bool and a tuple at the same time, and then indexed with `[1]` to make sure that the next person coding this doesn't what's going on.

lnbits/core/services.py
```diff
@@ -181,27 +179,27 @@ class PaymentKwargs(TypedDict):
         assert wallet
         if wallet.balance_msat < 0:
             logger.debug("balance is too low, deleting temporary payment")
-            if not internal_checking_id and wallet.balance_msat > -fee_reserve_msat:
+            if not internal_check and wallet.balance_msat > -fee_reserve_msat:
```
callebtc, line 183: do not change these parts

lnbits/core/services.py
```diff
@@ -181,27 +179,27 @@ class PaymentKwargs(TypedDict):
...
+                checking_id=internal_check[0], pending=False, conn=conn
             )
 
         # notify receiver asynchronously
         from lnbits.tasks import internal_invoice_queue
 
-        logger.debug(f"enqueuing internal invoice {internal_checking_id}")
-        await internal_invoice_queue.put(internal_checking_id)
+        logger.debug(f"enqueuing internal invoice {internal_check[0]}")
+        await internal_invoice_queue.put(internal_check[0])
```
callebtc: do not index a variable like this. unpack properly or bust.

tests/core/views/test_api.py
```diff
@@ -170,16 +170,15 @@ async def test_pay_invoice_invoicekey(client, invoice, inkey_headers_from):
...
 @pytest.mark.asyncio
-@pytest.mark.skipif(is_regtest, reason="this only works in fakewallet")
 async def test_pay_invoice_adminkey(client, invoice, adminkey_headers_from):
     data = {"out": True, "bolt11": invoice["payment_request"]}
     # try payment with admin key
     response = await client.post(
         "/api/v1/payments", json=data, headers=adminkey_headers_from
     )
-    assert response.status_code < 300  # should pass
+    assert response.status_code > 300  # should fail
```
callebtc, line 181: Is this a related change?

dni: yes the test fails, because it tries to pay an internal invoice that is not pending.

dni: this was also the reason i skipped this testcase for regtest, because before the paid check, it failed there aswell.

PR #1601, opened 2023-03-31
FEAT: build static files with npm (sass, concat, minify), remove build step from python, include minified bundle files

dni:
this pr removed the build.py and building process of vendored js / css. it is now built statically with npm and the appropriate tools.

minified bundles are now included. also all used library and compiled css are inside /static for enabling debugging without the use of npm.


arcbtc (approved):
I have seen the light. DNI is right, this is a great improvement

motorina0 (approved):
LGTM

PR #1610, opened 2023-04-05
remove boltz_client as dependency

dni:
removin dep from core and moving intp extension


callebtc:
@dni Could you resolve the conflict before we merge this?

PR #1611, opened 2023-04-05
Add a wallet list to topup dialog

### LINKED ISSUE #1436: [Feature request] Dropdown for Topup LNbits walletID
## Description
Yes, kind of problem.
Right now the topup module is based on the copy/paste walletID by the super admin.
So if the LNbits owner have hundreds of users, would be quite hard to identify the specific waleltID for each user.
Also users usually don't remember or write down their walletID, so when the super amin will ask them "which is your walletID" to be able to topup they wouldn't know it.

## Solution proposal
Make a dropdown in that topup module, that will list all the walletID available on that LNbits instance and also listed in alphabetical order, by the alias name of the wallet.

The super admin should be able to manage all the walletIDs in that LNbits instance.

talvasconcelos:
Fixes #1436

Adds a dropdown to select a wallet from all users, to topup

### DIFF
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -68,11 +68,13 @@ async def get_user(user_id: str, conn: Optional[Connection] = None) -> Optional[
         or user["id"] in settings.lnbits_admin_users,
     )
 
+
 async def get_all_users() -> List[User]:
     users = await db.fetchall("SELECT id FROM accounts")
     assert users
-    
+
     return [await get_user(user["id"]) for user in users]
+
 
 # extensions
 # -------
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -71,8 +71,7 @@ async def get_user(user_id: str, conn: Optional[Connection] = None) -> Optional[
 
 async def get_all_users() -> List[User]:
     users = await db.fetchall("SELECT id FROM accounts")
-    assert users
-
+    
     return [await get_user(user["id"]) for user in users]
 
 
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -69,7 +69,7 @@ async def get_user(user_id: str, conn: Optional[Connection] = None) -> Optional[
     )
 
 
-async def get_all_users() -> List[User]:
+async def get_all_users() -> List[Optional[User]]:
     users = await db.fetchall("SELECT id FROM accounts")
     
     return [await get_user(user["id"]) for user in users]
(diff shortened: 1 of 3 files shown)


lnbits/core/crud.py
```diff
@@ -69,6 +69,13 @@ async def get_user(user_id: str, conn: Optional[Connection] = None) -> Optional[
     )
 
 
+async def get_all_users() -> List[User]:
+    users = await db.fetchall("SELECT id FROM accounts")
+    assert users
```
motorina0, line 74: - I think `fetchall` returns an empty list if noting found. So this will always be true

motorina0:
Returning all the `user IDs` to the front-end is a scary thought.
The `user ID` is not actually needed, right?
Isn't it better to return only the `wallet IDs`?

lnbits/core/crud.py
talvasconcelos: fixed

callebtc:
I think the privacy implication is ok since it's only available to the super_user but I don't see why this would be very useful. If you know the user ID, you could just as well go to that users wallet and top it up that way. 

Also, for larger deployments, it could put a burden on the DB to get all users every time the dropdown is loaded in the frontend.

Soft cNACK from me.

talvasconcelos:
@Darth-Coin and @arbadacarbaYK

Darth-Coin:
@callebtc I understand that can be a burden on the DB.
Then come the question: how the super_admin can know a userID if it was created by another admin?
Super admin will have to login then as admin to check the userID and then go back to his login to make the topup.
At least make it as a search box if the drop down is too heavy. Searching by the user name or something that more human readable.

We need that restriction of using the user manager ext only for specific admins or even only for super_admin I would say.

arbadacarbaYK:
Privacy not the problem here, i also see the data load a problem. 

Maybe we can prevent it like this and also sort out Darths concern

show two comboboxes, one for admin and the second showin user<->wallet combinations
Nothing selected per default.

dni (changes requested):
i don't like it aswell, if the feature is needed i would go for a usersearch like darth said aswell, not a dropdown or something with prefetched data.

arbadacarbaYK:
Jjust fetch first when someone uses the "user" combo  then the amount of data load for the second for wallets is already reduced, too. Some just open the extension and do nothing so would stay with all off and nothing prefetched. 

The issue was made because of the amount of requests for it in the lnbits group :)

talvasconcelos:
Closed as not viable!

PR #1612, opened 2023-04-05
Fix admin ext access

motorina0:
### Summary
- the admin has the option of making some extensions `Admin Only`:
  - image

- when a non-admin tries to access one of these extensions, it will receive an error message:
  - image

- access is also denied on the API level (but in JSON format):
  - image

- the admin extension should not be visible on the `Extensions` page or on the left hand-side extension list

- this rule does not apply for public pages (where no `usr` is present)
  - for example, `nostrrelay` can be an admin extension, but the public page for paying for access should be accessible to all

### DIFF
--- a/lnbits/middleware.py
+++ b/lnbits/middleware.py
@@ -67,6 +67,7 @@ class InstalledExtensionMiddleware:
         if "query_string" not in scope:
             return True
 
+        # parse the URL query string into a `dict`
         q = parse_qs(scope["query_string"].decode("UTF-8"))
         user = q.get("usr", [""])[0]
         if not user:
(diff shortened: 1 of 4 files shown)


dni (approved):

lnbits/middleware.py
```diff
@@ -51,6 +61,47 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+        user = q.get("usr", [""])[0]
+        if not user:
+            return True
+
+        if user == settings.super_user or user in settings.lnbits_admin_users:
+            return True
+
+        return False
+
+    def _response_by_accepted_type(
```
callebtc, line 80: Could you add a comment here what this method does? It's a bit cryptic.

lnbits/middleware.py
```diff
@@ -51,6 +61,47 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
 
         await self.app(scope, receive, send)
 
+    def _user_allowed_to_extension(self, ext_name: str, scope) -> bool:
+        if ext_name not in settings.lnbits_admin_extensions:
+            return True
+        if "query_string" not in scope:
+            return True
+
+        q = parse_qs(scope["query_string"].decode("UTF-8"))
```
callebtc, line 70: This could also use a comment

callebtc (approved):
LGTM, please add a few comments for the archeologists that will read this code in 1000 years.

lnbits/middleware.py
motorina0: done

PR #1616, opened 2023-04-09
FEAT: lock name inside package-lock, by adding name to package.json

dni:
FEAT: lock name inside package-lock, by adding name to package.json


callebtc (approved):
LGTM

PR #1613, opened 2023-04-07
Serverside Pagination for payments

jackstar12:
This PR moves the pagination, sorting and searching of payments from the frontend to the backend, boosting efficiency.

Note that this is a breaking change to the payments api because the total amount of records has to be returned aswell for pagination to work.

Essential Additions:
- `fetch_page` utility to paginate a query.
- `FromRowModel` is a convenient class to declare a model can be parsed from a row (used by `fetch_page` to parse rows)
- `FilterModel` allows models to declare fields that can be searched by (included in openapi docs aswell)

PR #1618, opened 2023-04-10
Filter broke get_payments on extensions

talvasconcelos:
added `**kwargs` to the function params

### DIFF
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -352,6 +352,7 @@ async def get_payments(
     exclude_uncheckable: bool = False,
     filters: Optional[Filters[Payment]] = None,
     conn: Optional[Connection] = None,
+    **kwargs,
 ) -> List[Payment]:
     """
     Filters payments to be returned by complete | pending | outgoing | incoming.


dni:
where and how does it break? @jackstar12

talvasconcelos:
> where and how does it break? @jackstar12

Here for example: https://github.com/lnbits/lndhub/blob/47ca0ac1b952dde6e4304264df1d680bd025f693/views_api.py#L121

`limit` and `offset` are not parameters of `get_payment`

dni:
ok so we should probably, introduce a /v2/ of the api, use that in frontend, and revert the code to the original /v1/ 
@jackstar12

jackstar12:
> ok so we should probably, introduce a /v2/ of the api, use that in frontend, and revert the code to the original /v1/ @jackstar12

I could include that in https://github.com/lnbits/lnbits/pull/1613

arcbtc:
V2 for entire api? Sounds dramatic, surely this can be fixed without?

callebtc:
Disagree, we should not make a new endpoint but fix the old one can keep it 100% backwards compatible. That was the goal of the Filter PR but seems like it didn't work. Lets fix it.

callebtc:
Adding `**kwargs` to the params, doesn't change anything at all, it just makes the error go away. The bug is actually in the lndhub extension that needs to use `Filters`.

callebtc:
Here is the issue we need to fix: https://github.com/lnbits/lndhub/issues/7

jackstar12:
> Disagree, we should not make a new endpoint but fix the old one can keep it 100% backwards compatible. That was the goal of the Filter PR but seems like it didn't work. Lets fix it.

@callebtc 100% backwords compatibility wont really be possible for https://github.com/lnbits/lnbits/pull/1613 because it changes the response model so I think keeping the new stuff on a seperate endpoint for now should be the easisest and safest (there might be more stuff that depends on the current api). At some point the old one can be removed / deprecated

arcbtc:
A vote for new api endpoint, then when people complain we point to the pagination endpoint

callebtc:
> A vote for new api endpoint, then when people complain we point to the pagination endpoint

Agreed, I assumed the response remains the same. The endpoint should either not change or be 100% backwards compatible. So let's do a separate endpoint for paginated payments and restore the functionality of the old endpoint to exactly the same as it was before.

Issue #764, opened 2022-07-20
remove websockets module from lnbits/wallets/eclair.py


dni:
after discussion with @callebtc we wonder why websockets is used there, it would be best to remove it.

it throws mypy errors because of https://github.com/aaugustin/websockets/issues/940

[lnbits/wallets/eclair.py](https://github.com/lnbits/lnbits-legend/blob/main/lnbits/wallets/eclair.py#L10)

talvasconcelos:
Eclair uses the API to listen to the wallet! 

Can I close this, as there's a `TODO` note already in the code?

PR #1619, opened 2023-04-12
Fix cliche wallet

### LINKED ISSUE #827: Cliché wallet issues
I see 2 potential issues with [the cliché wallet](https://github.com/lnbits/lnbits-legend/blob/1374aabf926944ebe740eb9833eeaae704711c49/lnbits/wallets/cliche.py):

1. It is using the "human-readable" command format, which makes the implementation bloated  and more error-prone. Probably not a big deal, but the invoice descriptions will break. Instead there is the JSON-RPC interface, in which you instead of `create-invoice --msatoshi 12 --description bla` you write `{"method": "create-invoice", "params": {"msatoshi": 12, "description": "bla"}}`.
2. It is opening a websocket, writing a  command and reading the response immediately after. That is not safe if other commands are being called simultaneously. Cliché will write _all responses_ and also _event notifications_ to _all_ websockets that are open at _anytime_. So to filter for the response you want you must check the id in the response object.

Proposed fix:
- Open a single websocket.
- Write commands to it, Assign a random id to each command called (`{"method": "...", "params": {...}, "id": "892173"}`). Keep track of the id and wait.
- Listen to everything that comes from it. If it has an id that is known, wake up the method that was waiting for it so it can proceed. If it is an invoice paid notification, wake up the paid invoices listener.

I actually also do not know how to do this. I was trying just now, but I realized it would take me about 6 months to learn how to implement the flow above using `asyncio`, so I decided to just open this issue.

talvasconcelos:
Major refactor of ClicheWallet, needs a proper python magician to review!

Possible fix #827

### DIFF
--- a/lnbits/wallets/cliche.py
+++ b/lnbits/wallets/cliche.py
@@ -1,12 +1,12 @@
 import asyncio
 import hashlib
 import json
+from asyncio import Future
 from typing import AsyncGenerator, Optional
 
 from loguru import logger
-from websocket import create_connection
+from websockets.client import connect
 
-from lnbits.helpers import urlsafe_short_hash
 from lnbits.settings import settings
 
 from .base import (
@@ -25,17 +25,25 @@ class ClicheWallet(Wallet):
         self.endpoint = settings.cliche_endpoint
         if not self.endpoint:
             raise Exception("cannot initialize cliche")
+        self.next_id = 0
+        self.futures = {}
+        self.ws = None
+
+    async def send_command(self, method: str, params: Optional[dict] = None):
+        self.next_id += 1
+        command = {"id": self.next_id, "method": method}
+        if params is not None:
+            command["params"] = params
+
+        self.ws.send(json.dumps(command))
+        future = Future()
+        self.futures[self.next_id] = future
+        data = await future
+        return data
 
     async def status(self) -> StatusResponse:
-        id = urlsafe_short_hash()
-        try:
-            ws = create_connection(self.endpoint)
-            ws.send({"id":id,"method":"get-info"})
-            r = ws.recv()
-        except Exception as exc:
-            return StatusResponse(
-                f"Failed to connect to {self.endpoint} due to: {exc}", 0
-            )
+        r = await self.send_command(method="get-info")
+
         try:
             data = json.loads(r)
         except Exception:
@@ -53,7 +61,6 @@ class ClicheWallet(Wallet):
         unhashed_description: Optional[bytes] = None,
         **kwargs,
     ) -> InvoiceResponse:
-        id = urlsafe_short_hash()
         if unhashed_description or description_hash:
             description_hash_str = (
                 description_hash.hex()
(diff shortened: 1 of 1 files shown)


lnbits/wallets/cliche.py
```diff
@@ -1,10 +1,12 @@
 import asyncio
 import hashlib
 import json
+from asyncio import Future
 from typing import AsyncGenerator, Optional
 
 from loguru import logger
-from websocket import create_connection
+from websocket import WebSocketApp
+from websockets.client import connect
```
dni, line 9: we do we need 2 different libraries?

lnbits/wallets/cliche.py
```diff
@@ -157,19 +184,30 @@ async def get_payment_status(self, checking_id: str) -> PaymentStatus:
...
 
+    async def on_message(self, message: str):
+        try:
+            data = json.loads(message)
+            msg_id = data.get("id")
+            if msg_id is not None and msg_id in self.futures:
+                future = self.futures.pop(msg_id)
+                future.set_result(data)
+            elif data.get("result", {}).get("status"):
+                yield data["result"]["payment_hash"]
```
dni, line 195: isn't it possible to fall through here without anything happening in if, elif?

lnbits/wallets/cliche.py
```diff
@@ -157,19 +184,30 @@ async def get_payment_status(self, checking_id: str) -> PaymentStatus:
...
+        except Exception as exc:
+            logger.exception(f"Error processing message: {exc}")
+    
+    def on_error(self, error):
+        logger.warning(f"Error from websocket: {error}")
+    
+    def on_close(self):
+        logger.error("Websocket closed")
+        self.ws.close()
```
dni, line 204: when is this triggered?

dni (commented):
overall, still some mypy, pyright checks to fix! :) run `make pyright`, `make mypy`, `make flake8`

dni:
also found this older, issues, can you look at this aswell?
https://github.com/lnbits/lnbits/issues/764

dni:
this PR is also related https://github.com/lnbits/lnbits/pull/1589

lnbits/wallets/cliche.py
talvasconcelos: it's a WIP! websockets to try async, `WebSocketApp` has hability to pass `on_message` method. will fix

lnbits/wallets/cliche.py
talvasconcelos: as intended i guess! if it's not a paid invoice message, or a response from a specific command sent to cliche, we don't care for it!

lnbits/wallets/cliche.py
talvasconcelos: here (line 33):
`async def connect(self):
        self.ws = WebSocketApp(
            self.endpoint,
            on_message=self.on_message,
            on_error=self.on_error,
            on_close=self.on_close,
        )`

lnbits/wallets/cliche.py
Bashy: While I understand that using 2 different libraries is a problem, i think websockets is more suitable that websocket in a coroutine app like fastapi as websockets library provides coroutines natively

motorina0:
@talvasconcelos is this PR still relevant?

Bashy:
> @talvasconcelos is this PR still relevant?

In my opinion yes, the following issue #1655 is still present and makes lnbits unusable out-of-the-box with current cliche wallet code.

dni:
closes https://github.com/lnbits/lnbits/issues/827

dni (approved):

dni:
not maintained

PR #1628, opened 2023-04-16
FEAT: add PYPI python package release workflow on tag

dni:
- add pypi workflow, pyproject types, package metadata
- ignore python package build
- additional metadata for pypi package

for the workflow to properly work we need to set to `PYPI_API_KEY` inside the github repos secrets.

i released current main just for testing, i can transfer ownership to a lnbits pypi account, if needed

https://pypi.org/project/lnbits/


callebtc:
We don't release any pip packages, so this is not really applicable.

dni:
https://github.com/dni/lnbits-pip-setup

motorina0 (approved):

PR #1630, opened 2023-04-17
FEAT: improve on flake8 linting

dni:
- move flake config from setup.cfg to .flake8, to not confuse it with setup.cfg python package settings
- fix E266
- E265
- E241, E501
- E203, E221
- E471
- reduce max string length


arcbtc (approved):
Just formatting,  nothing to worry about

PR #1651, opened 2023-04-24
generate .super_user file on startup if adminui is enabled

dni:
this is very useful for lnbits bundlers, to have easy access to the super_user url to show it to their users

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -447,6 +447,10 @@ async def check_admin_settings():
         admin_url = f"{settings.lnbits_baseurl}wallet?usr={settings.super_user}"
         logger.success(f"✔️ Access super user account at: {admin_url}")
 
+        # saving it to .super_user file
+        with open(".super_user", "w") as file:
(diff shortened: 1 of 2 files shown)


lnbits/core/services.py
```diff
@@ -447,6 +447,10 @@ async def check_admin_settings():
         admin_url = f"{settings.lnbits_baseurl}wallet?usr={settings.super_user}"
         logger.success(f"✔️ Access super user account at: {admin_url}")
 
+        # saving it to .super_user file
+        with open(".super_user", "w") as file:
+            file.write(admin_url)
```
motorina0, line 452: - `admin_url` will always be of the form: `http://127.0.0.1:5000/wallet?usr=xxx...xxx`, but localhost will rarely be used directly.
- in order to avoid parsing the URL for getting the `super_user` it could only save the `super_user` value in the file

dni: 0c98551d

callebtc (approved):

PR #1653, opened 2023-04-24
[FEAT] Auth, Login, create account with username and password

dni:
# new login and accounts based authentication

- introducing session based auth for browsers and and jwt bearer access-tokens for api
- backwards compatible with, ?usr get params and x-api-key

## notes
- there should be a road to drop support for the old authentications style or some kind of new version without that.
- user.id is encoded inside the bearer token

## todo

- [x] remove usr from nodemanagment
- [x] fix 3 failing tests


bitkarrot:
looking like good stuff here @dni 

can we haz Nostr login? or maybe later.....  :D

motorina0:
This PR uses [fastapi_login](https://github.com/tiangolo/fastapi/tree/master/fastapi/security) which is a 3rd party library.
At the same time [fastap.security](https://github.com/MushroomMaula/fastapi_login) offers the same functionality.

I believe we need a good reason for bringing in a new library instead of using the standard one.

motorina0:
There should be a field in settings: `lnbits_allowed_auth_methods=`. Possible values:
 - `user_id` (default, what it is now. Necessary for backwards compability)
 - `email` (email and password)
 - `nostr` (NIP-07, browser extension recommended)
 - `google` (supported by OAuth2)
 - `twitter` (supported by OAuth2)
 - ...

Eg:
```env
# lnbits_allowed_auth_methods=user_id
# lnbits_allowed_auth_methods=email
lnbits_allowed_auth_methods=email,nostr,twitter

```

jackstar12:
> This PR uses [fastapi_login](https://github.com/tiangolo/fastapi/tree/master/fastapi/security) which is a 3rd party library. At the same time [fastap.security](https://github.com/MushroomMaula/fastapi_login) offers the same functionality.
> 
> I believe we need a good reason for bringing in a new library instead of using the standard one.

fast api login uses fastapi security under the hood. Its just more convenient.
Just like using sqlalchemy can be considered more convenient than using psycopg2 directly.

dni:
> There should be a field in settings: `lnbits_allowed_auth_methods=`. Possible values:
> 
>     * `user_id` (default, what it is now. Necessary for backwards compability)
> 
>     * `email` (email and password)
> 
>     * `nostr` (NIP-07, browser extension recommended)
> 
>     * `google` (supported by OAuth2)
> 
>     * `twitter` (supported by OAuth2)
> 
>     * ...
> 
> 
> Eg:
> 
> ```dotenv
> # lnbits_allowed_auth_methods=user_id
> # lnbits_allowed_auth_methods=email
> lnbits_allowed_auth_methods=email,nostr,twitter
> ```

its a good idea, but it exceeds the scope of this pr. 
this pr is meant to get rid of the ?usr inside url.  we can build upon this PR further down the line.

motorina0:
> fast api login uses fastapi security under the hood. Its just more convenient - but I dont know about the exact differences, Just like using sqlalchemy can be considered more convenient than using psycopg2 directly.

Ok, but any 3rd party library has some risk&disadvantages compared with the base framework, some examples:
 - it will not be maintained after a while
 - fewer people know how to use it
 - it does not allow access to some low-level APIs
 - footprint (an extra installed library)
 - maintenance (extra lib to upgrade)
 - ...

I'm not saying A or B is better, but if A is the built-in default and we go with B then we need some good reasons to do so.

motorina0:
> its a good idea, but it exceeds the scope of this pr.
> this pr is meant to get rid of the ?usr inside url. we can build upon this PR further down the line.

- it is important to maintain backwards compatibility, some people will not want to switch to `email+pass` login right away
- other users might want to ditch immediately `user_id` only auth and only use `email+pass`, they need to specify this somehow
- I agree that `nostr`, `google`, `twitter` can come with a second PR

lnbits/core/views/api.py
```diff
@@ -135,12 +135,9 @@ async def api_wallet(wallet: WalletTypeInfo = Depends(get_key_type)):
...
 ):
-    await update_wallet(wallet.wallet.id, new_name)
-    return {
-        "id": wallet.wallet.id,
-        "name": wallet.wallet.name,
-        "balance": wallet.wallet.balance_msat,
-    }
+    new_wallet = await create_wallet(user_id=wallet.wallet.user, wallet_name=new_name)
+    logger.debug(f"Created wallet {new_wallet.id} of user {wallet.wallet.user}")
+    return new_wallet
```
callebtc, line 140: Is this related to this PR? What does it do?

arcbtc:
> > its a good idea, but it exceeds the scope of this pr.
> > this pr is meant to get rid of the ?usr inside url. we can build upon this PR further down the line.
> 
> * it is important to maintain backwards compatibility, some people will not want to switch to `email+pass` login right away
> * other users might want to ditch immediately `user_id` only auth and only use `email+pass`, they need to specify this somehow
> * I agree that `nostr`, `google`, `twitter` can come with a second PR

Agreed, should be transitional. I also hate the idea of being the only lightning wallet you have to log in to, have discussed with DNI and session token will last forever unless user uses log out

arcbtc:
I am also strongly against ditching regular key authentication completely. LNbits makes it easy to build on, and that should continue. With nodes and the like using key authentication, its not something we need ditch.

We have spoken about being able to optionally turn off key authentication.

lnbits/core/models.py
```diff
@@ -79,6 +80,17 @@ class WalletTypeInfo:
     wallet: Wallet
 
 
+class createUser(BaseModel):
```
motorina0, line 83: - class names must start with upper case

lnbits/core/views/user_api.py
```diff
@@ -0,0 +1,86 @@
...
+async def user(user=Depends(login_manager)) -> User:
+    return user
+
+
+@user_router.post(
+    "/api/v1/login", description="Login to the API via the username and password"
+)
+async def login_endpoint(
+    response: Response,
+    username: str = Body(),
```
motorina0, line 23: - should be called `email`. See line 35: `user = await get_account_by_email(username)`

lnbits/core/models.py
```diff
@@ -88,6 +100,14 @@ class User(BaseModel):
     admin: bool = False
     super_user: bool = False
 
+    def login(self, password: str) -> bool:
+        pwd_bytes: bytes = password.encode("utf-8")
+        if not self.password:
+            raise Exception("User does not have a password.")
```
motorina0, line 106: - information leak. It should also say `"Incorrect login."`. Not more.

lnbits/decorators.py
```diff
@@ -220,8 +243,20 @@ async def require_invoice_key(
...
+        g().user = user
+        return user
+
+    if not usr:
+        raise HTTPException(
+            status_code=HTTPStatus.NOT_FOUND,
+            detail="Not logged in or provided ?usr argument.",
+        )
+
+    g().user = await get_user(usr)
```
motorina0, line 259: - why is this `g().user` required, why not use a local variable?

motorina0: - the code is fragile, is a new `async` call is introduced then race conditions will occur
- by running this code:
```py
    try:
        g_user_id = g().user.id if g().user else None
        print('### before [{id, g_id]',usr, g_user_id)
    except:
        pass

    g().user = await get_user(usr) # global?
    await asyncio.sleep(1)

    g_user_id = g().user.id if g().user else None
    print('### after [{id, g_id]', usr == g_user_id, usr, g_user_id)
```

 - I have managed to get into an inconsistent state
```log
### after [{id, g_id] False bf590154da614dbc943bb1fed4cf2661 f9270eec46404a38a1fda2e8af0df34a
```

lnbits/decorators.py
```diff
@@ -241,8 +276,8 @@ async def check_user_exists(usr: UUID4) -> User:
     return g().user
 
 
-async def check_admin(usr: UUID4) -> User:
-    user = await check_user_exists(usr)
+async def check_admin(req: Request, usr: Optional[str] = None) -> User:
```
motorina0, line 279: - why not use `UUID4` for `usr`?
```suggestion
async def check_admin(req: Request, usr: Optional[UUID4] = None) -> User:
```

lnbits/decorators.py
```diff
@@ -252,8 +287,8 @@ async def check_admin(usr: UUID4) -> User:
     return user
 
 
-async def check_super_user(usr: UUID4) -> User:
-    user = await check_admin(usr)
+async def check_super_user(req: Request, usr: Optional[str] = None) -> User:
+    user = await check_admin(req, usr)
```
motorina0, line 291: - why not use UUID4 for usr?
```suggestion
async def check_super_user(req: Request, usr: Optional[UUID4] = None) -> User:
```

lnbits/decorators.py
```diff
@@ -252,8 +287,8 @@ async def check_admin(usr: UUID4) -> User:
     return user
 
 
-async def check_super_user(usr: UUID4) -> User:
-    user = await check_admin(usr)
+async def check_super_user(req: Request, usr: Optional[str] = None) -> User:
```
motorina0, line 290: - why not use UUID4 for usr?
```suggestion
async def check_super_user(req: Request, usr: Optional[UUID4] = None) -> User:
```

lnbits/core/crud.py
```diff
@@ -28,12 +29,45 @@
...
     Wallet,
     WebPushSubscription,
+    createUser,
 )
 
 # accounts
 # --------
 
 
+async def create_user(data: createUser) -> User:
```
motorina0, line 39:
```suggestion
async def create_user(data: CreateUser) -> User:
```

lnbits/core/crud.py
```diff
@@ -28,12 +29,45 @@
     User,
     Wallet,
     WebPushSubscription,
+    createUser,
```
motorina0, line 32:
```suggestion
    CreateUser
```

lnbits/core/views/user_api.py
```diff
@@ -0,0 +1,86 @@
+from typing import Optional
+
+from fastapi import APIRouter, Body, Depends, Response, status
+from fastapi.responses import JSONResponse
+
+from ..crud import create_user, get_account_by_email, get_user
+from ..models import User, createUser
```
motorina0, line 7:
```suggestion
from ..models import User, CreateUser
```

lnbits/core/views/user_api.py
```diff
@@ -0,0 +1,86 @@
...
+    "/api/v1/login", description="Login to the API via the username and password"
+)
+async def login_endpoint(
+    response: Response,
+    username: str = Body(),
+    password: str = Body(),
+    usr: Optional[str] = Body(None),
+) -> JSONResponse:
+    if usr:
+        user = await get_user(usr)
```
motorina0, line 28: - a request with the body as listed below will always produce a valid access token (if `usr` exists)
- this means that `usr` has to remain forever secret

```json
{
    "username":"xxx",
    "password": "fake",
    "usr": "bf885e9c94084489a62c1e730dd3b389"
}
```

lnbits/core/views/user_api.py
```diff
@@ -0,0 +1,86 @@
...
+        return JSONResponse(
+            {
+                "detail": [
+                    {
+                        "loc": ["body", "password_repeat"],
+                        "msg": "passwords do not match",
+                    }
+                ]
+            },
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
```
motorina0, line 72: - why not 401?
```suggestion
            status_code=status.HTTP_401_UNAUTHORIZED,
```

lnbits/core/crud.py
```diff
@@ -28,12 +29,45 @@
...
+async def create_user(data: createUser) -> User:
+    if await get_account_by_email(data.email):
+        raise Exception("user exists")
+
+    pwd_bytes = data.password.encode("utf-8")
+    salt = gensalt()
+    password = hashpw(pwd_bytes, salt)
+
+    user_id = uuid4().hex
+    await db.execute(
```
motorina0, line 48: - should not allow duplicate emails

lnbits/core/static/js/index.js
```diff
@@ -7,10 +7,61 @@ new Vue({
         show: false,
         data: {}
       },
-      walletName: ''
+      usr: '',
+      username: '',
+      password: '',
+      password_repeat: '',
```
motorina0, line 13: - cammelCase preferred

lnbits/core/static/js/index.js
```diff
@@ -7,10 +7,61 @@ new Vue({
...
+      walletName: '',
+      signup: false
     }
   },
   methods: {
+    register: function () {
```
motorina0, line 19: - why not use `LNbits.api.request()`?

lnbits/core/static/js/index.js
```diff
@@ -7,10 +7,61 @@ new Vue({
...
+        .then(response => {
+          this.$q.cookies.set('access-token', response.data.access_token)
+          this.$q.localStorage.set('lnbits.token', response.data.access_token)
+          this.$q.localStorage.set('lnbits.usr', response.data.usr)
+          window.location.href = '/wallet'
+        })
+        .catch(LNbits.utils.notifyApiError)
+    },
+    login: function () {
+      axios({
```
motorina0, line 38: - why not use `LNbits.api.request()`?

lnbits/core/static/js/index.js
```diff
@@ -7,10 +7,61 @@ new Vue({
...
+        .then(response => {
+          this.$q.cookies.set('access-token', response.data.access_token)
+          this.$q.localStorage.set('lnbits.token', response.data.access_token)
+          // this.$q.localStorage.set('lnbits.usr', response.data.usr)
+          window.location.href = '/wallet'
+        })
+        .catch(LNbits.utils.notifyApiError)
+    },
+    login_usr: function () {
+      axios({
```
motorina0, line 52: - why not use `LNbits.api.request()`?

lnbits/core/static/js/index.js
```diff
@@ -7,10 +7,61 @@ new Vue({
...
+      })
+        .then(response => {
+          this.$q.cookies.set('access-token', response.data.access_token)
+          this.$q.localStorage.set('lnbits.token', response.data.access_token)
+          // this.$q.localStorage.set('lnbits.usr', response.data.usr)
+          window.location.href = '/wallet'
+        })
+        .catch(LNbits.utils.notifyApiError)
+    },
+    login_usr: function () {
```
motorina0, line 51: - cammelCase prefered

```suggestion
    loginUser: function () {
```

lnbits/core/static/js/index.js
```diff
@@ -7,10 +7,61 @@ new Vue({
...
+        url: '/api/v1/register',
+        data: {
+          email: this.username,
+          password: this.password,
+          password_repeat: this.password_repeat
+        }
+      })
+        .then(response => {
+          this.$q.cookies.set('access-token', response.data.access_token)
+          this.$q.localStorage.set('lnbits.token', response.data.access_token)
```
motorina0, line 31: - is there a `this.$q.localStorage.set('lnbits.token')` somewhere? Where is this used?

lnbits/core/static/js/index.js
```diff
@@ -7,10 +7,61 @@ new Vue({
...
+          this.$q.localStorage.set('lnbits.usr', response.data.usr)
```
motorina0, line 32: - where is this value read?

lnbits/core/views/api.py
motorina0: +1

lnbits/settings.py
```diff
@@ -292,6 +292,9 @@ class EnvSettings(LNbitsSettings):
     lnbits_commit: str = Field(default="unknown")
     super_user: str = Field(default="")
     version: str = Field(default="0.0.0")
+    login_secret: str = Field(
```
motorina0, line 295: - possible security vulnerability if users forget to specify the environment variable
- can it default to a derivation `superuser_id` for example?
  - this way the secret is unique and the same after restarts and also can be recovered from the DB

lnbits/core/views/generic.py
```diff
@@ -164,22 +163,52 @@ async def extensions_install(
...
+    user: User = Depends(check_user_exists),
+    wal: Optional[UUID4] = None,
 ):
-    user_id = usr.hex
-    user = await get_user(user_id)
+    user_id = user.id
 
-    if not user:
+    if (
+        len(settings.lnbits_allowed_users) > 0
```
motorina0, line 172: - this logic could be a method on `settings`

lnbits/decorators.py
```diff
@@ -6,14 +6,37 @@
...
 from lnbits.core.crud import get_user, get_wallet_for_key
 from lnbits.core.models import User, WalletType, WalletTypeInfo
 from lnbits.db import Filter, Filters, TFilterModel
 from lnbits.requestvars import g
 from lnbits.settings import settings
 
+# from pydantic.types import UUID4
+
+
+def is_user(request: Request):
```
motorina0, line 19: - from the name of the function I expect for the return type to be a `Bool`
- this function is in fact a getter

lnbits/decorators.py
```diff
@@ -6,14 +6,37 @@
...
+    user = request.state.user
+    if user is None:
+        raise HTTPException(401)
+    return user
+
+
+def is_admin(request: Request):
```
motorina0, line 26: - has naming convention of boolean function

lnbits/decorators.py
```diff
@@ -6,14 +6,37 @@
...
+    user = is_user(request)
+    if not user.is_admin and not user.super_user:
+        raise HTTPException(401)
+
+    return user
+
+
+def is_super_user(request: Request):
```
motorina0, line 34: - has naming convention of boolean function

lnbits/decorators.py
```diff
@@ -220,8 +243,20 @@ async def require_invoice_key(
...
+    if req.state.user:
+        user = await get_user(req.state.user.id)
+        assert user, "Logged in user has to exist."
+        g().user = user
+        return user
+
+    if not usr:
+        raise HTTPException(
+            status_code=HTTPStatus.NOT_FOUND,
+            detail="Not logged in or provided ?usr argument.",
```
motorina0, line 256: - argument -> query parameter

lnbits/static/js/base.js
```diff
@@ -406,6 +431,15 @@ window.windowMixin = {
     }
   },
   created: function () {
+    // crop usr from url and login that user in the background
+    if (window.location.href.search('usr=') !== -1) {
```
motorina0, line 435:
```js
const urlParams = new URLSearchParams(window.location.search);
const usrParam = urlParams.get('usr');
```

lnbits/decorators.py
dni: i think we should rethink the usage of `g()` altogether, i dont see much point in it

lnbits/core/views/user_api.py
dni: its the `/register` endpoint it is public

lnbits/static/js/base.js
dni: awesome

lnbits/decorators.py
dni: it is a decorator, maybe calling them require_user and so on is better?

lnbits/core/static/js/index.js
dni: `$q.cookies` are sent with the request for auth.
`$q.localstorage` as far is a know ia bonus

lnbits/core/views/user_api.py
dni: my issue with email is, i dont want to expect an email from new users to create an account.

lnbits/core/crud.py
dni: its checked on line, 39 above, what do you mean?

lnbits/core/views/user_api.py
dni: yes. it is meant for backward compatibility

lnbits/core/views/user_api.py
```diff
@@ -0,0 +1,86 @@
...
+                {"error": "not found"}, status_code=status.HTTP_404_NOT_FOUND
+            )
+    else:
+        try:
+            user = await get_account_by_email(username)
+            if not user:
+                return JSONResponse(
+                    {"error": "not found"}, status_code=status.HTTP_404_NOT_FOUND
+                )
+            _ = await load_user(user.id)  # type: ignore
```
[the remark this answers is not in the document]
dni: `load_user` populates `req.state.user` for fastapi login

lnbits/core/views/user_api.py
motorina0: - some users will not want to upgrade to email+pass
- some user will want only email+pass (no user login)
- we do need a setting for the allowed login modes
- this PR needs to be fully backwards compatible
- suggestion:

```py
settings.allowed_auth_methods = ["basic", "email", "pubkey", "..."]
```

- the admin can choose one or more of the above

lnbits/settings.py
dni: you will need to help me here, that is over my head. keep in mind superuser can change

motorina0:
If new users are not allowed (see https://github.com/lnbits/lnbits/pull/2017/) then the `Sign Up` flow should not be allowed (UI and API checks)

lnbits/core/static/js/index.js
motorina0: I mean this one:
```js
          this.$q.localStorage.set('lnbits.usr', response.data.usr)
```

dni: ye it is not needed, because its also available one `window.user.id`

motorina0:
Will be deprecated in favor of https://github.com/lnbits/lnbits/pull/2092

Issue #1661, opened 2023-04-27
[BUG] CRITICAL - unsupported operand type(s) for /: 'str' and 'float' for all payments


Perlover:
**Describe the bug**

If we switch to database mode (from .env mode), and then as admin change the "fee reserve" parameters (base fee + fee reserve in %) - sending payments stop to work!

**To Reproduce**

1. Install and switch to database mode
2. As admin go to "Manage Server" -&gt; FUNDING
3. Try there fill other parameters "Reserve fee in msats" as for example as 0, 1, ... etc and "Reserve fee in percent" as 0, 1, etc
4. Press SAVE Button
5. Open normal user wallet and try to pay any invoice

**Expected behavior**

To pay invoice

**Screenshots**

There is error: https://nostr.build/p/nb6651.png for all payments.

**Desktop (please complete the following information):**
 - LNbits version: Commit version: adae3b4302756f3d75498f8fb05658a9a565aa9c
 - Database sqlite

**Additional context**

callebtc:
This is not a critical bug

callebtc:
What do you mean with "switch to database mode"?

dni:
> @talvasconcelos @dni could this be an issue in the AdminUI?
i would guess it is adminui PUT a string

callebtc:
> > @talvasconcelos @dni could this be an issue in the AdminUI?
> > i would guess it is adminui PUT a string

Can you take over this one?

PR #1662, opened 2023-04-27
Fix unsupported operand type(s) for /: 'str' and 'float' for all payments errors after fee changings by admin

Perlover:
Patch for https://github.com/lnbits/lnbits/issues/1661
It works! I tested!


Perlover:
I rebased my patch to main branch
But I tested on `pull/1636/head` ref before

PR #1663, opened 2023-04-29
FIX: #1661

dni:
cached settings where not typed
https://github.com/lnbits/lnbits/issues/1661


dni:
tested it, and it worked!

motorina0 (approved):
LGTM

PR #1671, opened 2023-05-02
fix cln wallet for msat migration

daywalker90:
This fix should be back and forward compatible for clns msat migration

### DIFF
--- a/lnbits/wallets/cln.py
+++ b/lnbits/wallets/cln.py
@@ -69,7 +69,7 @@ class CoreLightningWallet(Wallet):
         try:
             funds = self.ln.listfunds()
             return StatusResponse(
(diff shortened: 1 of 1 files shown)


dni:
on which version of cln would this fail?

daywalker90:
i'm not sure but probably on some ancient versions (pre 0.10.0)

daywalker90:
i've fixed status to only sum up our sats

dni:
i mean in which future version of cln, our current setup without that PR, will break. it works for 23.02.2 just fine

daywalker90:
23.05

dni (approved):

daywalker90:
do you see this in the logs?
`| WARNING | Retrying connection to backend in 5 seconds...`
then you are in luck this pr will probably fix it for you :)

PR #1673, opened 2023-05-03
add workflow for cln 23 node

dni:
testing latest cln release


callebtc:
Isn't this duplicating the tests?

PR #1678, opened 2023-05-04
Update print.html

talvasconcelos:
Fix print.html not getting the JS libraries


dni:
css changed aswell to included_css

dni (approved):

PR #1688, opened 2023-05-08
Create super user account if it does not exist

motorina0:
### Summary
This PR allows for an external environment to define a value for `SUPER_USER` even if that user_id is not yet present in the system.
The value for `SUPER_USER`  must be a hex string with a minimum length of 32 chars (128 bits).

LNbits will normalize the `SUPER_USER` value such that:
 - its length is 32 chars (128 bits).  Any characters after the 32nd position will be ignored
   - for example the valid hex `d858dc48ab1ef6bc1e5ce27c44b508eea9c73214776b5b38d92b4292d8c961c2` will be normalized to `d858dc48ab1e46bc9e5ce27c44b508ee`
 - it must be a valid UUID4
   - for example the valid hex `ffffffffffffffffffffffffffffffff` will be normalized to `ffffffffffff4fffbfffffffffffffff`

The normalization is required so that a external environments do not have to comply with LNbits, but the other way around.

A user should visit `http://127.0.0.1:5000/uuidv4/{secret_hex}` and it will be redirected to the home page as super admin.

Eg: `http://127.0.0.1:5000/uuidv4/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa` -> `http://127.0.0.1:5000/wallet?usr=aaaaaaaaaaaa4aaaaaaaaaaaaaaaaaaa`

**Preconditions**:
 - `LNBITS_ADMIN_UI=true`

### DIFF
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -2,10 +2,9 @@ import datetime
 import json
 from typing import Any, Dict, List, Optional
 from urllib.parse import urlparse
-from uuid import uuid4
+from uuid import UUID, uuid4
 
 import shortuuid
-from pydantic.types import UUID4
 
 from lnbits import bolt11
 from lnbits.db import COCKROACH, POSTGRES, Connection, Filters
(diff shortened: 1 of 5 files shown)


callebtc (approved):

arcbtc (commented):
Can anyone use this to create the super user, how is access restriced?

dni (approved):

motorina0:
> Can anyone use this to create the super user
- no, the endpoint does not create any new data
- the endpoint "normalizes" (to valid UUID4) a `secret` then does a redirect

PR #1697, opened 2023-05-10
fix: sqlite json

jackstar12:
this pr makes the json behaviour of sqlite3 driver identical to psycopg2 (automatically parse json data to python object) and also removes the usage of the `->>` operator because its only supported by sqlite >= 3.38

Im also wondering if this is ever used because if I'm not mistaken the code might be wrong anyways:
https://github.com/lnbits/lnbits/blob/95071eb87e94481a0b2e5bcd29517eaa1f8c8dee/lnbits/db.py#L50-L58
see: https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.new_type


jackstar12:
I think I should clarify why supporting JSON columns in db would be beneficial.

I'll use the recently added extra column in the https://github.com/lnbits/usermanager (which currently doesnt work on sqlite!) as an example.

It is important that the usermanager api gives a way of filtering by custom data in the extra column.
There are essentially two ways to do this:

    fetch all users and filter application side
    filter directly in sql and only return required rows

The latter option is obviously much more performant, but it requires the column type to be json so that we can write queries accordingly.
It also allows for easier partial updates. With a text column you'd have to select the row, do the update in python and then update. JSON type makes this possible in sql directly.

But it doesnt end here:
If the extra column on payments was changed to JSON type aswell, the api could support filtering by extra data too. So you could easily get all payments with the tag boltz. Something like that isnt possible right now.
If the extra column becomes more viable more extensions could use it to put some metadata alongside the payments, e.g. the discordbot could store the channel id the payment happened in so it can display stats.

motorina0:
Can you please add some unit tests. I would like to understand how to use the new syntax for each of the CRUD operations.

jackstar12:
> Can you please add some unit tests. I would like to understand how to use the new syntax for each of the CRUD operations.

Theres an existing test here:
https://github.com/lnbits/lnbits/blob/c960f718f5baf107cd2d9a079055138e5080b6f7/tests/core/views/test_api.py#LL190C1-L190C13

Altough this does not test filters on JSON you san see how these are used over here
https://github.com/lnbits/discordbot/blob/53125379423e3ae8f0516a0e93515bc79826e702/bot/api.py#L35C7-L41

lnbits/db.py
```diff
@@ -61,6 +63,8 @@ def _parse_timestamp(value, _):
 else:
     if os.path.isdir(settings.lnbits_data_folder):
         DB_TYPE = SQLITE
+        sqlite3.register_converter("json", lambda val: json.loads(val.decode("utf-8")))
+        sqlite3.register_converter("timestamp", lambda val: int(val))
```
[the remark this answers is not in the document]
jackstar12: no, they are native in sqlite (timestamp already used before). 
The problem was that psycopg2 automatically decodes json cols into python objects by default but sqlite3 doesnt. This converter fixes that.
The timestamp converter is there because when `PRASE_DECLTYPES` (see below) is active the default timestamp converter sqlite3 has registered causes trouble. The `timestamp` converter makes sure the behaviour doesnt change.

lnbits/db.py
```diff
@@ -245,7 +249,15 @@ def __init__(self, db_name: str):
         else:
             self.schema = None
 
-        self.engine = create_engine(database_uri, strategy=ASYNCIO_STRATEGY)
+        if DB_TYPE == SQLITE:
+            self.engine = create_engine(
+                database_uri,
+                strategy=ASYNCIO_STRATEGY,
+                connect_args={"detect_types": sqlite3.PARSE_DECLTYPES},
```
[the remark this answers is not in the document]
jackstar12: This argument enables the above registered converters. I could add that as a comment

motorina0:
> Theres an existing test here: https://github.com/lnbits/lnbits/blob/c960f718f5baf107cd2d9a079055138e5080b6f7/tests/core/views/test_api.py#LL190C1-L190C13

That is good, but not complete. The test should include:
 - create a table with `JSON` type column
 - insert **invalid** data into the JSON data -> should raise and exception
 - insert valid data then retrieve it -> should be the same JSON string
 - update JSON object then retrieve it -> should be the updated JSON string
 - query for top level JSON field
 - query for nested JSON field (`aa.bb.cc`)
 - can an INDEX be defined on a JSON field and how?

This is critical logic that will be used by the server and the extensions. We need to make sure it works exactly as expected and that it is not easy to introduce regression bugs.

jackstar12:
Ill add one later today, most importantly we should be testing that the python drivers behave as expected with parsing the outputs / inputs. 

I think testing if updates or nested json queries work wouldnt really add anything because its out of our control anyways and I think its safe to assume that the databases take care of that.

Indexes on standard JSON fields arent possible. Postgres supports through a custom JSONB type (encodes the json into a binary format)

jackstar12:
Added a test and also made the `rewrite_values` parse dicts to json as a convenience feature.

I also tried out removing the above mentioned code https://github.com/lnbits/lnbits/blob/95071eb87e94481a0b2e5bcd29517eaa1f8c8dee/lnbits/db.py#L50-L58.

Nothing breaks and I think it was never used (its probably broken), please correct me if you know anymore

motorina0:
> Nothing breaks and I think it was never used (its probably broken), please correct me if you know anymore

- I do not know more, but it is a good practice for PRs be small and focus on one thing alone (they will be merged faster).

> I think testing if updates or nested json queries work wouldnt really add anything because its out of our control anyways and I think its safe to assume that the databases take care of that.

- it is good to test end-2-end, many things can vary over time (python version, db driver, db version)
- also, I was under the impression that filtering/updating will be supported. See:
> It is important that the usermanager api gives a way of filtering by custom data in the extra column.
- is this the case? If yes, a test (at least to serve as an example) is required.

jackstar12:
You were right.

I added a partial update to the first test with a compatibility helper and more tests around nested filtering pushing the code to some edge cases and I found some small things that broke.

The Filters were already implicitly tested by the payments test but not well enough

tests/core/test_db.py
```diff
@@ -0,0 +1,24 @@
...
+@pytest.mark.asyncio
+async def test_sql_json(db, json_test_table):
+    obj = {"a": 2, "b": 2}
+    await db.execute("INSERT INTO test VALUES(?)", (obj,))
+    row = await db.fetchone("SELECT * FROM test")
+    assert row.data == obj
+    update = {"b": 3}
+    obj |= update
+    await db.execute(f"UPDATE test SET {db.json_partial_update('data')}", (update,))
+    row = await db.fetchone("SELECT * FROM test")
```
motorina0, line 23: - can you please add a test with `WHERE` clause with JSON query?

jackstar12: Ill add an additional helper in `Compat` for that aswell

lnbits/db.py
```diff
@@ -47,20 +49,13 @@ def _parse_timestamp(value, _):
...
-            if value is not None
-            else None,
-        )
-    )
 
     register_type(new_type((1184, 1114), "TIMESTAMP2INT", _parse_timestamp))
 else:
     if os.path.isdir(settings.lnbits_data_folder):
         DB_TYPE = SQLITE
+        sqlite3.register_converter("json", lambda val: json.loads(val.decode("utf-8")))
```
motorina0, line 57: - can `val` be `None`?

tests/core/test_db.py
```diff
@@ -0,0 +1,33 @@
...
+    try:
+        await db.execute("CREATE TABLE test (data JSON)")
+        yield db
+    finally:
+        await db.execute("DROP TABLE test")
+
+
+@pytest.mark.asyncio
+async def test_sql_json(db, json_test_table):
+    obj = {"a": "foo", "b": "bar"}
```
motorina0, line 16: - can you please add a nested field and query for it please

```suggestion
    obj = {"a": "foo", "b": "bar", "c": {"x": "zzz"}}
```

- are queries for numbers possible (using `>`, `<=`, etc)?

jackstar12: numbers are a bit tricky. Postgres makes it a bit difficult here as we would need to explicitly cast the accessed object into a number for such comparison

jackstar12: one solution for this is to add functionality to json_path to specify a type which is needed for postgresqls stronger type system which is no problem for sqlite

lnbits/db.py
jackstar12: I don't think so
https://docs.python.org/3/library/sqlite3.html#sqlite3.register_converter

> The converter is invoked for all SQLite values of type typename; it is passed a [bytes](https://docs.python.org/3/library/stdtypes.html#bytes) object and should return an object of the desired Python type

motorina0: - what happens if the value in the SQL column is `NULL` ?

jackstar12: just tried it out, the converter is not called in that case and it just gives a `None`. should I add that in the test?

dni:
needs more tests

jackstar12:
I had an idea to mitigate one of the issues with the previous approach of `json_path`: easy to inject sql if you dont validate the path.

Instead of directly putting the json path into the query we can use a placeholder and pass the path as a value, so that it is being sanitized. But this introduces the problem that the location of the placeholder in the query and in the values list has to be the same. This would be tedious if done manually which is why a small helper class could be of great help: `QueryValues`.

Its basically a wrapper over a normal list. But when the `json_path` function is called it returns the placeholder and adds the corresponding value to the list (itself) to make sure there is no mismatch. The same should be done with normal value placeholders (see tests).

jackstar12:
@motorina0 
Regarding invalid json:
while postgres throws an exception, sqlite doesnt, it just stores the invalid data. I dont know if it makes sense to test that, as the json generated by  `json.dumps` is valid anyway

jackstar12:
To avoid type issues I had to make the type of `values` in the `fetchall` and `fetchone` calls more accurately represent what is being accepted (any sequence of values)

lnbits/db.py
```diff
@@ -47,27 +49,55 @@ def _parse_timestamp(value, _):
...
     )
-    register_type(
-        new_type(
-            (1082, 1083, 1266),
-            "DATE2INT",
-            lambda value, curs: time.mktime(value.timetuple())
-            if value is not None
-            else None,
-        )
-    )
```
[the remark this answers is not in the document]
jackstar12: this is somewhat unrelated (postgres only), I thought I removed it, but seems like I forgot to.
I had some initial suspicions on it as described above which proved to be true when working on https://github.com/lnbits/lnbits/pull/1756 because psycopg tried to convert postgres dates with our provided function - which failed and resulted in `None` values.

motorina0:
This is an interesting PR, but I'm inclined to reject merging it for these reasons:
 - I don't think it will be used in the near future (next year or so), so we just add complexity to our core codebase with little or no benefit
 - on the mid-term (one year plus) more users will upgrade their SQLite so this feature might not be required
 - it touches very sensitive code that is not well covered by unit tests

@dni what is your input on this?

dni:
i think this feature is a must for the usermanager / discord bot to properly work right now? @jackstar12 right?
it is useful and i think we should go for better test coverage on that to take the scare out.

imo it would be a pity to postpone it.

motorina0:
> imo it would be a pity to postpone it.

@dni can you please also review and Approve/Reject

dni (approved):
tested, reviewed, LGTM

motorina0 (approved):

jackstar12:
> i think this feature is a must for the usermanager / discord bot to properly work right now? @jackstar12 right?
yes, and it could be used in core aswell

> it is useful and i think we should go for better test coverage on that to take the scare out.
patch coverage is already at 98%, github comment somehow shows it wrong

jackstar12:
@motorina0 

> * on the mid-term (one year plus) more users will upgrade their SQLite so this feature might not be required

If this is the case, it would probably make it simpler but not completely useless - at its core `QueryValues` helper is just an easy way to make sure there is no mismatch between placeholders in the query and the values in the list which could be useful event without json.

> * it touches very sensitive code that is not well covered by unit tests

I think current test coverage is pretty good. The modifications in `db.py`, apart from typings, just adapt the json usage in `Filters`  and `fetch_page` accordingly which is tested pretty well on the payments endpoint. I'm not very happy with the state of `Filters` (which I now think were merged a bit prematurely) aswell, since I essentially yanked them from another project of mine. They could need some small tests, some docs (https://github.com/lnbits/lnbits/issues/1830) and a bit of simplification but I dont wanna include that in this pr.

tests/core/test_db.py
```diff
@@ -0,0 +1,58 @@
...
+async def json_test_table(db):
+    try:
+        await db.execute("CREATE TABLE test (data JSON)")
+        yield db
+    finally:
+        await db.execute("DROP TABLE test")
+
+
+@pytest.mark.asyncio
+async def test_sql_json(db, json_test_table):
```
motorina0, line 17: `json_test_table` not used

lnbits/db.py
```diff
@@ -47,27 +49,55 @@ def _parse_timestamp(value, _):
...
+        sqlite3.register_converter("json", lambda val: json.loads(val.decode("utf-8")))
+        sqlite3.register_converter("timestamp", lambda val: int(val))
     else:
         raise NotADirectoryError(
             f"LNBITS_DATA_FOLDER named {settings.lnbits_data_folder} was not created"
             f" - please 'mkdir {settings.lnbits_data_folder}' and try again"
         )
 
 
+class QueryValues(list):
```
motorina0, line 66: Please add doc to this class

tests/core/test_db.py
```diff
@@ -0,0 +1,58 @@
...
+    values = QueryValues()
+    row = await db.fetchone(
+        f"SELECT * FROM test WHERE {values.json_path('data', 'c', 'nested')} = {values(obj['c']['nested'])}",
+        values,
+    )
+    assert row.data == obj
+
+    values = QueryValues()
+    row = await db.fetchone(
+        f"SELECT * FROM test WHERE {values.json_path('data', 'b')} != {values(obj['b'])}",
```
motorina0, line 39: - the test passes also for the query ` "SELECT * FROM test",`
- more rows to the `test` table are required in order to make this test meaningful

tests/core/test_db.py
```diff
@@ -0,0 +1,64 @@
...
+        f"""
+        SELECT * FROM test WHERE {values.json_path('data', 'a', type_=int)} >= {values(obj["a"])}
+        """,
+        values,
+    )
+    assert len(rows) == 1
+    assert rows[0].data == obj
+
+    values = QueryValues()
+    rows = await db.fetchall(
```
motorina0, line 35: - there is the risk that the parameters for `json_path` will be built dynamically
- there is no check for these values, below I have managed to insert a new row (SQL Injection)
- we should not rely on the fact that developers will use the API correctly (use a hardcoded value for the "column" param)
```py
    dataColumn = """data IS NOT NULL; INSERT INTO test VALUES ('{"inject": 3}'); SELECT * FROM test WHERE data"""
    rows = await db.fetchall(
        f"SELECT * FROM test WHERE {values.json_path(dataColumn, 'c', 'nested')} = {values(obj['c']['nested'])}",
        values,
    )
```

lnbits/db.py
```diff
@@ -47,27 +49,67 @@ def _parse_timestamp(value, _):
...
+    by appending each value when the placeholder inserted.
+
+    It is especially useful in cases like `json_path`,
+    which produces database-varying sql containing a placeholder with an according value.
+    Without this helper, the function would have to inject the value into the query string directly, which is a security risk.
+
+    For example usage, see tests/core/test_db.py
+    """
+
+    def json_path(self, colname: str, *path: str, type_: Type = None):
```
motorina0, line 79: please add return type for all functions

lnbits/db.py
```diff
@@ -47,27 +49,67 @@ def _parse_timestamp(value, _):
...
+        return accessor
+
+    def json_partial_update(self, colname: str, value: dict):
+        self.append(value)
+        if DB_TYPE == SQLITE:
+            return f"{colname} = json_patch({colname}, ?)"
+        else:
+            return f"{colname} = {colname}::jsonb || ?"
+
+    def placeholder(self, value: Any):
```
motorina0, line 99: - can this method be "private" (`_placeholder`)?
- is it ever called outside this class?

tests/core/test_db.py
```diff
@@ -0,0 +1,64 @@
...
+    assert rows[0].data == another_obj
+
+    update = {"b": "baz"}
+    obj |= update
+    values = QueryValues()
+    await db.execute(
+        f"UPDATE test SET {values.json_partial_update('data', update)}", values
+    )
+    row = await db.fetchone("SELECT * FROM test")
+    assert row.data == obj
```
motorina0, line 57: -now there are two rows in the table,  this test passes by chance. It might as well return the `another_obj` row

tests/core/test_db.py
```diff
@@ -0,0 +1,64 @@
...
+
+    values = QueryValues()
+    rows = await db.fetchall(
+        f"SELECT * FROM test WHERE {values.json_path('data', 'b')} != {values(obj['b'])}",
+        values,
+    )
+    assert len(rows) == 1
+    assert rows[0].data == another_obj
+
+    update = {"b": "baz"}
```
motorina0, line 50: - please also add test with nested level update

lnbits/db.py
```diff
@@ -47,27 +49,67 @@ def _parse_timestamp(value, _):
...
+        if DB_TYPE == SQLITE:
+            path_value = f"$.{'.'.join(path)}"
+            accessor = f"json_extract({colname}, ?)"
+        else:
+            # https://www.postgresql.org/docs/9.3/functions-json.html
+            path_value = "{" + ",".join(path) + "}"
+            accessor = f"{colname} #>> ?"
+            if type_ == int:
+                accessor = f"({accessor})::int"
+        self.append(path_value)
```
motorina0, line 89: - why is `self.append()` called?
- where are the values from the list used?

tests/core/test_db.py
jackstar12: The column name unfortunately cant be passed as a value, and usually the dynamic part would be the path and not the column

tests/core/test_db.py
jackstar12: Isn't that test somewhat useless now? it essentially tests that the data is being parsed correctly which the above do aswell

lnbits/db.py
jackstar12: It is not called, but I thought it can be leaved if anyone wanted to be specific about a placeholder being inserted. The `__call__` wraps it for convenience. But it could be private yes

tests/core/test_db.py
motorina0: > it essentially tests that the data is being parsed correctly which the above do aswell
- it tests that the data is actually updated
- if `row = await db.fetchone("SELECT * FROM test")` would return the row containing `another_obj` then the test would fail
- alternatively, if you switch the order of the inserts from
```py
    await db.execute("INSERT INTO test VALUES(?)", (obj,))
    await db.execute("INSERT INTO test VALUES(?)", (another_obj,))
```
  - to
```py
    await db.execute("INSERT INTO test VALUES(?)", (another_obj,))
    await db.execute("INSERT INTO test VALUES(?)", (obj,))
```
 - then the test will fail
 - this can lead to unpredictable test success/failure in the CI builds (sometimes tests just fail apparently for no reason)

tests/core/test_db.py
jackstar12: https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.quote_ident exists but it seems like a pg-only thing and also requires a connection

motorina0: - the column name could be sanitized as described in the first part of this: https://www.postgresql.org/docs/7.0/syntax525.htm 
> Names in SQL must begin with a letter (a-z) or underscore (_). Subsequent characters in a name can be letters, digits (0-9), or underscores. 

- I do not think it is necessary to cover the special cases where names are surrounded in double quotes (second part of the doc)

tests/core/test_db.py
jackstar12: sqlite and postgres behave slightly different in that regard. sqlite does a deep merge while postgres does a shallow one

motorina0: - what does this mean, can you please provide some examples?
- it is important that the persisted data is identical, people will start by using SQLite then move to PostgreSQL. There should be no difference in behavior when switching.

jackstar12: nested updates are inconsistent, so updating `{"a": {"b": "c"}}` with `{"a": {"d": "e"}}` results in `{"a": {"b": "c", "d": "e"}}` for sqlite but `{"a": {"d": "e"}}` for postgres.

jackstar12: there is no simple way to achieve the same in postgres, so the update function should rather look like `json_set(col, attr, value)`

tests/core/test_db.py
jackstar12: i added a somewhat primitive `_sanitize` function. if there is any non-alphabetic character, it rejects. Do you think this is sufficient?

motorina0:
dear @jackstar12 
 - I know you have put a lot of work in this PR and this approach looked like a good idea for me too when the PR was open, but at this moment I have to give a **No Go**
 - the functionality is useful, but it has to have a more broad use-case in order to be added to the `db.py`
 - the way to use the JSON querying is not generic enough to be easily understood by non-lnbits developers (too specific), there is a high risk to be misused (and cause vulnerabilities)
 - the DB logic has to be identical across all supported DBs (SQLite and Postgres in this case)

Not all PRs are meant to be merged.
I'm closing this PR now.

jackstar12:
> the functionality is useful, but it has to have a more broad use-case in order to be added to the `db.py`

Apart from the mentioned usermanager use case I would've been excited for convertning the `extra` field in apipayments, making it much more useful. Extensions could store some payment-related metadata and efficiently query by it later. For example, TPOS extension could display all transactions that were made through it.

> * the way to use the JSON querying is not generic enough to be easily understood by non-lnbits developers (too specific), there is a high risk to be misused (and cause vulnerabilities)
> 
> * the DB logic has to be identical across all supported DBs (SQLite and Postgres in this case)

Indeed, its in a bit of a  weird place, but standards are finally starting to pick it up, altough I think all inconsitencies are solved now. The cross-db really makes it a pain to work with raw sql.

I certainly respect your decision, but this leaves one more question to ask:
What is usermanager now supposed to do? Fetching all users and then filtering in python just seems ridicolous to me and without such functionality usermanager just uses a lot of value.

motorina0:
> What is usermanager now supposed to do? Fetching all users and then filtering in python just seems ridicolous to me and without such functionality usermanager just uses a lot of value.

- fields that require indexing or querying on should just be extracted as individual columns
- there is no such thing as "ridiculous code". The code must be easy to understand, modify and test. 
- the performance of the code is relative to the expected use-cases
  - if its a piece of logic that executes once an hour and loops over 1000 entries then it is probably fine
  - if its a piece of logic that executes every second and loops over 10.000 entries, then some performance improvements should be considered
     - the performance improvements can be to improve the SQL queries, but can also be caching, batching, etc
  - premature optimization is still the root of all evil: https://www.quora.com/What-do-engineers-mean-by-Premature-optimization-is-the-root-of-all-evil

motorina0:
> Should nested filters now be removed since they dont work on sqlite?
- if its not identical behaviour on SQLite and Postgres, then yes

jackstar12:
> fields that require indexing or querying on should just be extracted as individual columns

Obviously. But thats not the case with an `extra` column which is supposed to store additional user-defined data.
So just remove that feature? Or implement it in pure python?

motorina0:
> Obviously. But thats not the case with an `extra` column which is supposed to store additional user-defined data. So just remove that feature? Or implement it in pure python?

- if you have a data structure like this in the `extra` column:
```json
{
  "foo": {
    "bar": "a",
    "baz": "b"
  }
}
```
- and you need to query on `foo.baz` then you create a column `foo_baz` where the value is duplicated
- if the data structure is different across records, then you can just have a generic column(s)

But:
- **first** the performance problem must be clearly described and reproducible. Otherwise it cannot be measured how much a certain improvement actually words.
- **second** a performance improvement approach must be decided (it will not always be SQL query related)
- **third** implementation

jackstar12:
youve got to be kidding me

motorina0:
> @callebtc reopened this 5 days ago

- @callebtc please re-open with a comment
  - what is the reason to re-open
  - how the previous blockers will be solved

callebtc:
Forgot why I reopened it. Could've been mistake.

PR #1698, opened 2023-05-10
fix old payments endpoint for lndhub

dni:


jackstar12:
need to do:
```python
if not filters.sortby:
    filters.sortby = 'desc'
    

if not filters.direction:
    filters.direction = 'desc'
```

dni:
can you take another, look i think putting it into ```if not filters``` is basically the same right?

callebtc:
Tested, LGTM

callebtc (approved):

PR #1700, opened 2023-05-10
Refactor language selection

talvasconcelos:
Refactored language selection with a loop.

Languages are defined in base.html as an array.

Possibly we should add an ISO code, for languages ex: `en_US, pt_PT, pt_BR` and an optional display also `display: 'EN'` , `display: 'PT'`, `display: PT(BR)`


dni (approved):

motorina0:
@dni do I remember correctly that there was (related) a PR that added flags next to the language?
Will that be in conflict with this one?

lnbits/templates/base.html
```diff
@@ -325,9 +285,22 @@
...
+      window.langs = [
+        { value: 'en', label: 'English', display: 'EN' },
+        { value: 'de', label: 'Deutsch', display: 'DE' },
+        { value: 'es', label: 'Español', display: 'ES' },
+        { value: 'jp', label: '日本語', display: 'JP' },
+        { value: 'fr', label: 'Français', display: 'FR' },
+        { value: 'it', label: 'Italiano', display: 'IT' },
+        { value: 'pi', label: 'Pirate', display: 'PI' },
+        { value: 'nl', label: 'Nederlands', display: 'NL' },
+        { value: 'we', label: 'Welsh' },
```
motorina0, line 301: Why no `display` field for `Polish` and `Pirate`

motorina0 (approved):

lnbits/templates/base.html
callebtc: `Welsh` isn't a `Pirate` language afaik, but I could be wrong ;)

talvasconcelos: @motorina0 , it's an optional field, i don't know, and copilot didn't step in, how to spell `Polish` in ... polish!

@callebtc ??

arbadacarbaYK: Polski

arcbtc: Welsh should be CY, yep, please change

PR #1702, opened 2023-05-11
WIP: Node management

jackstar12:
Introduces a new Node Management UI, still in progress.
Only cln is implemented at the moment


jackstar12:
current state of dashboard:
image

channels:
image

transactions are showing the payments that happened directly on the node:
image

arcbtc:
Amazing. Can you remove "apps", thats for another pr

arcbtc:
I love how it doesn't look boilerplate at all, just another lnbits menu item (for those who choose to enable). Very cool

jackstar12:
been working on the channels tab:
image

jackstar12:
I think I should clarify why supporting JSON columns in db would be beneficial.

I'll use the recently added `extra` column in the https://github.com/lnbits/usermanager (which currently doesnt work on sqlite!) as an example.

It is important that the usermanager api gives a way of filtering by custom data in the extra column.
There are essentially two ways to do this:
- fetch all users and filter application side
- filter directly in sql and only return required rows

The latter option is obviously much more performant, but it requires the column type to be json so that we can write queries accordingly. 
It also allows for easier partial updates. With a text column you'd have to select the row, do the update in python and then update. JSON type makes this possible in sql directly.

But it doesnt end here:
If the extra column on payments was changed to JSON type aswell, the api could support filtering by extra data too. So you could easily get all payments with the tag `boltz`. Something like that isnt possible right now.
If the extra column becomes more viable more extensions could use it to put some metadata alongside the payments, e.g. the discordbot could store the channel id the payment happened in so it can display stats.

jackstar12:
node ui can now be turned on or off in the admin ui:
image

You can optionally expose some info about your node on a public path (/node/public) aswell:
image

jackstar12:
The changes regarding the `services.py` fix type errors which now arise because I've typed the get_wallet_class function correctly

arcbtc:
Getting 
image

jackstar12:
When writing the node tests I also figured out that the wallet class isn't initialized correctly, so `is_fake` was always true. This is now fixed and the `test_pay_real_invoice` test now actually runs (never ran before) and seems to cause some problems.

kirito41dd:
very good job, i very need this feature! when can i try it？

jackstar12:
> very good job, i very need this feature! when can i try it？
if you _really_ want to you can checkout to this branch. Will take a bit more until its fully finished, Ive been extracting some of the changes here into other smaller prs to maoe this one less bulky.

callebtc:
Resolved merge conflicts to new main

arcbtc:
LFGTM!
Its the best node management software on the planet, its not fair the world doesn't have it yet

lnbits/app.py
```diff
@@ -328,6 +328,8 @@ async def lnbits_startup():
             # check extensions after restart
             await check_installed_extensions(app)
 
+            asyncio.create_task(cache.invalidate_forever())
```
motorina0, line 331: - shouldn't this be part on `main` (not part of node management functionality)?

lnbits/core/views/generic.py
```diff
@@ -387,6 +387,44 @@ async def manifest(usr: str):
     }
 
 
+@core_html_routes.get("/node", response_class=HTMLResponse)
+async def node(request: Request, user: User = Depends(check_admin)):
+    if not settings.lnbits_node_ui:
+        raise HTTPException(status_code=HTTPStatus.NOT_FOUND)
```
motorina0, line 393: - maybe `HTTP 503 Service Unavailable` is more appropriate
- `404 Not Found` is if the route/resource does not exist

lnbits/core/views/node_api.py
```diff
@@ -0,0 +1,189 @@
...
+)
+@public_node_api.get(
+    "/rank",
+    description="Retrieve node ranks from https://1ml.com",
+    response_model=Optional[NodeRank],
+)
+async def api_get_1ml_stats(node: Node = Depends(require_node)) -> Optional[NodeRank]:
+    node_id = await node.get_id()
+    async with httpx.AsyncClient() as client:
+        # node_id = "026165850492521f4ac8abd9bd8088123446d126f648ca35e60f88177dc149ceb2"
```
motorina0, line 179: - remove dead code

lnbits/static/js/components.js
```diff
@@ -178,6 +178,7 @@ Vue.component('lnbits-extension-list', {
 })
 
 Vue.component('lnbits-admin-ui', {
+  props: ['node'],
```
motorina0, line 181: - nitpick: this is a `bool`, maybe `has-node` is a more suggestive name

lnbits/nodes/base.py
```diff
@@ -0,0 +1,239 @@
...
+from pydantic import BaseModel
+
+from lnbits.cache import cache
+from lnbits.db import FilterModel, Filters, Page
+
+if TYPE_CHECKING:
+    from lnbits.wallets.base import Wallet
+
+
+class NodeStatsResponse(NamedTuple):
```
motorina0, line 16: - not used anywhere

lnbits/nodes/base.py
```diff
@@ -0,0 +1,239 @@
...
+    amount: int
+    memo: Optional[str]
+    bolt11: str
+    preimage: Optional[str]
+    payment_hash: str
+    paid_at: Optional[int] = None
+    expiry: Optional[int] = None
+
+
+class NodePaymentsResponse(BaseModel):
```
motorina0, line 137: - not used anywhere

lnbits/nodes/base.py
```diff
@@ -0,0 +1,239 @@
...
+    @abstractmethod
+    async def _get_peer_info(self, pubkey: str) -> NodePeerInfo:
+        pass
+
+    async def get_peer_info(self, pubkey: str) -> NodePeerInfo:
+        info = cache.get(f"peers:{pubkey}")
+        if not info:
+            info = await self._get_peer_info(pubkey)
+            if info.last_timestamp:
+                cache.set(f"peers:{pubkey}", info)
```
motorina0, line 195: - the cache key might be too generic resulting in conflicts.
- something like `node,peers:{pubkey}` would be safer

lnbits/nodes/cln.py
```diff
@@ -0,0 +1,322 @@
...
+
+    @catch_rpc_errors
+    async def connect_peer(self, uri: str):
+        # https://docs.corelightning.org/reference/lightning-connect
+        try:
+            await self.ln_rpc("connect", uri)
+        except RpcError as e:
+            message = e.error["message"]
+            code = e.error["code"]
+            print("HII")
```
motorina0, line 73: - Hello!

lnbits/wallets/base.py
```diff
@@ -1,5 +1,10 @@
+from __future__ import annotations
+
 from abc import ABC, abstractmethod
-from typing import AsyncGenerator, Coroutine, NamedTuple, Optional
+from typing import TYPE_CHECKING, AsyncGenerator, Coroutine, NamedTuple, Optional, Type
+
+if TYPE_CHECKING:
```
motorina0, line 6: - why this condition, is it to avoid circular dependencies?

lnbits/wallets/lndrest.py
```diff
@@ -74,6 +77,8 @@ async def cleanup(self):
         except RuntimeError as e:
             logger.warning(f"Error closing wallet connection: {e}")
 
+        self.client = httpx.AsyncClient(verify=self.cert, headers=self.auth)
```
motorina0, line 80: - why add this?

lnbits/core/views/node_api.py
```diff
@@ -0,0 +1,189 @@
...
+            status_code=404,
+            detail="Not enabled",
+        )
+    return NODE
+
+
+def check_public():
+    if not (settings.lnbits_node_ui and settings.lnbits_public_node_ui):
+        raise HTTPException(
+            status_code=404,
```
motorina0, line 46: - HTTP 503?

lnbits/core/views/node_api.py
```diff
@@ -0,0 +1,189 @@
...
+    status_code=200,
+    dependencies=[Depends(require_node)],
+)
+async def api_get_ok():
+    pass
+
+
+@public_node_api.get("/info", response_model=PublicNodeInfo)
+async def api_get_public_info(node: Node = Depends(require_node)) -> PublicNodeInfo:
+    return await node.get_public_info()
```
motorina0, line 69: - this goes directly to the LN node, can't it be abused (since it is public)?

motorina0: - same for all `node: Node = Depends(require_node)`

lnbits/settings.py
```diff
@@ -218,6 +220,12 @@ class FundingSourcesSettings(
     lnbits_backend_wallet_class: str = Field(default="VoidWallet")
 
 
+class NodeUISettings(LNbitsSettings):
```
motorina0, line 223: please add some comments to these new fields

lnbits/core/views/node_api.py
```diff
@@ -0,0 +1,189 @@
...
+@node_api.get("/channels")
+async def api_get_channels(
+    node: Node = Depends(require_node),
+) -> Optional[List[NodeChannel]]:
+    return await node.get_channels()
+
+
+@node_api.post("/channels", response_model=ChannelPoint)
+async def api_create_channel(
+    node: Node = Depends(require_node),
```
motorina0, line 88: - this is a `POST` with no user permission required (anyone can call it if a certain flag is true)?

lnbits/core/views/node_api.py
```diff
@@ -0,0 +1,189 @@
...
+    return await node.open_channel(peer_id, funding_amount, push_amount, fee_rate)
+
+
+@node_api.delete("/channels")
+async def api_delete_channel(
+    short_id: Optional[str],
+    funding_txid: Optional[str],
+    output_index: Optional[int],
+    force: bool = False,
+    node: Node = Depends(require_node),
```
motorina0, line 103: - this is a `DELETE` with no user permission required (anyone can call it if a certain flag is true)?

lnbits/core/views/node_api.py
```diff
@@ -0,0 +1,189 @@
...
+        ChannelPoint(funding_txid=funding_txid, output_index=output_index)
+        if funding_txid is not None and output_index is not None
+        else None,
+        force,
+    )
+
+
+@node_api.get("/payments", response_model=Page[NodePayment])
+async def api_get_payments(
+    node: Node = Depends(require_node),
```
motorina0, line 116: - is it ok for this info to be accessed by anyone?

lnbits/core/views/node_api.py
jackstar12: The entire `node_api` router is protected by the `check_admin` dependency. It is declared above.

lnbits/core/views/node_api.py
```diff
@@ -0,0 +1,189 @@
...
+        raise HTTPException(
+            HTTP_503_SERVICE_UNAVAILABLE,
+            detail="You can enable node transactions in the Admin UI",
+        )
+    return await node.get_payments(filters)
+
+
+@node_api.get("/invoices", response_model=Page[NodeInvoice])
+async def api_get_invoices(
+    node: Node = Depends(require_node),
```
motorina0, line 129: - is it ok for this info to be accessed by anyone?

lnbits/core/views/node_api.py
motorina0: ah, ok!

lnbits/core/views/node_api.py
jackstar12: Its probably a good idea to cache it

lnbits/wallets/base.py
jackstar12: yes

lnbits/core/views/node_api.py
motorina0: - even so, for some operations (like `create_channel`) it should be reserved for `super_admin`
- the funding source selection & config are reserved for `super_admin`

jackstar12: if normal admins cant create / close channels, why bother showing them? I'd say either all or nothing

motorina0: > I'd say either all or nothing
- why?

- normal admins can be non-technical users with special privileges to manage resources
- normal admins already are forbidden from configuring the funding source

motorina0 (approved):
LGTM, see comments.

Especially: 
- only `super_admin` manages node operations
- others must watch-only (including regular admins)

motorina0:
@dni @jackstar12 do you know why this PR is closed?

PR #1727, opened 2023-05-23
fix "boolean" query for sqlite

talvasconcelos:
`get_latest_payments_by_extension` isn't working on SQlite. The query `WHERE pending = 'false'` doesn't work for SQLIte, changed it to `0`!

Issue was that TPOS wouldn't get the latest payments to display!


callebtc (approved):

motorina0 (approved):
LGTM!

PR #1740, opened 2023-05-25
quickfix spacing und mobile footer

dni:
screenshot-1686817719
screenshot-1686817738

the footer nav is show until md, but the space only for mobile, should be on both.


motorina0:
- PR now as a conflict
- please add before/after screenshot

motorina0 (approved):

talvasconcelos (approved):

Issue #1709, opened 2023-05-16
[Feature request] Option to nuke the db of an uninstalled extension


callebtc:
It's nice that we keep the db of an uninstall extension by default but it would be great to have the option to nuke the whole schema in case something goes wrong with an install.
<img width="584" alt="image" src="https://github.com/lnbits/lnbits/assets/93376500/e37c6ee1-3554-4037-bc58-a47a2c9e3af8">

motorina0:
Handled here: https://github.com/lnbits/lnbits/pull/1746

PR #1746, opened 2023-05-30
Add option to drop extension db at un-install time or later

motorina0:
Requested here: https://github.com/lnbits/lnbits/issues/1709

### Summary
- by default when an extension is un-installed, all the data is still stored on the LNbits DB
- sometimes admin users might want to also remove all the data. Some reasons:
  - there was a corrupt migration file and now the extension cannot be upgraded/reinstalled
  - the existing data was created for demo/testing purposes and should be dropped
  - ...

This PR adds a `Remove Database` option at un-install time, or later.
### Tests Steps
 - see flow in screen recording below
 - use both Postgres and SQLite

### Uninstall and Drop DB
https://github.com/lnbits/lnbits/assets/2951406/8cc29d63-5698-4fba-90ee-152b5a9404a6

### Drop DB after Uninstall

https://github.com/lnbits/lnbits/assets/2951406/8d6d279d-842e-430d-abb3-de8a97b5539f


lnbits/core/static/js/service-worker.js-e
```diff
@@ -0,0 +1,58 @@
+// update cache version every time there is a new deployment
```
dni, line 1: lnbits/core/static/js/service-worker.js-e ? is this a error?

lnbits/core/static/js/extensions.js
```diff
@@ -1,43 +0,0 @@
-new Vue({
```
dni, line 1: was this file unnecessary?

motorina0: It was a leftover before the merge of  `Extensions` (`/extensions`) and `Manage Extensions` (`/install`)

lnbits/core/static/js/service-worker.js-e
motorina0: - `make bundle` generated it for me
- I have removed it

dni (approved):

talvasconcelos (approved):

Issue #1682, opened 2023-05-05
[Feature request] Hide the superuser UI URL from logs that are visible to all Unix system users


Perlover:
### Summary

When I start LNBits in logs I see by command `journalctl -u lnbits.service`:

```
SUCCESS | ✔️ Access super user account at: http://127.0.0.1:5000/wallet?usr=XXXXXXXXXXXXXXXX
```

But! Logs are viewed by any user in system by command:

`journalctl -u lnbits.service`

I think it's very bad. Any user in Unix can see logs, to see admin account and after to send all funds from lightning node to other wallet or to see macaroon admin files and etc.

It's easy to protect files with file permissions (for example `.env` file with `chmod 600 ...`), but we can't protect a view of logs of systemd. It is very dangerous to write authorization data there.

### Impact

Any instances with admin account when made as database setup

From doc:

> The LNbits Admin UI lets you change LNbits settings via the LNbits frontend. It is disabled by default and the first time you set the enviroment variable LNBITS_ADMIN_UI=true in the .env the settings are initialized and saved to the database and will be used (as long the UI is enabled). From there on the according settings from the database instead those of the .env file are used. If you want to return to those settings you can just set LNBITS_ADMIN_UI=false and restart LNbits to take the .env again.

> **On restart of LNbits you will find the link to the superuser-AdminUI in your terminal resp**. in your logs. Save if good and never share it. When the AdminUI is opened we can move on.

Perlover:
I would like to suggest a way to solve this problem. Instead of writing to the log, you can open a file for creation with 600 rights from the lnbits user and write the admin URL there.  In advance, in advance, add the name of such a file to `.gitignore`

dni:
we recently added this... https://github.com/lnbits/lnbits/pull/1651

Perlover:
I want to make a clarification. As the user Hamish suggested to me in the Telegram chat, the logs are not always visible to everyone. I created a test user and really, I can't view the log from under it

I think I see logs either because my user is in the adm group, and I see logs from under `lnbits` user because the lnbits process is started from the lnbits user himself.

From `man journalctl`:

> All users are granted access to their private per-user journals. However, by default, only root and users who are members of a few **special groups are granted access to the system journal and the journals of other users**. Members of the groups "systemd-journal", "**adm**", and "**wheel**" can read all journal
> files. Note that the two latter groups traditionally have additional privileges specified by the distribution. Members of the "wheel" group can often perform administrative tasks.

[...]

PR #1749, opened 2023-05-31
Hide the superuser ID URL from logs with comment where to find superuser ID

Perlover:
I have not tested the code, but I suggest such logging than logging the administrator ID. In addition, the administrator ID is already saved to a specific file.

It fixes this:

https://github.com/lnbits/lnbits/issues/1682

### DIFF
--- a/docs/guide/admin_ui.md
+++ b/docs/guide/admin_ui.md
@@ -55,11 +55,12 @@ Now start LNbits once in the terminal window
 ``` 
 $ poetry run lnbits 
 ```
-It will now show you the Super User Account:
-
-`SUCCESS | ✔️ Access super user account at: https://127.0.0.1:5000/wallet?usr=<ID from ".super_user" file>`
-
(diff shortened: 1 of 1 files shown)


motorina0:
Are the rest of the `settings` still logged?
The `settings` object contains DB conn string, admin users, etc. It should be excluded also.

Perlover:
> Are the rest of the `settings` still logged? The `settings` object contains DB conn string, admin users, etc. It should be excluded also.

There is no logging of the complete object. Only `settings.lnbits_baseurl` is logged there, but this is the base URL, for example, to the address `127.0.0.1` and there is no private information there, except for the path. But it is standard for all installations.

Example of logging after patch is here: 

https://github.com/lnbits/lnbits/issues/1682#issuecomment-1570695004

Perlover:
> Your organization is not using the GitHub App Integration

Good <s>boy</s> bot :)

motorina0 (approved):

Issue #1754, opened 2023-06-10
[BUG] TX chart/export


arbadacarbaYK:
**Describe the bug**
- Chart shows negative balance value
- "Export to csv" field "memo" shows date instead message

Bad to reproduce sales. 
For a proper tax report for merchants on top a column with a denominated value at the time of action would be needed here.
Since more merchants just act through LNbits (SaaS/Cloud Setups) it has gotten even more important to deliver a valid report.

**To Reproduce**
Press "Export to csv" or the chart-symbol

**Expected behavior**

Chart
- no negative balance :P  
-Balance should overlap with the blocks

Export
- Shows complete message in "Memo"
- Date in a separate column (than memo to be able to order by messages too)
- only successful transactions are listed 

**Screenshots**

<img width="927" alt="Bildschirmfoto 2023-06-10 um 23 23 39" src="https://github.com/lnbits/lnbits/assets/63317640/863125ba-e2a5-46f0-9e54-8e833c6330a3">

<img width="887" alt="Bildschirmfoto 2023-06-10 um 23 40 19" src="https://github.com/lnbits/lnbits/assets/63317640/63fab413-e88c-4d52-82a7-7c24d3fdda55">

<img width="652" alt="Bildschirmfoto 2023-06-10 um 23 36 44" src="https://github.com/lnbits/lnbits/assets/63317640/5bec86fc-5a6c-418f-8b15-bc1c6b002eca">

<img width="755" alt="Bildschirmfoto 2023-06-10 um 23 31 18" src="https://github.com/lnbits/lnbits/assets/63317640/64115f23-869b-4c15-bd55-1ea9edcd528c">

jackstar12:
This has probably because pagination was moved to the server to increase page performance. The exportCSV and Chart haven't been adjusted to that.

jackstar12:
@talvasconcelos no need, already got it

jackstar12:
saving the corresponding might be a cool addition aswell. Could build on https://github.com/lnbits/lnbits/pull/1748

arbadacarbaYK:
Its a need, not a nice to have actually :P

jackstar12:
@arbadacarbaYK how would you imagine that feature? Maybe as a per-wallet setting where you can set your preferred fiat whose value is then stored per payment?

arbadacarbaYK:
You might not like the answer 😂
I think it makes sense to have the possibility to override in some extensions so e.g.

- per lnurlp
[...]

PR #1756, opened 2023-06-12
fix csv-export and payments chart

jackstar12:
Related: https://github.com/lnbits/lnbits/issues/1754

Adds an endpoint for retrieving payments in a csv format.
The provided colums are the same as previously exported in the frontend.

The `StreamingResponse` is used to avoid having to save the entire file locally, which results in better performance.

I also extracted the common logic of updating pending payments into its own function.
`check_payment_status` was replaced with `payment.check_status` because the first one calls the latter anyways while doing an additional query.


jackstar12:
added the new payments chart here aswell. wanted to do a seperate pr but the new `/payments/history` endpoint uses the new function for updating pending payments aswell. This endpoint essentially returns the data as expected by the chart and makes it possible to reuse it in the future (maybe for a graph of the total node activity)

jackstar12:
Note that I now removed the code I was already suspicious of in https://github.com/lnbits/lnbits/pull/1697 because it was now causing bugs when converting postgres dates
https://github.com/lnbits/lnbits/blob/95071eb87e94481a0b2e5bcd29517eaa1f8c8dee/lnbits/db.py#L50-L58

lnbits/db.py
```diff
@@ -126,6 +120,24 @@ def timestamp_placeholder(cls):
         else:
             return "?"
 
+    @classmethod
+    def truncate_date(cls, col: str, trunc: DateTrunc) -> str:
```
motorina0, line 124: - value format/truncate should take place at the Business Logic layer (python) and not at the Persistence layer (SQL)

lnbits/core/crud.py
```diff
@@ -619,6 +627,64 @@ async def update_payment_extra(
...
+            balance = wallet.balance_msat
+        else:
+            raise ValueError("Unknown wallet")
+    else:
+        balance = await get_total_balance()
+
+    # since we dont know the balance at the starting point, we take the current balance and walk backwards
+    results: list[PaymentHistoryPoint] = []
+    for row in transactions:
+        print(row)
```
motorina0, line 677: - remove `print()`

lnbits/db.py
```diff
@@ -126,6 +120,24 @@ def timestamp_placeholder(cls):
...
+            formats = {
+                "hour": "%Y-%m-%d %H:00:00",
+                "day": "%Y-%m-%d 00:00:00",
+                "month": "%Y-%m-01 00:00:00",
+            }
+            if trunc in formats:
+                return f"strftime('{formats[trunc]}', {col}, 'unixepoch')"
+        elif trunc in ("day", "hour", "month"):
+            return f"date_trunc('{trunc}', {col})"
+        raise ValueError(f"Invalid truncation: {trunc}")
```
callebtc, line 139: What exactly does this do? In sqlite, we used to store unix timestamps whereas in postgres, we used the `date` type. Does this affect anything? Would be nice to clarify.

lnbits/core/crud.py
```diff
@@ -619,6 +627,63 @@ async def update_payment_extra(
...
+        filters = Filters()
+    where = ["(pending = False OR amount < 0)"]
+    values = []
+    if wallet_id:
+        where.append("wallet = ?")
+        values.append(wallet_id)
+
+    transactions = await db.fetchall(
+        f"""
+        SELECT {db.truncate_date('time', group)} date,
```
callebtc, line 655: An equivalent of `db.truncate_date` should be applied in the frontend (chart) that needs the truncated time stamps, not in the db level.

lnbits/db.py
callebtc: Agreed. There are three options to truncate:

- in the db
- in the endpoint
- in the frontend

only the fronend needs it (for this specific use case of charting the payments), so it should be handled there.

callebtc:
There are multiple undocumented logic changes in this PR, like `update_pending_payments` and returned `get_payments` calls. Please clarify each change so we can properly review it! Thanks 🙏

lnbits/db.py
jackstar12: No, it just affects the way those timestamps are being used in the query.

jackstar12:
> There are multiple undocumented logic changes in this PR, like `update_pending_payments` and returned `get_payments` calls. Please clarify each change so we can properly review it! Thanks 🙏

From above (regarding `update_pending_payments`):
I also extracted the common logic of updating pending payments into its own function.
check_payment_status was replaced with payment.check_status because the first one calls the latter anyways while doing an additional query.

Which `get_payments` calls do you mean?

talvasconcelos:
Fixes #1754

jackstar12:
@callebtc @motorina0 
what about a middleground where the csv-export just fetches all payments into the frontend and creates the csv there? Altough less efficient it would make for less added code. Except retrieving csv through api is desired.

jackstar12:
I think the balance history should be on the server tough for above stated reasons (and for creating a admin balance chart of the entire instance)

PR #1757, opened 2023-06-13
Changes asdict to dict for payment websocket

arcbtc:
asdict worked before, now it doesn't so I switched to dict.


talvasconcelos (approved):
Works fine for me!

dni (approved):
i think this actually broke the webhook

PR #1761, opened 2023-06-14
TEST: fix tests for real wallets

dni:
as @jackstar12 discover the regtest were not using the actual fundingsources.

- switched invoice create to lnd-1 node because it is connected the best inside regtest
- make invoice amount smaller
- switch LndWallet, LndRestWallet to use lnd-3 (selfpayments)

had to change the docker regtest setup aswell, because eclair was actually not working. using a wrong zmq endpoint

also @jackstar12 fixed eclair backend, it was also not beeing able to pay a real invoice.

### DIFF
--- a/lnbits/wallets/eclair.py
+++ b/lnbits/wallets/eclair.py
@@ -3,7 +3,7 @@ import base64
 import hashlib
 import json
 import urllib.parse
-from typing import AsyncGenerator, Dict, Optional
+from typing import Any, AsyncGenerator, Dict, Optional
 
 import httpx
 from loguru import logger
@@ -70,7 +70,7 @@ class EclairWallet(Wallet):
         **kwargs,
     ) -> InvoiceResponse:
 
-        data: Dict = {
+        data: Dict[str, Any] = {
             "amountMsat": amount * 1000,
         }
         if kwargs.get("expiry"):
--- a/lnbits/wallets/eclair.py
+++ b/lnbits/wallets/eclair.py
@@ -72,8 +72,6 @@ class EclairWallet(Wallet):
 
         data: Dict = {
             "amountMsat": amount * 1000,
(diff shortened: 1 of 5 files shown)


jackstar12:
Got a fix for the 2 eclair tests:
```patch
diff --git a/lnbits/wallets/eclair.py b/lnbits/wallets/eclair.py
index 97d39064..7eea5a37 100644
--- a/lnbits/wallets/eclair.py
+++ b/lnbits/wallets/eclair.py
@@ -71,8 +71,6 @@ class EclairWallet(Wallet):
     ) -> InvoiceResponse:
         data: Dict = {
             "amountMsat": amount * 1000,
-            "description_hash": b"",
-            "description": memo,
         }
         if kwargs.get("expiry"):
             data["expireIn"] = kwargs["expiry"]
@@ -81,6 +79,8 @@ class EclairWallet(Wallet):
             data["descriptionHash"] = description_hash.hex()
         elif unhashed_description:
             data["descriptionHash"] = hashlib.sha256(unhashed_description).hexdigest()
+        else:
+            data["description"] = memo
 
         async with httpx.AsyncClient() as client:
             r = await client.post(
```

lnbits/wallets/eclair.py
```diff
@@ -70,10 +70,8 @@ async def create_invoice(
         **kwargs,
     ) -> InvoiceResponse:
 
-        data: Dict = {
+        data: Dict[str, Any] = {
             "amountMsat": amount * 1000,
-            "description_hash": b"",
-            "description": memo,
```
motorina0, line 76: - previously the `description` was always sent
- now `description` is sent only if both `description_hash` and `unhashed_description` are missing

motorina0 (approved):
LGTM

talvasconcelos (approved):

PR #1762, opened 2023-06-15
BUGFIX: Docker image wont build on debian12, use 11 instead

dni:
- use debian 11 instead of 12

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-slim
+FROM python:3.10-slim-bullseye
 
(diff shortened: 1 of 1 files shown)


motorina0:
- what is different between `python:3.10-slim` and `python:3.10-slim-bullseye`?
- what is the problem with `python:3.10-slim`, can it affect users running in no-docker environment?

motorina0 (approved):
LGTM

dni:
included here https://github.com/lnbits/lnbits/pull/1761

PR #1770, opened 2023-06-19
[Wallets] CLN: fix pending state check

callebtc:


lnbits/wallets/cln.py
```diff
@@ -161,15 +161,18 @@ async def get_invoice_status(self, checking_id: str) -> PaymentStatus:
...
         return PaymentStatus(None)
 
     async def get_payment_status(self, checking_id: str) -> PaymentStatus:
         try:
-            r = self.ln.call("listpays", {"payment_hash": checking_id})
+            r = self.ln.listpays(payment_hash=checking_id)
         except:
             return PaymentStatus(None)
-        if not r["pays"]:
+        if "pays" not in r or not r["pays"]:
```
dni, line 175:
```suggestion
        if "pays" not in r and not r["pays"]:
```

callebtc: Why and? It should fail either way...?

dni: ok, sry my mistake!

tests/conftest.py
```diff
@@ -125,7 +125,7 @@ async def invoice(to_wallet):
     del invoice
 
 
-@pytest_asyncio.fixture(scope="session")
+@pytest_asyncio.fixture(scope="function")
```
dni, line 128: what is function?

dni (approved):

PR #1729, opened 2023-05-24
Wallets refactor

jackstar12:
While working on the node implementation I noticed that all wallet implementations which work over HTTP have a pretty obvious inefficiency: they create a new `httpx.AsyncClient` for every request. This causes a new connection to be made everytime which adds unnecessary overhead (especially with https). 

Instead, the client should be reused. I also added a `cleanup` method to allow proper cleanup on shutdown.

PR #1775, opened 2023-06-20
Wallets: add cln-rest

callebtc:
This PR adds [cln-rest](https://github.com/Ride-The-Lightning/c-lightning-REST) as a backend.

Several quircks of the API made it necessary to change a lot of things:
- `get_payment_status` and `get_invoice_status` now need a `Payment` object as an argument and not the `checking_id`.  The reason for this is that cln-rest wants the label and the bolt11 invoice for both of these functions, instead of the same identifier. 
- In order to import `Payment`, I had to change `settings` which imported the `core.wallet` module and caused a circular import. Now the `core.wallet` module in `settings` is only imported when actually needed and not done in the global scope (Update: Fixed in #1776)

## Bug fixes:
- [fix checking_id return in api_payments_create_invoice](https://github.com/lnbits/lnbits/pull/1775/commits/4af4c702159a4388f0ceb1b054213ef61a390d0c)

Todo:
- [x] status
- [x] create invoice
- [x] pay invoice
- [x] invoice status
- [x] payment status
- [x] invoice stream
- [x] check whether I can put `PaymentStatus` back where it came from (done)
- [x] enable regtest for cln-rest


lnbits/settings.py
```diff
@@ -343,15 +350,16 @@ def set_cli_settings(**kwargs):
...
 
 
-def get_wallet_class():
-    # wallet_class = getattr(wallets_module, settings.lnbits_backend_wallet_class)
-    return WALLET
+def get_wallet_class(wallet: Optional[str] = None):
+    wallets_module = importlib.import_module("lnbits.wallets")
+    if wallet:
+        return getattr(wallets_module, wallet)()
+    return getattr(wallets_module, settings.lnbits_backend_wallet_class)()
```
callebtc, line 362: Reviewer: Please check whether this affects any of your code.

lnbits/settings.py
```diff
@@ -343,15 +350,16 @@ def set_cli_settings(**kwargs):
 
 # set wallet class after settings are loaded
 def set_wallet_class(class_name: Optional[str] = None):
-    backend_wallet_class = class_name or settings.lnbits_backend_wallet_class
-    wallet_class = getattr(wallets_module, backend_wallet_class)
-    global WALLET
-    WALLET = wallet_class()
+    settings.lnbits_backend_wallet_class = (
+        class_name or settings.lnbits_backend_wallet_class
+    )
```
callebtc, line 355: Reviewer: Please check whether this affects any of your code.

jackstar12:
I've also been facing some issues with circular imports over at https://github.com/lnbits/lnbits/pull/1702. I extracted my approach into a seperate PR, I think those changes would address your issues aswell: https://github.com/lnbits/lnbits/pull/1776

tests/core/views/test_api.py
```diff
@@ -346,7 +346,23 @@ async def test_pay_real_invoice(
...
+            amount=0,
+            time=0,
+            bolt11="",
+            memo="",
+            expiry=0,
+            preimage="",
+            wallet_id="",
+            webhook=None,
+            webhook_status=None,
+        )
```
jackstar12, line 364: Those dummies are pretty common now, a little helper would probably keep everything a bit cleaner.

Something like
`Payment.dummy(invoice["payment_hash"])`

lnbits/wallets/clnrest.py
```diff
@@ -0,0 +1,214 @@
...
+            "macaroon": macaroon,
+            "accept": "application/json",
+        }
+        if macaroon_ishex:
+            self.auth["encodingtype"] = "hex"
+
+        self.cert = settings.cln_rest_cert or False
+
+    async def status(self) -> StatusResponse:
+        async with httpx.AsyncClient(verify=self.cert) as client:
```
jackstar12, line 40: The httpx client should be reused just like the other wallets now do

lnbits/settings.py
jackstar12: These changes would make https://github.com/lnbits/lnbits/pull/1729 redundant as the wallet class would be re-instantiated on every call

lnbits/wallets/clnrest.py
callebtc: will do that!

lnbits/wallets/cln.py
```diff
@@ -146,34 +145,34 @@ async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse
...
+        logger.warning(f"supplied an invalid checking_id: {payment.checking_id}")
         return PaymentStatus(None)
 
-    async def get_payment_status(self, checking_id: str) -> PaymentStatus:
+    async def get_payment_status(self, payment: Payment) -> PaymentStatus:
         try:
-            r = self.ln.call("listpays", {"payment_hash": checking_id})
+            r: dict = self.ln.call("listpays", {"payment_hash": checking_id})  # type: ignore
         except:
             return PaymentStatus(None)
```
jackstar12, line 170: the `checking_id` variable here doesnt exist which causes the tests to fail. (needs to be `payment.checking_id`
doesnt get caught because of the `# type: ignore` though

the broad `except` doesnt really help either as it swallows the error

lnbits/wallets/cln.py
```diff
@@ -146,34 +145,34 @@ async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse
             True, r["payment_hash"], fee_msat, r["payment_preimage"], None
         )
 
-    async def get_invoice_status(self, checking_id: str) -> PaymentStatus:
+    async def get_invoice_status(self, payment: Payment) -> PaymentStatus:
         try:
-            r = self.ln.listinvoices(payment_hash=checking_id)
+            r: dict = self.ln.listinvoices(payment_hash=checking_id)  # type: ignore
         except:
             return PaymentStatus(None)
```
jackstar12, line 152: same as with `get_payment_status`

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+
+            if r.is_error or "error" in data or data.get("invoices") is None:
+                raise Exception("error in cln response")
+            return PaymentStatus(self.statuses.get(data["invoices"][0]["status"]))
+        except Exception as e:
+            logger.error(f"Error getting invoice status: {e}")
+            return PaymentStatus(None)
+
+    async def get_payment_status(self, payment: Payment) -> PaymentStatus:
+        # corelightning-rest wants the "bolt11" here.... sigh
```
jackstar12, line 177: what if, instead of changing up the whole wallets interface we simply get the bolt11 from the database using the checking id? This wouldnt work for payments we did not initate, but thats never done anyways. Not ideal, but I guess its fine until cln-rest supports it
```suggestion
    async def get_invoice_status(self, checking_id: str) -> PaymentStatus:
        payment = await get_standalone_payment(checking_id)
        if not payment:
            raise ValueError("Unknown checking_id")
```

jackstar12 (changes requested):

tests/core/views/test_api.py
```diff
@@ -367,12 +371,18 @@ async def test_pay_real_invoice(
...
+    payment_status = response.json()
+    assert payment_status["paid"]
 
-    status = await WALLET.get_payment_status(invoice["payment_hash"])
+    # create a dummy payment object of which we will
+    # only use checking_id in get_payment_status
+    status = await WALLET.get_payment_status(
+        Payment.dummy(checking_id=invoice["payment_hash"])
+    )
     assert status.paid
```
jackstar12, line 386: how is it possible that this test is passing on cln-rest? Get payment status requires a bolt11, but none is given here.

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+
+    async def cleanup(self):
+        try:
+            await self.client.aclose()
+        except RuntimeError as e:
+            logger.warning(f"Error closing wallet connection: {e}")
+
+    async def status(self) -> StatusResponse:
+        r = await self.client.get(f"{self.url}/v1/channel/localremotebal", timeout=5)
+
```
motorina0, line 53: why not use `raise_for_status()`?

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+        if r.is_error or "error" in r.json():
+            try:
+                data = r.json()
+                error_message = data["error"]
```
motorina0, line 57: - it is possible for `data` not to have an  `"error"` field (`or` condition at line `54`)

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+        if kwargs.get("expiry"):
+            data["expiry"] = kwargs["expiry"]
+
+        if kwargs.get("preimage"):
+            data["preimage"] = kwargs["preimage"]
+
+        r = await self.client.post(
+            f"{self.url}/v1/invoice/genInvoice",
+            data=data,
+            timeout=40,
```
motorina0, line 102: is `40` sec a known heuristic?

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+
```
motorina0, line 104: why not use `raise_for_status()`?

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+        if r.is_error or "error" in r.json():
+            try:
+                data = r.json()
+                error_message = data["error"]
```
motorina0, line 108: - it is possible for `data` not to have an  `"error"` field (`or` condition above)

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+        fee_limit_percent = fee_limit_msat / invoice.amount_msat * 100
+        r = await self.client.post(
+            f"{self.url}/v1/pay",
+            data={
+                "invoice": bolt11,
+                "maxfeepercent": f"{fee_limit_percent:.11}",
+                "exemptfee": 0,  # so fee_limit_percent is applied even on payments
+                # with fee < 5000 millisatoshi (which is default value of exemptfee)
+            },
+            timeout=None,
```
motorina0, line 132: why not timeout to `40` sec here? :)

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+        r = await self.client.get(
+            f"{self.url}/v1/pay/listPays",
+            params={"invoice": payment.bolt11},
+        )
+        try:
+            r.raise_for_status()
+            data = r.json()
+
+            if r.is_error or "error" in data or not data.get("pays"):
+                raise Exception("error in corelightning-rest response")
```
motorina0, line 187: - why not return a `PaymentStatus` in this particular case (the surrounding `try-catch` does)?

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+            if r.is_error or "error" in data or data.get("invoices") is None:
+                raise Exception("error in cln response")
+            return PaymentStatus(self.statuses.get(data["invoices"][0]["status"]))
+        except Exception as e:
+            logger.error(f"Error getting invoice status: {e}")
+            return PaymentStatus(None)
+
+    async def get_payment_status(self, payment: Payment) -> PaymentStatus:
+        # corelightning-rest wants the "bolt11" here.... sigh
+        r = await self.client.get(
```
motorina0, line 178: - this should be inside the `try-catch`. I believe it throws an exception if the URL is invalid or if the server is not reachable)

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+                        # paid_invoce = r.json()
+                        # logger.trace(f"paid invoice: {paid_invoce}")
+                        # yield paid_invoce["invoices"][0]["payment_hash"]
+
+            except Exception as exc:
+                logger.debug(
+                    f"lost connection to corelightning-rest invoices stream: '{exc}', "
+                    "reconnecting..."
+                )
+                await asyncio.sleep(0.02)
```
motorina0, line 241: in other places it slees `5` seconds (which feels more right)

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+class CoreLightningRestWallet(Wallet):
+    def __init__(self):
+        macaroon = settings.corelightning_rest_macaroon
+        assert macaroon, "missing cln-rest macaroon"
+
+        self.macaroon = load_macaroon(macaroon)
+
+        url = settings.corelightning_rest_url
+        if not url or not macaroon:
+            raise Exception("cannot initialize corelightning-rest")
```
motorina0, line 26: add a more detailed message

lnbits/wallets/corelightningrest.py
motorina0: I'll call this calle's constant

lnbits/wallets/corelightningrest.py
```diff
@@ -0,0 +1,241 @@
...
+
+        self.url = url[:-1] if url.endswith("/") else url
```
motorina0, line 28: - should check that it is a valid URL

.env.example
```diff
@@ -69,9 +69,9 @@ LNBITS_SITE_DESCRIPTION="Some description about your service, will display if ti
 LNBITS_THEME_OPTIONS="classic, bitcoin, flamingo, freedom, mint, autumn, monochrome, salvador, cyber"
 # LNBITS_CUSTOM_LOGO="https://lnbits.com/assets/images/logo/logo.svg"
 
-# Choose from LNPayWallet, OpenNodeWallet, ClicheWallet,
-#              LndWallet, LndRestWallet, CoreLightningWallet, EclairWallet,
-#              LnTipsWallet, LNbitsWallet, SparkWallet, FakeWallet,
+# which fundingsources are allowed in the admin ui
+LNBITS_ALLOWED_FUNDING_SOURCES="VoidWallet, FakeWallet, CoreLightningWallet, CoreLightningRestWallet, LndRestWallet, EclairWallet, LndWallet, LnTipsWallet, LNPayWallet, LNbitsWallet, OpenNodeWallet"
+
```
motorina0, line 74: - add:

```suggestion
CORELIGHTNING_REST_URL: https://localhost:3001
CORELIGHTNING_REST_MACAROON: ./access.macaroon
CORELIGHTNING_REST_CERT: .
```

PR #1777, opened 2023-06-22
update embit to 0.7.0

dni:
updating embit to latest, i am working on adding liquid support for the boltz extension and latest embit is crucial for that.
screenshot-1687432737

basically only boltz and watchonly are using it. @motorina0 can you check?


motorina0 (approved):
tACK

image

callebtc (approved):

PR #1781, opened 2023-06-25
[REFACTOR] payments sse endpoint

jackstar12:
The sse endpoint doesnt work if there is no properly configured reverse proxy that takes care of disabling gzip. A small wrapper around `GzipMiddleware` can fix that tough.

The code for the endpoint itself seems a bit overcomplicated to me aswell (why an additional queue?), refactored that aswell.

### DIFF
--- a/lnbits/core/models.py
+++ b/lnbits/core/models.py
@@ -169,6 +169,8 @@ class Payment(FromRowModel):
     async def set_pending(self, pending: bool) -> None:
         from .crud import update_payment_status
 
+        self.pending = pending
+
         await update_payment_status(self.checking_id, pending)
 
     async def check_status(
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -4,10 +4,9 @@ import json
 import uuid
 from http import HTTPStatus
 from io import BytesIO
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Union
 from urllib.parse import ParseResult, parse_qs, urlencode, urlparse, urlunparse
 
-import async_timeout
 import httpx
 import pyqrcode
 from fastapi import (
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -421,34 +421,18 @@ async def subscribe_wallet_invoices(request: Request, wallet: Wallet):
     logger.debug(f"adding sse listener for wallet: {uid}")
     api_invoice_listeners[uid] = payment_queue
 
-    send_queue: asyncio.Queue[Tuple[str, Payment]] = asyncio.Queue(0)
-
-    async def payment_received() -> None:
-        while True:
-            try:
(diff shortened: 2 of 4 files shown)


lnbits/core/views/api.py
```diff
@@ -421,34 +421,18 @@ async def subscribe_wallet_invoices(request: Request, wallet: Wallet):
     logger.debug(f"adding sse listener for wallet: {uid}")
     api_invoice_listeners[uid] = payment_queue
 
-    send_queue: asyncio.Queue[Tuple[str, Payment]] = asyncio.Queue(0)
-
-    async def payment_received() -> None:
-        while True:
-            try:
-                async with async_timeout.timeout(1):
```
jackstar12, line 429: why was there a timeout previously? it seems like it did nothing at all

lnbits/core/models.py
```diff
@@ -169,6 +169,8 @@ async def update_status(
     async def set_pending(self, pending: bool) -> None:
         from .crud import update_payment_status
 
+        self.pending = pending
```
jackstar12, line 172: the function previously only updated the db, now updates the pending attribute aswell (see comment later)

lnbits/core/views/api.py
```diff
@@ -421,34 +420,18 @@ async def subscribe_wallet_invoices(request: Request, wallet: Wallet):
...
-                    payment: Payment = await payment_queue.get()
-                    if payment.wallet_id == this_wallet_id:
-                        logger.debug("sse listener: payment received", payment)
-                        await send_queue.put(("payment-received", payment))
-            except asyncio.TimeoutError:
-                pass
-
-    task = asyncio.create_task(payment_received())
```
dni, line 437: haha, what a useless task this was! good catch

lnbits/core/views/api.py
```diff
@@ -421,34 +421,18 @@ async def subscribe_wallet_invoices(request: Request, wallet: Wallet):
...
-
     try:
         while True:
             if await request.is_disconnected():
                 await request.close()
                 break
-            typ, data = await send_queue.get()
-            if data:
-                jdata = json.dumps(dict(data.dict(), pending=False))
```
jackstar12, line 446: because `set_pending` previously did not set pending, it was set in here. Not necessary anymore

dni (approved):

lnbits/core/views/api.py
motorina0: why useless?`

jackstar12: what use did it have? It seems like it just added an extra queue values had to unnecessarily pass through

motorina0: > what use did it have?
- I do not know
- but also I couldn't figure out where messages are added to `payment_queue` (or to `api_invoice_listeners`). I could only locate `get/pop/items` calls.

motorina0 (approved):

Issue #1731, opened 2023-05-24
[BUG] New invoice not added to list


motorina0:
**Steps**:
- create invoice
- close pop-up

**Expected Result**:
 - new invoice is added to the table

**Actual Result**:
 - the new invoice is not added to the table
 - a browser refresh is required

https://github.com/lnbits/lnbits/assets/2951406/8c3da45e-e226-4050-88ba-3ca090c6c9b7

PR #1782, opened 2023-06-25
Update payments on invoice creation

jackstar12:
Closes: https://github.com/lnbits/lnbits/issues/1731

Payments are now refetched upon invoice creation. 

I was also wondering why the payment status is being polled regulary while sse does the same anyways. If there's a reason for that, let me know.

### DIFF
--- a/lnbits/core/static/js/wallet.js
+++ b/lnbits/core/static/js/wallet.js
@@ -419,6 +419,8 @@ new Vue({
             }
           }
 
+          this.fetchPayments()
+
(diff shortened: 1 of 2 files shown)


dni (approved):

callebtc:
I think the reason might have been that the SSE was very unreliable a while back. Maybe @arcbtc can chime in.

talvasconcelos (approved):

arcbtc (approved):

Issue #1666, opened 2023-04-30
[Feature request] Save FIAT value of transaction when possible


bota87:
**Is your feature request related to a problem? Please describe.**
Merchants that use the tpos extension at the end of the year need to give to the accountant an export with all the transactions with counter value in fiat

**Describe the solution you'd like**
Add the amount in fiat and change rate on the transaction history grid and on the `export to csv` button.

**Describe alternatives you've considered**
Use the current export with amounts in sats and integrate later with formulas on excel, but this require some skills and can make it more difficult to match the amounts between different lists (cash register logs for example) due to rounding

**Additional context**
![image](https://user-images.githubusercontent.com/52842374/235363228-1a21b2a1-265c-4a88-a01e-1899b2cb45e1.png)

callebtc:
We've recently added the fiat rate in the invoice comment of TPoS payments. Is that something that would serve your purpose?

bota87:
Sure that's a great help.
If it can be alone in a dedicated column to let easy sums it's even better

MichaelAntonFischer:
+1 for this. Having the fiat rate at the time of the transaction in the LNBits transaction report will be a great feature for merchants, since most countries mandate tracking of this info.

motorina0:
@talvasconcelos can you please kink this issue with the PRs (and other issues) that you are working on

PR #1789, opened 2023-06-28
track fiat value of payments

jackstar12:
Related: https://github.com/lnbits/lnbits/issues/1754, https://github.com/lnbits/lnbits/issues/1666

This is a basic implementation of storing the fiat value for payments at the transaction time.

I've just added a `fiat_amount` and `fiat_currency` column to the apipayments for now (store rate aswell?).

The user can set their preferred fiat currency in their wallet.
I also added a more general update wallet endpoint to avoid a new endpoint for each attribute of a wallet. Old one is still there

Current state:
image


talvasconcelos:
It would be cool to have the fiat value when received/paid and the now value!! for merchants, watching NgU is all they want!

arbadacarbaYK:
Please include in export too

jackstar12:
> Please include in export too

of course, one step at a time

callebtc:
Gave this a thought. I am against adding new columns to the `apipayments` table. LNbits is not required to track fiat prices, and works without contacting exchanges. 

The way to do is is to make use of the `extra` field in each row and not for all payments but only for those who care. One example is the TPoS extension, where people have to report taxes. Here, we have to put the **ENTERED** fiat value into the `extra` field (not the converted one) and make it possible to export (which can then include these new columns).

arcbtc:
Im with Calle. Although this level of functionality is better included as a separate extension that goes and fetches historic data. Been thinking of "Taxes" extension for exactly this. Im not against a general `extra` field. I am against  pinging exchanges.

arcbtc:
So we could have that functionality in TPoS

jackstar12:
If we integrate it into core I'd feel like columns would be more suitable. Putting it into `extra` would be unnecessary bloat for the database. If put into an extension, use of `extra` would obviously be the way to go

The name in `extra` would have to be consistent anyways - so its just a column hidden inside the json and any migration for that would be a lot more complicated than if the columns are there rightaway.
I think `extra` is best suited for extensions that might want to store some custom metadata - which they could even filter by with https://github.com/lnbits/lnbits/pull/1697.

arbadacarbaYK:
If its just in the tpos the devices (Vendingmachines, Beertaps, ATMs) are totally left apart, where it might be also needed. How about making it an admins choice and only write from the time its set to "on" and delete on "off" ?  Or query for the fiatprice asynchronously at not so busy times ?
Its about the export, where all payments from all extensions are mixed.

arcbtc:
I like this, but can we put the fiat currency where the current fee is (like wallet of satoshi, as they have the best UI)

arcbtc:
> If its just in the tpos the devices (Vendingmachines, Beertaps, ATMs) are totally left apart, where it might be also needed. How about making it an admins choice and only write from the time its set to "on" and delete on "off" ? Or query for the fiatprice asynchronously at not so busy times ? To me its about the export, where all payments from all extensions are mixed.

This is fine, as its using the same conversion function at the same time

lnbits/core/static/js/wallet.js
```diff
@@ -119,8 +119,8 @@ new Vue({
         paymentHash: null,
         minMax: [0, 2100000000000000],
         lnurl: null,
-        units: ['sat'],
-        unit: 'sat',
+        units: ['USD'],
+        unit: 'USD',
```
jackstar12, line 123: I dont really get these ones? The `unit` is always set to `sat` initially when opening the receive dialog and the units list shouldnt contain USD by default because it might not be included in the list of allowed currencies.

jackstar12:
> For tax purposes, it is not enough to later add an exchange rate, you need to track the exchange rate when making the transaction. Whenever the amount is tracked in a merchant's system in Fiat (connection to Woocommerce, Tpos, LNpos, etc.) the exact rate used at the time of the payment must be reported.

The exchange rate isnt added later, it stores the fiat value of the payment at the time of creation in the db. While the rate used could be stored aswell, its already stored implicitly through the ratio between the sat amount and fiat amount, so adding a computed column for an export wont be hard.

MichaelAntonFischer:
> > For tax purposes, it is not enough to later add an exchange rate, you need to track the exchange rate when making the transaction. Whenever the amount is tracked in a merchant's system in Fiat (connection to Woocommerce, Tpos, LNpos, etc.) the exact rate used at the time of the payment must be reported.
> 
> The exchange rate isnt added later, it stores the fiat value of the payment at the time of creation in the db. While the rate used could be stored aswell, its already stored implicitly through the ratio between the sat amount and fiat amount, so adding a computed column for an export wont be hard.

I think you misunderstanding me. This pull request is exactly what is needed. An extra colum in the transaction that can be exported.

But if you just have sats in the report and later on export/for taxes add the exchange rate, say from Coingecko, it's not accepted. Just had this checked by a tax advisor.

dni (approved):

callebtc:
Concept ACK
Implementation nACK

API payments should not get additional fiat columns.

jackstar12:
> Concept ACK Implementation nACK
> 
> API payments should not get additional fiat columns.

Putting data with concrete form that never changes into a json blob is just a lot of unnecessary bloat to the database. Applying any migration to such data would require additional effort aswell. I'm not dying on this hill tough

lnbits/core/static/js/wallet.js
arcbtc: Good point, I was going off wallets like WoS, which defaults to a fiat

arcbtc: But feel free to revert

lnbits/core/services.py
```diff
@@ -142,6 +163,10 @@ class PaymentKwargs(TypedDict):
             memo: str
             extra: Optional[Dict]
 
+        extra = await calculate_fiat_amounts(
+            invoice.amount_msat, wallet_id, extra=extra, conn=conn
+        )
+
```
callebtc, line 169: Could you move it above `class PaymentKwargs(TypedDict):` pls?

jackstar12:
> > Concept ACK Implementation nACK
> > API payments should not get additional fiat columns.
> 
> But this is exactly what we need. Maybe we can make it with a toggle, but for me every single payment needs to have a fiat value attached.

his comments were only about the internal implementation, the user will get the same info, no worries

callebtc:
> Putting data with concrete form that never changes into a json blob is just a lot of unnecessary bloat to the database. Applying any migration to such data would require additional effort aswell. I'm not dying on this hill tough

I apologize for my past ignorance as I now also grasp the value of #1697 – will reopen.

motorina0:
Just to sum up what my understanding is at this moment (after some telegram chats).

### Requirements
 - track the fiat value of the payment at the moment of the payment
 - having the value only in `sat` and then computing the fiat value is **NOT** acceptable
 - if an user enters the value in fiat (eg $2.35 for a coffee) then exactly this value has to be saved
   - back and forth conversions that might alter this value are **NOT** acceptable
 - do this in `core` so any extension can benefit from the functionality
 - grouping, sorting and filtering based on the fiat currency and fiat amount will take place outside LNbits (in the generated CSV/SreadSheet)

### Suggested Solution
The following changes have to be implemented:


1. The system must have a default currency (defined at the `.env` level and at the UI level)
  -  image

2. Each wallet has the option to overwrite the default currency and use its own
 -  image

3. update the `core.services.create_invoice` function signature (backwards compatible). The new function signature:
   - will take one extra parameters `fiat_amount`
   - either the `amount` or the `fiat_amount` must be provided
```py
async def create_invoice(
    *,
    wallet_id: str,
    amount: Optional[int],  # in satoshis
    amount_fiat: Optional[float],  # in fiat currency
   ...
```

4. update the `core.services.create_invoice` function implementation such that:
    - if  `fiat_amount` is provided then the `sat` value will be computed inside the `create_invoice` function
    - if the wallet corresponding to the `wallet_id` parameter has a fiat currency then this one will be used for computing the conversion, otherwise the system-wide fiat currency will be used
    - it saves the `amount_fiat`, `fiat_currency` and `exchange_rate` in the  `extra` field of the persisted `Payment`
      - yes, just save all 3 even if one can be deduced from the other two


PS: 
 - I can see that part of these requirements have already been implemented. I'm putting them here to make sure everyone has the same understanding

jackstar12:
> update the `core.services.create_invoice` function signature (backwards compatible). The new function signature:
> * will take one extra parameters `fiat_amount`
> * either the `amount` or the `fiat_amount` must be provided

A `fiat_currency` param would be required aswell, we cant know what currency the `fiat_amount` is in otherwise

jackstar12:
> it saves the `amount_fiat`, `fiat_currency` and `exchange_rate` in the `extra` field of the persisted `Payment`
> * yes, just save all 3 even if one can be deduced from the other two

I'm more for calculating the `exchange_rate` in the export step if its really required. Storing that would just be bloat

jackstar12:
What if the user creates an invoice with exactly $2.35 but the wallet tracking is set to EUR? Should it store USD or EUR? I think it would make more sense to store EUR in that case.

MichaelAntonFischer:
> What if the user creates an invoice with exactly $2.35 but the wallet tracking is set to EUR? Should it store USD or EUR? I think it would make more sense to store EUR in that case.

From tax perspective it needs to be $

jackstar12:
> > What if the user creates an invoice with exactly $2.35 but the wallet tracking is set to EUR? Should it store USD or EUR? I think it would make more sense to store EUR in that case.
> 
> From tax perspective it needs to be $, maybe it can track both?

If you're an european caffe and for whatever reason you make an invoice with the value of $2.35, wouldn't the tax authorities want to know the value in EUR? Do they care if you calculated the amount of satoshis with USD or Yuan?

Storing both would be possible aswell.

dni:
here is a simple pr storing whatever the user has requested from the ```api_create_invoice```
https://github.com/lnbits/lnbits/pull/1869

jackstar12:
image

We could show both in frontend like this:
- wallet currency first
- per-invoice currency second

If the invoice is created with the same currency as the wallet uses, no back and forth conversion is made

MichaelAntonFischer:
> @jackstar12
> 
> > What about this signature...
> 
> * then the system and wallet currency are not needed anymore (this can be a plus)
> * how will the `float` fractional value be handled for `sats`?
> 
> > I'm more for calculating the exchange_rate in the export step if its really required. Storing that would just be bloat
> 
> * this is not block-space, its DB space. Cheap as dirt.
> 
> > What if the user creates an invoice with exactly $2.35 but the wallet tracking is set to EUR? Should it store USD or EUR? I think it would make more sense to store EUR in that case.
> 
> * the merchant will create the invoice in its jurisdiction's currency
> * on the books the merchant prices its products in only one currency
> * if a client/tourist wants to see the amount in its own currency, then this is simply a UI requirement, the merchant does not have to persist this
> 
> > If you're an european caffe and for whatever reason you make an invoice with the value of $2.35, wouldn't the tax authorities want to know the value in EUR? Do they care if you calculated the amount of satoshis with USD or Yuan?
> 
> * we're just over-engineering now
> * I agree with the comment from @arbadacarbaYK : [track fiat value of payments #1789 (comment)](https://github.com/lnbits/lnbits/pull/1789#issuecomment-1669552410)
> 
> > We could show both in frontend like this: ...
> 
> * the requirement is really much simpler

On the last point:
At least the German authorities do care. They want exact tracking of what currency the user saw, what currency he paid and what currency the merchant has on the books.

Since this has something to do with AML, I assume the whole EU will treat it similarly.

So if the PoS shows USD, the merchant books on EUR and the payment is made in BTC, all three values would need to be tracked.

But I guess that almost no merchant will use a different currency for his book s than he will show to the customer.

> @jackstar12
> 
> > What about this signature...
> 
> * then the system and wallet currency are not needed anymore (this can be a plus)
> * how will the `float` fractional value be handled for `sats`?
> 
> > I'm more for calculating the exchange_rate in the export step if its really required. Storing that would just be bloat
> 
> * this is not block-space, its DB space. Cheap as dirt.
> 
> > What if the user creates an invoice with exactly $2.35 but the wallet tracking is set to EUR? Should it store USD or EUR? I think it would make more sense to store EUR in that case.
> 
> * the merchant will create the invoice in its jurisdiction's currency
> * on the books the merchant prices its products in only one currency
> * if a client/tourist wants to see the amount in its own currency, then this is simply a UI requirement, the merchant does not have to persist this
> 
> > If you're an european caffe and for whatever reason you make an invoice with the value of $2.35, wouldn't the tax authorities want to know the value in EUR? Do they care if you calculated the amount of satoshis with USD or Yuan?
> 
> * we're just over-engineering now
> * I agree with the comment from @arbadacarbaYK : [track fiat value of payments #1789 (comment)](https://github.com/lnbits/lnbits/pull/1789#issuecomment-1669552410)
> 
> > We could show both in frontend like this: ...
> 
> * the requirement is really much simpler

On the last point:
At least the German authorities do care. They want exact tracking of what currency the user saw, what currency he paid and what currency the merchant has on the books.

Since this has something to do with AML, I assume the whole EU will treat it similarly.

So if the PoS shows USD, the merchant books on EUR and the payment is made in BTC, all three values would need to be tracked.

But I guess that almost no merchant will use a different currency for his book s than he will show to the customer.

arbadacarbaYK:
imo this AML comment is only true if there is a conversion is done after payment from usd to euro which is not the case here, its just a frontend-setting

motorina0:
> imo this AML comment is only true if there a conversion is done after payment from usd to euro which is not the case here, its just a frontend-setting

I agree:
- the USD value (customer currency) its just a nice-to-have UX
- if there are any legal issues, the merchant can hide the USD and just display its jurisdiction currency (EUR)

jackstar12:
@motorina0 

> * then the system and wallet currency are not needed anymore (this can be a plus)

Thats not the case. The `currency` would specify the input amount - the wallet / system currency still defines what currency should be tracked. But this helps to avoid the back and forth conversion you mentioned above. All the conversions now happen in `calculate_fiat_amounts`, including the previous one in the api endpoint so that extensions could use it aswell.

> * how will the `float` fractional value be handled for `sats`?

I'd say just omit it (cast to int). The function signature could specify `float | int` aswell

jackstar12:
> I agree:
> 
> * the USD value (customer currency) its just a nice-to-have UX
> * if there are any legal issues, the merchant can hide the USD and just display its jurisdiction currency (EUR)

Its not required but also easy to implement - I'll leave that decision to you

jackstar12:
> @motorina0
> 
> > * then the system and wallet currency are not needed anymore (this can be a plus)
> 
> Thats not the case. The `currency` would specify the input amount - the wallet / system currency still defines what currency should be tracked. But this helps to avoid the back and forth conversion you mentioned above. All the conversions now happen in `calculate_fiat_amounts`, including the previous one in the api endpoint so that extensions could use it aswell. The wallet / system currency is still needed for calculating the value of outgoung payments aswell.
> 
> > * how will the `float` fractional value be handled for `sats`?
> 
> I'd say just omit it (cast to int). The function signature could specify `float | int` aswell

arbadacarbaYK:
> > I agree:
> > 
> > * the USD value (customer currency) its just a nice-to-have UX
> > * if there are any legal issues, the merchant can hide the USD and just display its jurisdiction currency (EUR)
> 
> Its not required but also easy to implement - I'll leave that decision to you

pls at least dont in the export, there it will just give a reason to speculate on conversions after payment by the authorities which have never happened

jackstar12:
Since https://github.com/lnbits/lnbits/pull/1869 now has been merged I see this as confirmation in storing both. Whether or not to include them in the export is another question.

dni (approved):

arcbtc (approved):

PR #1790, opened 2023-06-28
add cache utility

jackstar12:
This pr introduces a simple redis-like cache utility.

### Possible use cases:

- temporary caching of exchange rates, especially usefull with something like https://github.com/lnbits/lnbits/pull/1789
- caching node peer information or payment history in cln at https://github.com/lnbits/lnbits/pull/1702
- improving api key validation which currently queries the entire wallet, including balance, see: https://github.com/lnbits/lnbits/blob/bc55d52ea25475a524059fccbc5a35039f8be533/lnbits/decorators.py#L52-L55

### DIFF
--- a/lnbits/cache.py
+++ b/lnbits/cache.py
@@ -26,7 +26,7 @@ class Cache:
             if cached.expiry > time():
                 return cached.value
             else:
-                self.pop(key)
+                self._values.pop(key)
         return default
 
     def set(self, key: str, value: Any, expiry: float = 10):
@@ -34,7 +34,7 @@ class Cache:
 
     def pop(self, key: str, default=None) -> Optional[Any]:
         cached = self._values.pop(key, None)
-        if cached:
+        if cached and cached.expiry > time():
             return cached.value
         return default
 
--- a/lnbits/cache.py
+++ b/lnbits/cache.py
@@ -50,12 +50,14 @@ class Cache:
             self.set(key, value, expiry=expiry)
             return value
 
-    async def invalidate_forever(self, interval: float = 1):
+    async def invalidate_forever(self, interval: float = 10):
         while True:
             try:
                 await asyncio.sleep(interval)
                 ts = time()
-                self._values = {k: v for k, v in self._values.items() if v.expiry > ts}
+                expired = [k for k, v in self._values.items() if v.expiry < ts]
+                for k in expired:
+                    self._values.pop(k)
             except Exception:
                 logger.error("Error invalidating cache")
 
--- a/lnbits/cache.py
+++ b/lnbits/cache.py
@@ -1,0 +1,67 @@
+from __future__ import annotations
+
+import asyncio
+from time import time
+from typing import Any, NamedTuple, Optional
+
+from loguru import logger
+
+
+class Cached(NamedTuple):
+    value: Any
+    expiry: float
+
+
+def _add_prefix(key: str, prefix: str):
+    return prefix + ":" + key if prefix else key
+
+
+class Cache:
+    """
+    Small caching utility providing simple get/set interface (very much like redis)
+    """
+
+    def __init__(self):
+        self._values: dict[Any, Cached] = {}
+
+    def get(self, key: str, prefix: str = "") -> Optional[Any]:
+        cached = self._values.get(_add_prefix(key, prefix))
+        if cached is not None:
+            if cached.expiry > time():
+                return cached.value
+            else:
+                self.pop(key, prefix)
+        return None
+
+    def set(self, key: str, value: Any, expiry: float = 10, prefix: str = ""):
+        self._values[_add_prefix(key, prefix)] = Cached(value, time() + expiry)
+
+    def pop(self, key: str, prefix: str = "", default=None) -> Optional[Any]:
+        cached = self._values.pop(_add_prefix(key, prefix), None)
+        if cached:
+            return cached.value
+        return default
+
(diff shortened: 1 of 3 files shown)


callebtc:
cACK

jackstar12:
something like this could be used alternatively: https://pypi.org/project/expiring-dict/

arcbtc:
LGTM!
Caching exchange info would be very useful

lnbits/cache.py
```diff
@@ -0,0 +1,67 @@
...
+    def get(self, key: str, prefix: str = "") -> Optional[Any]:
+        cached = self._values.get(_add_prefix(key, prefix))
+        if cached is not None:
+            if cached.expiry > time():
+                return cached.value
+            else:
+                self.pop(key, prefix)
+        return None
+
+    def set(self, key: str, value: Any, expiry: float = 10, prefix: str = ""):
```
motorina0, line 36: `prefix` should go after `key` (for consistency with `get`)

lnbits/cache.py
```diff
@@ -0,0 +1,67 @@
...
+
+class Cache:
+    """
+    Small caching utility providing simple get/set interface (very much like redis)
+    """
+
+    def __init__(self):
+        self._values: dict[Any, Cached] = {}
+
+    def get(self, key: str, prefix: str = "") -> Optional[Any]:
```
motorina0, line 27: why no `defaultValue` (like `pop()`)

lnbits/cache.py
```diff
@@ -0,0 +1,67 @@
...
+            else:
+                self.pop(key, prefix)
+        return None
+
+    def set(self, key: str, value: Any, expiry: float = 10, prefix: str = ""):
+        self._values[_add_prefix(key, prefix)] = Cached(value, time() + expiry)
+
+    def pop(self, key: str, prefix: str = "", default=None) -> Optional[Any]:
+        cached = self._values.pop(_add_prefix(key, prefix), None)
+        if cached:
```
motorina0, line 41: shouldn't `cached.expiry > time()` also be checked here?

lnbits/cache.py
```diff
@@ -0,0 +1,67 @@
...
+            return cached.value
+        return default
+
+    async def save_result(self, coro, key: str, expiry: float = 10, prefix: str = ""):
```
motorina0, line 45: - order of params should be consistent with the other methods
- what is the advantage of this functionality? 
   - `coro` can be called at the line before `save_results` and the results passed to `save_results`
   - it only saves one line of code

callebtc:
LGTM except for the prefix, either I don't get it or it looks redundant to me.

lnbits/cache.py
callebtc: The cache is an unordered key-value store... what would you expect with an argument-less `pop()`?

lnbits/cache.py
```diff
@@ -0,0 +1,67 @@
...
+            value = await coro()
+            self.set(key, value, expiry=expiry, prefix=prefix)
+            return value
+
+    async def invalidate_forever(self, interval: float = 1):
+        while True:
+            try:
+                await asyncio.sleep(interval)
+                ts = time()
+                self._values = {k: v for k, v in self._values.items() if v.expiry > ts}
```
motorina0, line 62: - `invalidate_forever` runs as part of an `asyncio` task
- `invalidate_forever` has a `for` loop

What happens if an element is removed from `self._values.items()` (via `pop`) during the iteration?
Will it throw `"RuntimeError: dictionary changed size during iteration`?

lnbits/cache.py
motorina0: - if `coro` has params then a wrapper haws to be created
- I would remove this function, keep the interface as simple as possible

tests/core/test_cache.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+key = "foo"
+value = "bar"
+
+
+@pytest.mark.asyncio
+async def test_cache_get_set(cache):
+    cache.set(key, value)
+    assert cache.get(key) == value
```
motorina0, line 25: assert that calling `cache.get()` twice will still return `value`

lnbits/cache.py
jackstar12: > * it only saves one line of code

```python
rate = cache.get("ccy-usd")
if not rate:
    rate = await fetch_rate("usd")
    cache.set("ccy-usd", rate)
```
vs
```python
rate = await cache.save_result(lambda: fetch_rate("usd") key="ccy-usd")
```

lnbits/cache.py
jackstar12: thats not possible because asyncio is single-threaded

dni (approved):

jackstar12:
> Cache should not loop over all dictionary entries at every time step and not create a new dictionary to return. Instead use the common pattern for implementing cache:
> 
> * set value
> * kick of async task that waits
> * remove value

I adjusted the task to run every 10s instead of 1 and removed the recreation of the dict - which should be fine. Looping over dict with 100,000 entries takes 4ms on my machine. I think asyncio thanks us aswell for not scheduling thousands of callbacks

jackstar12:
In regards to `save_result`. I could remove it but I think its a pretty common pattern. I already use it in https://github.com/lnbits/lnbits/pull/1702 and it would fit really well into the currency rates and wallet info caching aswell

callebtc (approved):
LGTM

PR #1791, opened 2023-06-29
ip blocker only call next after validation

dni:
only call next after validation

### DIFF
--- a/lnbits/middleware.py
+++ b/lnbits/middleware.py
@@ -1,3 +1,4 @@
+import asyncio
 from http import HTTPStatus
 from typing import Any, List, Tuple, Union
 from urllib.parse import parse_qs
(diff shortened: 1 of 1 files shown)


lnbits/middleware.py
```diff
@@ -205,19 +205,20 @@ def add_ratelimit_middleware(app: FastAPI):
...
+                status_code=403,  # Forbidden
                 content={"detail": "No request client"},
             )
         if request.client.host in settings.lnbits_allowed_ips:
+            response = await call_next(request)
             return response
         if request.client.host in settings.lnbits_blocked_ips:
             return JSONResponse(
-                status_code=429,
+                status_code=403,  # Forbidden
```
motorina0, line 218: - can we add a `sleep(n)` so the blocked IP does not get an immediate response?

dni: done i choose `5`

motorina0 (approved):

callebtc (approved):
LGTM

PR #1793, opened 2023-06-29
cancel all long-running tasks

jackstar12:
Not all long-running tasks are cancelled upon shutdown, which caused some issues with `WALLET.cleanup`.

I introduced a few new functions:

- `create_task` basically a wrapper around the asyncio equivalent, adds the task to a list
- `cancel_all_tasks` cancels all tasks previously created with `create_task` (if any are already stopped, `cancel` will do nothing)
- `create_permanent_task` wrapper around `create_task` with `catch_everything_and_restart`

Now all tasks are cancelled on shutdown and `WALLET.cleanup` goes without errors.

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -357,7 +357,7 @@ def initialize_server_logger():
             msg = await serverlog_queue.get()
             await websocketUpdater(super_user_hash, msg)
 
-    asyncio.create_task(update_websocket_serverlog())
+    create_permanent_task(update_websocket_serverlog)
 
     logger.add(
         lambda msg: serverlog_queue.put_nowait(msg),
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -28,9 +28,9 @@ from lnbits.core.services import websocketUpdater
 from lnbits.core.tasks import (  # register_watchdog,; unregister_watchdog,
     register_killswitch,
     register_task_listeners,
-    unregister_killswitch,
 )
 from lnbits.settings import settings
+from lnbits.tasks import cancel_all_tasks, create_permanent_task
 from lnbits.wallets import get_wallet_class, set_wallet_class
 
 from .commands import db_versions, load_disabled_extension_list, migrate_databases
@@ -52,7 +52,6 @@ from .middleware import (
 )
 from .requestvars import g
 from .tasks import (
-    catch_everything_and_restart,
     check_pending_payments,
     internal_invoice_listener,
     invoice_listener,
@@ -95,7 +94,7 @@ def create_app() -> FastAPI:
     register_routes(app)
     register_async_tasks(app)
     register_exception_handlers(app)
-    register_shutdown(app)
+    register_wallet_cleanup(app)
 
     # Allow registering new extensions routes without direct access to the `app` object
     setattr(core_app_extra, "register_new_ext_routes", register_new_ext_routes(app))
@@ -338,7 +337,7 @@ def register_startup(app: FastAPI):
             raise ImportError("Failed to run 'startup' event.")
 
 
-def register_shutdown(app: FastAPI):
+def register_wallet_cleanup(app: FastAPI):
     @app.on_event("shutdown")
     async def on_shutdown():
         WALLET = get_wallet_class()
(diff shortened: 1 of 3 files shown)


dni (approved):

dni (approved):
LGTM

lnbits/tasks.py
```diff
@@ -19,6 +19,23 @@
...
+    return task
+
+
+def create_permanent_task(func):
+    return create_task(catch_everything_and_restart(func))
+
+
+def cancel_all_tasks():
+    for task in tasks:
+        task.cancel()
```
motorina0, line 37: - loop body should be wrapped in `try-catch`

lnbits/app.py
```diff
@@ -396,21 +398,13 @@ async def webhook_listener():
 
     @app.on_event("startup")
     async def listeners():
```
motorina0, line 400: - any reason for `async def listeners()` to be `async` now?

lnbits/core/tasks.py
```diff
@@ -20,24 +25,13 @@
...
-async def register_killswitch():
+def register_killswitch():
     """
     Registers a killswitch which will check lnbits-status repository
     for a signal from LNbits and will switch to VoidWallet if the killswitch is triggered.
     """
     logger.debug("Starting killswitch task")
-    global killswitch
-    killswitch = asyncio.create_task(killswitch_task())
-
```
motorina0, line 31: - any reason for `killswitch: Optional[asyncio.Task] = None` to still be present?
- see: https://github.com/lnbits/lnbits/blob/2928209659950f27cadd4701f5a10c84a75efe2b/lnbits/core/tasks.py#L24

lnbits/core/tasks.py
```diff
@@ -73,16 +67,6 @@ async def register_watchdog():
     # watchdog = asyncio.create_task(watchdog_task())
 
 
-async def unregister_watchdog():
```
motorina0, line 76: - since `watchdog` was also converted please adapt the commented code to the new paradigm

https://github.com/lnbits/lnbits/blob/2928209659950f27cadd4701f5a10c84a75efe2b/lnbits/core/tasks.py#L67

lnbits/server.py
```diff
@@ -1,14 +1,10 @@
-import uvloop
-from uvicorn.supervisors import ChangeReload
-
-uvloop.install()
```
motorina0, line 4: - please help me understand, what was this line doing and why it is not required now

lnbits/tasks.py
```diff
@@ -19,6 +19,23 @@
 
 from .core import db
 
+tasks: List[asyncio.Task] = []
+
+
+def create_task(coro_or_fut):
```
motorina0, line 25: - name nitpick: change `coro_or_fut` to `coro_or_func` (consistent with the below use and more clear)

lnbits/wallets/cln.py
```diff
@@ -23,23 +22,9 @@
...
-    return run
-
-
-def _pay_invoice(ln, payload):
-    return ln.call("pay", payload)
-
-
-def _paid_invoices_stream(ln, last_pay_index):
-    return ln.waitanyinvoice(last_pay_index)
+async def run_sync(func) -> Any:
```
motorina0, line 25: - how is this related to `asyncio.create_task` improvements?

dni: https://github.com/lnbits/lnbits/pull/1858

lnbits/tasks.py
jackstar12: `task.cancel` doesnt throw, even if its already canceled

lnbits/tasks.py
jackstar12: `coro_or_fut` is the same as asyncio uses themselves. It doesnt have to be the same as below because it isn't, because the below takes a func which creates a coroutine

lnbits/wallets/cln.py
jackstar12: this fix was required to get tests to pass. cln would otherwise get stuck

lnbits/server.py
jackstar12: see commit description: https://github.com/lnbits/lnbits/pull/1860/commits/ec2aed001d79bec044fc46644bc55212f18c0a48

jackstar12:
another interesting side effect of this is that extensions are loaded inside tests aswell, but I dont think this should be considered an issue

lnbits/tasks.py
jackstar12: `coro` would be the most accurate in this case

motorina0 (approved):
LGTM

great job @jackstar12 !

dni:
> > another interesting side effect of this is that extensions are loaded inside tests aswell, but I dont think this should be considered an issue
> 
>     * only extensions listed in `LNBITS_EXTENSIONS_DEFAULT_INSTALL` should be installed by default
> 
>     * if this value is empty no extension will be installed

good catch, dont install default exts for test when run locally! :)
https://github.com/lnbits/lnbits/pull/1861

PR #1796, opened 2023-07-03
Fix ip blocker 2nd try

dni:
Fix ip blocker 2nd try

### DIFF
--- a/lnbits/middleware.py
+++ b/lnbits/middleware.py
@@ -215,7 +215,6 @@ def add_ip_block_middleware(app: FastAPI):
             request.client.host in settings.lnbits_blocked_ips
             and request.client.host not in settings.lnbits_allowed_ips
         ):
-            await asyncio.sleep(5)
             return JSONResponse(
                 status_code=403,  # Forbidden
                 content={"detail": "IP is blocked"},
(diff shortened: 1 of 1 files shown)


lnbits/middleware.py
```diff
@@ -215,11 +215,18 @@ async def block_allow_ip_middleware(request: Request, call_next):
...
                 content={"detail": "IP is blocked"},
             )
-        return await call_next(request)
+        # this is not needed on latest FastAPI
+        # https://stackoverflow.com/questions/71222144/runtimeerror-no-response-returned-in-fastapi-when-refresh-request
+        # TODO: remove after https://github.com/lnbits/lnbits/pull/1609 is merged
+        try:
+            return await call_next(request)
+        except RuntimeError as exc:
+            if str(exc) == "No response returned." and await request.is_disconnected():
```
motorina0, line 228: - is this logic the responsibility of `add_ip_block_middleware`?
- if need, it should be a special/dedicated middleware for this type of error handling. 
- `ip_block` should only block IPs

lnbits/middleware.py
```diff
@@ -215,11 +214,18 @@ async def block_allow_ip_middleware(request: Request, call_next):
             request.client.host in settings.lnbits_blocked_ips
             and request.client.host not in settings.lnbits_allowed_ips
         ):
-            await asyncio.sleep(5)
             return JSONResponse(
                 status_code=403,  # Forbidden
                 content={"detail": "IP is blocked"},
             )
-        return await call_next(request)
+        # this is not needed on latest FastAPI
```
callebtc, line 221:
```suggestion
        # this try: except: block is not needed on latest FastAPI (await call_next(request) is enough)
```

PR #1805, opened 2023-07-04
FEAT: proper failing invoice/payment responses for VoidWallet

dni:
before that change anyone using the api just got a 500 error without any error message. there was also a bug stacetrace in the logs each time.


motorina0 (approved):

arcbtc (commented):
LFGTM

arcbtc (approved):
Great

PR #1814, opened 2023-07-10
fix cln bug

jackstar12:


callebtc:
Thank you for this, I didn't spot it.

PR #1609, opened 2023-04-04
[CHORE] update dependencies, unsafe pip packages, fastapi

dni:
## security issues
* update fastapi / starlette because of this https://github.com/lnbits/lnbits/security/dependabot/20
* https://github.com/lnbits/lnbits/security/dependabot/18 this one is related to pyln-client depending on it. updating to latest did not help. same goes for this issue. created issue on corelightning repo, and it got into the roadmap for coming up release. https://github.com/ElementsProject/lightning/issues/6164

### parsing of url params
some of the extension are broken, see following screenshot, for latest fastapi it is not allowed anymore to Query for a url path parameter, like```("/api/v1/{name}") and myfn(name: str = Query(none))```, just use plain type.```myfn(name: str )```

screenshot-1681055593

screenshot-1681055547

- unfortunately this issue got copied across multiple extensions and will require fixing those aswell in some places.

### Request url_for
starlettes request.url_for changed the return type to `URL` instead of  `str` . you can just to `str(req.url_for())` to make the types work again.

## corelightning
update to latest pyln-client, it is also possible now to support latest corelightning node 23.02.2 because it does not break description_hash anymore

PR #1815, opened 2023-07-10
remove cashu, add websocket-client, python-dotenv, pycryptodomex

dni:
remove cashu lib and add missing ones

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,23 +2,24 @@
 
 [[package]]
 name = "anyio"
-version = "3.6.2"
+version = "3.7.1"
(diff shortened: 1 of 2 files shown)


motorina0:
The missing ones were installed because they were indirect dependencies?

motorina0 (approved):

dni:
this will break cashu extension, @callebtc needs to look at this, either keep cashu here, or add those missing dependencies from cashu here.
at the moment with cashu every lib update in lnbits will conflict with cashu, because it is locking to the same versions as lnbits has.

dni:
https://github.com/lnbits/lnbits/pull/1609 included in following

PR #1817, opened 2023-07-10
adding bolt11 lib and removing bolt11.py from the codebase

dni:
reintroducing updated bolt11 lib https://github.com/lnbits/bolt11


lnbits/bolt11.py
```diff
@@ -1,366 +1,5 @@
-import hashlib
-import re
-import time
-from decimal import Decimal
-from typing import List, NamedTuple, Optional
+from bolt11.decode import decode  # noqa: F401
+from bolt11.encode import encode  # noqa: F401
+from bolt11.types import Bolt11
```
jackstar12, line 3: Manual assignment can be replaced by that:
```suggestion
from bolt11.types import Bolt11 as Invoice
```

arcbtc (approved):

lnbits/wallets/fake.py
```diff
@@ -2,13 +2,18 @@
...
 from datetime import datetime
-from typing import AsyncGenerator, Dict, Optional
+from os import urandom
+from typing import AsyncGenerator, Optional
 
+from bolt11.decode import decode
+from bolt11.encode import encode
+from bolt11.exceptions import Bolt11Exception
+from bolt11.models.tags import TagChar, Tags
+from bolt11.types import Bolt11, MilliSatoshi
```
jackstar12, line 12: cant most of this be imported from `bolt11` directly?

pyproject.toml
```diff
@@ -32,8 +31,12 @@ protobuf = "4.21.12"
 Cerberus = "1.3.4"
 async-timeout = "4.0.2"
 pyln-client = "23.8"
-cashu = "0.9.0"
 slowapi = "^0.1.7"
+python-dotenv = "^1.0.0"
+websocket-client = "^1.6.1"
+secp256k1 = "^0.14.0"
+pycryptodomex = "^3.18.0"
```
jackstar12, line 38: whats with those ones?

jackstar12 (approved):

lnbits/core/views/api.py
```diff
@@ -621,29 +621,20 @@ async def api_lnurlscan(code: str, wallet: WalletTypeInfo = Depends(get_key_type
...
-                "amount_msat": invoice.amount_msat,
-                "description": invoice.description,
-                "description_hash": invoice.description_hash,
-                "payee": invoice.payee,
-                "date": invoice.date,
-                "expiry": invoice.expiry,
-                "secret": invoice.secret,
-                "route_hints": invoice.route_hints,
-                "min_final_cltv_expiry": invoice.min_final_cltv_expiry,
-            }
```
michael1011, line 643: This is a breaking API change. Above is the new version and underneath latest main
image

Shouldn't cause any major issues with backwards compatibility though

Funnily enough, the main version decodes the `min_final_cltv_expiry` incorrectly

pyproject.toml
dni: this is a bit weird because those deps are needed in lnbits but got removed at some point and are pulled from the `cashu` package, which should actually be removed at some point.

lnbits/bolt11.py
```diff
@@ -1,365 +1,3 @@
...
-def bitarray_to_u5(barr):
-    assert barr.len % 5 == 0
-    ret = []
-    s = bitstring.ConstBitStream(barr)
-    while s.pos != s.len:
-        ret.append(s.read(5).uint)  # type: ignore
-    return ret
+from bolt11.decode import decode  # noqa: F401
+from bolt11.encode import encode  # noqa: F401
+from bolt11.types import Bolt11 as Invoice  # noqa: F401
```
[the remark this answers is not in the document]
dni: it is backwards compatible

lnbits/core/views/api.py
dni: hm i dont see a lot of problem, there are only additional fields, but the ones that were there before are there also

theborakompanioni: > hm i dont see the problem, there are only additional fields [...]

As far as I can see, `secret` changed to `payment_secret` and `route_hints` is missing - this will easily break existing clients.

lnbits/core/crud.py
```diff
@@ -5,8 +5,8 @@
 from uuid import UUID, uuid4
 
 import shortuuid
+from bolt11.decode import decode
```
motorina0, line 8: - why not use `lnbits.bolt11` import?

lnbits/core/services.py
```diff
@@ -124,16 +124,17 @@ async def pay_invoice(
     will regularly check for the payment.
     """
     invoice = bolt11.decode(payment_request)
+
```
motorina0, line 127: - is this just a casual refactoring or is it related to the `bolt11` changes?

lnbits/core/crud.py
motorina0: - same in `cln.py` and `fake.py`

lnbits/core/views/api.py
motorina0: @dni why is this marked as resolved?

lnbits/wallets/corelightning.py
```diff
@@ -3,10 +3,11 @@
 from functools import partial, wraps
 from typing import AsyncGenerator, Optional
 
+from bolt11.decode import decode as bolt11_decode
```
motorina0, line 6: - why not import from `lnbits.bolt11`?

lnbits/wallets/corelightningrest.py
```diff
@@ -4,9 +4,9 @@
 from typing import AsyncGenerator, Dict, Optional
 
 import httpx
+from bolt11.decode import decode
```
motorina0, line 7: - why not import from `lnbits.bolt11` (like `lnbits/core/services.py`)?

lnbits/wallets/corelightningrest.py
```diff
@@ -129,7 +129,10 @@ async def create_invoice(
         return InvoiceResponse(True, label, data["bolt11"], None)
 
     async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse:
-        invoice = lnbits_bolt11.decode(bolt11)
+        invoice = decode(bolt11)
```
motorina0, line 132: - doesn't `decode` now throw `Bolt11Exception` if it fails?
- `try-catch` should be added as in `lnbits/wallets/corelightning.py`

lnbits/wallets/fake.py
```diff
@@ -2,13 +2,22 @@
 import hashlib
 import random
 from datetime import datetime
-from typing import AsyncGenerator, Dict, Optional
+from os import urandom
+from typing import AsyncGenerator, Optional
 
+from bolt11 import (
```
motorina0, line 8: - why not import from `lnbits.bolt11` (like in `lnbits/core/services.py`)?

lnbits/wallets/fake.py
```diff
@@ -42,44 +51,55 @@ async def create_invoice(
...
+            tags.add(
+                TagChar.description_hash,
+                hashlib.sha256(unhashed_description).hexdigest(),
+            )
         else:
-            data["tags_set"] = ["d"]
-            data["memo"] = memo
-            data["description"] = memo
-        randomHash = (
+            tags.add(TagChar.description, memo or "")
```
motorina0, line 68: - does `Tags` take care to set both the `memo` and `description` fields?

lnbits/wallets/fake.py
```diff
@@ -42,44 +51,55 @@ async def create_invoice(
         memo: Optional[str] = None,
         description_hash: Optional[bytes] = None,
         unhashed_description: Optional[bytes] = None,
-        **kwargs,
+        expiry: Optional[int] = None,
+        payment_secret: Optional[bytes] = None,
+        **_,
     ) -> InvoiceResponse:
-        data: Dict = {
-            "out": False,
```
motorina0, line 48: - `out` is no longer used, is that ok?

lnbits/core/crud.py
dni: `lnbits.bolt11` is only a compat wrapper for extensions that used it, it should be avoided.

lnbits/core/services.py
dni: so bolt11 lib supports amountless invoices, but lnbits does not. so this is casual refactoring to make mypy happy with those optional amounts.

also i moved it before we open a database connection, because we can actually raise before that.

lnbits/core/crud.py
motorina0: > it should be avoided
- why? I think it should be the other way around (use the wrapper if you already have it)
- if `bolt11` lib is replaced with some other lib, or if we want to do some pre/post processing then this wrapper is very useful
- one only needs to change `lnbits.bolt11` and not touch the rest of the codebase

motorina0: @dni ⬆️ ?

lnbits/wallets/fake.py
dni: there is no memo field in bolt11

lnbits/wallets/fake.py
dni: yes, it was unused, at least my tests worked fine.

lnbits/wallets/corelightningrest.py
dni: done

PR #1822, opened 2023-07-18
Don't create user if allowed users is set

### LINKED ISSUE #1747: [BUG] No way to block the creation of users.
**Describe the bug**
Setting allowed users in the GUI and/or the .env file blocks access to non-allowed users, but does not block the creation of new users through the public frontend page IE lnbits.mydomain.com

**To Reproduce**

1. As admin user, create a new users in the usermanager extension.
2. Set this new user as "allowed user" in the admin GUI and the .env.
3. Restart LNbits.
4. Visit the public frontend and try to create a new user. The GUI states "User not authorized." but the new user is in fact still created, his userID is visible in the URL and the creation is logged.

**Expected behavior**
Setting "allowed users" blocks the creation of new users by anyone other than the admin.
A clear message is displayed that visitors of the frontend are not allowed to create new users.

LNbits version: 0.10.8
Database: sqlite

talvasconcelos:
When allowed users is set, the new user is created and then invalidated in the backend. This PR aims to not create the user at all and fail before creating the user. 
Needs some more testing and edge cases... or possible issues from tackling the issue like this (or it it's worth it). From local testing, it works as expected: block creation of new users, if allowed_users is set and has at least a user! 

Closes #1747

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,6 +15,7 @@ LNBITS_ALLOWED_IPS=""
 LNBITS_BLOCKED_IPS=""
 
 # Allow users and admins by user IDs (comma separated list)
+# if set new users will not be able to create accounts
 LNBITS_ALLOWED_USERS=""
 LNBITS_ADMIN_USERS=""
 # Extensions only admin can access
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,7 +5,7 @@
 
 *.yml
 
-**/lnbits/extensions/*
+# **/lnbits/extensions/*
 **/lnbits/upgrades/*
 
 **/lnbits/static/vendor
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,7 +5,7 @@
 
 *.yml
 
-# **/lnbits/extensions/*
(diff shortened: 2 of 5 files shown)


dni:
i think it ok that usermanager cant create users when allowed users is used... they are not compatible.

we should though return a proper error message, that creating is not allowed due to this feature beeing enabled.
and i should not be checked in crud it should be on api level.

talvasconcelos:
What if i'm an admin creating users that way? otherwise, how are new users
created?

Probably this is not a bug... it's a feature, and not worth tackling, or
best wait for login?

motorina0:
- is this PR still relevant?
- it is in draft and tests fail

dni (approved):

motorina0:
It should be specified in the `.env.example` that new users will not be created if `LNBITS_ALLOWED_USERS` is set.

```env
# Allow users and admins by user IDs (comma separated list)
LNBITS_ALLOWED_USERS=""
LNBITS_ADMIN_USERS=""
```

PR #1823, opened 2023-07-18
Data returned from wallet websocket was not json

arcbtc:


motorina0 (approved):
LGTM

dni:
serviceworker.js is not needed here.

and why is it payment `.dict()` or `._asdict()` that weird.

PR #1824, opened 2023-07-19
Adds useful websocket price feed for all currencies

arcbtc:


arcbtc:
Wait, pause, https://github.com/lnbits/lnbits/pull/1790 will make it better

PR #1825, opened 2023-07-19
Increase default invoice expiry to 1 hour

dni:
default is 600 = 10 minutes, which is a pretty short amount of time, for example when using lnbits fundingsource inside lnbits itself it would fail if you do a submarine swap or anything involved with hold invoices

i think 3600 = 1 hour should not make a difference in performance and is a way more reasonable default expiry


callebtc:
ACK 

also changed the default in `.env.example`

callebtc (approved):

talvasconcelos (approved):

PR #1826, opened 2023-07-19
Add deleted flag wallet

### LINKED ISSUE #1706: [BUG] Direct link of a TPoS continues to work although the wallet to which it was linked has been deleted
**Describe the bug**

Today I created a wallet on legends demo server, there I created a TPoS terminal. After experimenting, I deleted the wallet. But the problem is that the direct link of the terminal continues to work, it even continues to accept payment, although the wallet to which it was linked has been deleted. The terminal even has the name I gave it. Apparently, some of the data in the database tables is not deleted after deleting the wallet.

talvasconcelos:
Adds a `deleted` column to wallets DB.

On `deleteWallet`, set flag to `true` don't prefix keys with `del:`

Invalidate wallet on KeyChecker if wallet is deleted, makes all request dependent on wallet key `Depends(...)` fail if wallet is deleted!

Closes #1706

### DIFF
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -253,11 +253,8 @@ async def delete_wallet(
 ) -> None:
     await (conn or db).execute(
         """
-        UPDATE wallets AS w
-        SET
-            "user" = 'del:' || w."user",
-            adminkey = 'del:' || w.adminkey,
-            inkey = 'del:' || w.inkey
+        UPDATE wallets
+        SET deleted = true
         WHERE id = ? AND "user" = ?
         """,
         (wallet_id, user_id),
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -290,7 +290,10 @@ async def get_wallet_for_key(
 
     if not row:
         return None
-
+    
+    if row["deleted"]:
+        return None
+    
     if key_type == "admin" and row["adminkey"] != key:
         return None
 
--- a/lnbits/core/migrations.py
+++ b/lnbits/core/migrations.py
@@ -301,3 +301,53 @@ async def m010_create_installed_extensions_table(db):
         );
     """
     )
+
+async def m011_add_deleted_to_wallets(db):
+    """
+    Adds deleted column to wallets.
+    """
+    try:
+        await db.execute("ALTER TABLE wallets ADD COLUMN deleted BOOLEAN")
+    except OperationalError:
+        pass
+
+async def m012_set_deleted_wallets(db):
+    """
+    Sets deleted column to wallets.
+    """
+    try:
+        rows = await (
+            await db.execute(
+                f"""
+                SELECT *
+                FROM wallets
+                WHERE user LIKE 'del:%'
+                AND adminkey LIKE 'del:%'
+                AND inkey LIKE 'del:%'
+                """
+            
+            )
+        ).fetchall()
+        
+        for row in rows:
+            try:
+                user = row[2].split(":")[1]
+                adminkey = row[3].split(":")[1]
+                inkey = row[4].split(":")[1]
+                await db.execute(
+                    """
+                    UPDATE wallets SET user = ?, adminkey = ?, inkey = ?, deleted = true
+                    WHERE id = ?
+                    """,
+                    (
+                        user, adminkey, inkey, row[0]
+                    ),
+                )
+            except:
+                continue
+    except OperationalError:
+        # this is necessary now because it may be the case that this migration will
+        # run twice in some environments.
+        # catching errors like this won't be necessary in anymore now that we
+        # keep track of db versions so no migration ever runs twice.
+        pass


lnbits/core/crud.py
```diff
@@ -294,6 +294,9 @@ async def get_wallet_for_key(
     if not row:
         return None
 
+    if row["deleted"]:
+        return None
```
callebtc, line 298:
```suggestion
    if row.get("deleted"):
        return None
```

callebtc:
Please write tests: 

1) delete existing wallet
2) use get_wallet and verify that the return is `None`

callebtc (changes requested):
LGTM but tests are missing. One suggestion:

1) delete existing wallet
2) use get_wallet and verify that the return is `None`

lnbits/core/crud.py
```diff
@@ -276,6 +273,9 @@ async def get_wallet(
         (wallet_id,),
     )
 
+    if row["deleted"]:
+        return None
+
```
dni, line 278: do this in the sql query ```WHERE id = ? and deleted = false

lnbits/core/crud.py
```diff
@@ -294,6 +294,9 @@ async def get_wallet_for_key(
     if not row:
         return None
 
+    if row["deleted"]:
+        return None
+
```
dni, line 299: here aswell

lnbits/core/migrations.py
```diff
@@ -301,3 +301,52 @@ async def m010_create_installed_extensions_table(db):
         );
     """
     )
+
+
+async def m011_add_deleted_to_wallets(db):
```
dni, line 306: i would combine the migrations, into 1

callebtc: I think I would keep it like this, the second one could fail but the first one is very safe. If the column is created but the second stage fails, the migration counter won't be incremented but the column will still exist.

lnbits/core/crud.py
talvasconcelos: deleted doesn't have a default `false`... if it's not `true` it's just `Null`, does it work like that?

talvasconcelos: ok, just on the safe side: `WHERE id = ? AND deleted is NULL or deleted = false`

talvasconcelos:
> 

i did those tests

talvasconcelos:
> LGTM but tests are missing. One suggestion:
> 
> 1. delete existing wallet
> 2. use get_wallet and verify that the return is `None`

I did those tests.

arcbtc:
Can you fix conflicts. Would like to merge

motorina0:
- should check for wallet `deleted` here:
   - `SELECT *, COALESCE((SELECT balance FROM balances WHERE wallet = wallets.id), 0) AS balance_msat`
   - https://github.com/lnbits/lnbits/blob/e0843e1b9db3b99e81d890059bb6374696824ebc/lnbits/core/crud.py#L60C4

lnbits/core/migrations.py
```diff
@@ -301,3 +301,52 @@ async def m010_create_installed_extensions_table(db):
...
+    """
+    Adds deleted column to wallets.
+    """
+    try:
+        await db.execute("ALTER TABLE wallets ADD COLUMN deleted BOOLEAN")
```
motorina0, line 311: - it is much simpler not to allow NULL: `BOOLEAN NOT NULL DEFAULT false`
  - if necessary the value will be set to `true` in `m012_set_deleted_wallets`
  - the above condition `AND deleted is NULL OR deleted = false` will reduce to `AND deleted = false`

arcbtc (approved):

dni (approved):

callebtc:
I've deleted the checks from `crud.py` and added appropriate checks in the `KeyChecker` decorator instead. Wallets should not just "disappear" when they are deleted. The code might still need to access them and changing the behavior of a crud is not the best way to filter for undeleted wallets I think.

Hope this works for everyone!

callebtc:
We still have to figure out a way how we want to handle the effect on the total instance's balance when a wallet is deleted. I think we should not ignore the balance of a wallet that is marked as deleted. Rather, if a user would truly want to get rid of the balance as well, they we should make a transaction similar to the admin top up that nulls the balance before deleting. This could be done with a "Drain wallet balance" check box after the delete wallet button in the frontend and via a parameter in the API.

callebtc (approved):

lnbits/core/migrations.py
```diff
@@ -327,3 +327,54 @@ async def m012_add_currency_to_wallet(db):
...
+                inkey = row[4].split(":")[1]
+                await db.execute(
+                    """
+                    UPDATE wallets SET user = ?, adminkey = ?, inkey = ?, deleted = true
+                    WHERE id = ?
+                    """,
+                    (user, adminkey, inkey, row[0]),
+                )
+            except Exception:
+                continue
```
motorina0, line 374: - add a log statement so it can be manually checked later if something goes wrong for a record

lnbits/core/migrations.py
```diff
@@ -327,3 +327,54 @@ async def m012_add_currency_to_wallet(db):
...
+                WHERE user LIKE 'del:%'
+                AND adminkey LIKE 'del:%'
+                AND inkey LIKE 'del:%'
+                """
+            )
+        ).fetchall()
+
+        for row in rows:
+            try:
+                user = row[2].split(":")[1]
```
motorina0, line 363: - why use the `int` index (eg: `row[2]`) instead of the key (eg: `row["user"]`)

motorina0:
- `wallets=[Wallet(**w) for w in wallets],` here: https://github.com/lnbits/lnbits/pull/1826/files#diff-c9e3edbdea19532e89995cca6a5265e261a98996514ef3af5ff1c4d3d1fe1751R78
  - are deleted wallets for user returned?
  - how did it worked before the change?

```py
return User(
        id=user["id"],
        email=user["email"],
        extensions=[
            e[0] for e in extensions if User.is_extension_for_user(e[0], user["id"])
        ],
        wallets=[Wallet(**w) for w in wallets],
        admin=user["id"] == settings.super_user
        or user["id"] in settings.lnbits_admin_users,
        super_user=user["id"] == settings.super_user,
)
``` 
- should `wallets=[Wallet(**w) for w in wallets],` 
  - be `wallets=[Wallet(**w) for w in wallets if not w.deleted],`?

motorina0:
- here: https://github.com/lnbits/lnbits/pull/1826/files#diff-0f59f07ce344c28b48f1e3ef2ee36b4428a8a7e9e66ffc53a4ea4cdebff29725L249

Should
```py
assert wallet, "Wallet for balancecheck could not be fetched"
```
be?
```py
assert wallet and not wallet.deleted, "Wallet for balancecheck could not be fetched"
```

- do we rely on the decorator check only? what if the call is made from a scheduled task (in `tasks.py`)?

motorina0:
- here: https://github.com/lnbits/lnbits/blob/add_deleted_flag_wallet/lnbits/core/tasks.py#L105
```py
        wallet = await get_wallet(payment.wallet_id)
        if wallet:
            await send_payment_notification(wallet, payment)
```
 - shouldn't be checked for `wallet.deleted`?

lnbits/core/views/api.py
```diff
@@ -466,7 +466,7 @@ async def api_payment(payment_hash, X_Api_Key: Optional[str] = Header(None)):
     # We use X_Api_Key here because we want this call to work with and without keys
     # If a valid key is given, we also return the field "details", otherwise not
     wallet = await get_wallet_for_key(X_Api_Key) if isinstance(X_Api_Key, str) else None
-
+    wallet = wallet if wallet and not wallet.deleted else None
```
jackstar12, line 469: shouldn't all these checks for the deleted flag be moved into the crud methods? Or is there any case where you would want to retrieve a deleted wlalet?

PR #1829, opened 2023-07-20
Wallets: Catch errors during cleanup

callebtc:
Leads to ugly stack traces when the cleanup routine is called while `paid_invoice_stream` is still streaming.

Example:
image


dni:
can you actually only catch that specific error, `RuntimeError` and print its message as a warning? i think its more accurate  and we and we don't swallow any other error that could happen.

dni (approved):

jackstar12:
https://github.com/lnbits/lnbits/pull/1793 solves the same problem, but in a more clean manner

Issue #1820, opened 2023-07-11
[BUG] WebSocket for incoming/outgoing payments does not trigger on outgoing internal payments.


pieterjm:
**Describe the bug**
WebSocket for incoming/outgoing payments does not trigger on outgoing internal payments. So, when a payment is made from the wallet that is monitored to another wallet on the same LNbits instance, not message is sent on the WebSocket. 

A different, but related issue is a lack of payment details (like the amount) for outgoing payments. When a payment is made to an external wallet, a message is sent, but the payment dict does not include an amount. 

**To Reproduce**

1. Create two wallets (A, B) on the same LNbits server instance.
2. Create a WebSocket client that is connected to wallet A.
3. In wallet B, create an invoice.
4. In wallet A, pay the invoice.

No message is received on the WebSocket connection.

Steps to reproduce a lack of payment information
1. Create a wallet on an LNbits server
2. Create a WebSocket client that is connected to the wallet
3. On an external wallet (I used Phoenix), create an invoice
4. Scan the invoice in the LNbits wallet and pay the invoice

A message similar to the one below is received on the WebSocket connection. An amount is missing. 

{'wallet_balance': 12345, 'payment': {'ok': True, 'checking_id': '0..0', 'fee_msat': 1007, 'preimage': '45...2f', 'error_message': None}}

**Expected behavior**
A JSON message should be received on the WebSocket. 

- LNbits version: 0.9.10, commit version  83a04d8f74053b4bb3e6d01b3e33b2467faa5c5f
- Database: sqlite

**Additional context**

A small python script to test messages on the WebSocket connection.

#!/usr/bin/env python

import asyncio
from websockets.sync.client import connect

def test_ws():
    with connect("wss://{lnbits server}/api/v1/ws/{wallet id}") as websocket:
        while True:
            message = websocket.recv()
            print(f"Received: {message}")

test_ws()

PR #1831, opened 2023-07-23
websocket internal payment notifications

jackstar12:
fixes: https://github.com/lnbits/lnbits/issues/1820

I introduced a `send_payment_notification` service because this logic is now needed in three seperate places in order to avoid inconsistent messages. Now all messages inclue the whole payment information ( `payment.dict`) which could be considered a breaking change (outgoing had the structure of `PaymentResponse`.

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -44,7 +44,7 @@ from .crud import (
     update_super_user,
 )
 from .helpers import to_valid_user_id
-from .models import Payment
+from .models import Payment, Wallet
 
 
 class PaymentFailure(Exception):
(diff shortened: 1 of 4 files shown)


jackstar12:
I also considered writing tests, but `httpx` doesnt include websocket functionality and the `websockets` library doesnt seem to have a way to connect to an asgi app directly. The [fastapi testclient](https://fastapi.tiangolo.com/advanced/testing-websockets/) could be used tough

arcbtc (approved):
Thank you!

lnbits/core/services.py
```diff
@@ -248,16 +249,11 @@ class PaymentKwargs(TypedDict):
...
-                    await websocketUpdater(
-                        wallet_id,
-                        json.dumps(
-                            {
-                                "wallet_balance": wallet.balance or None,
-                                "payment": payment._asdict(),
-                            }
-                        ),
-                    )
+                updated = await get_wallet_payment(
```
motorina0, line 252: why is `get_wallet_payment()` call needed?

jackstar12: because the above `update_payment_details` arent reflected otherwise. The `new_payment` variable could be updated instead aswell

dni (approved):

motorina0 (approved):

callebtc (approved):

PR #1852, opened 2023-08-03
Fix csv export

jackstar12:
Continuation of https://github.com/lnbits/lnbits/pull/1756

Instead of creating a new endpoint for csv just fetch all payments and create the csv in the browser.
It also exports the fiat value and currency from https://github.com/lnbits/lnbits/pull/1789


arbadacarbaYK:
This is a huge win for merchants and also builders, thank you for caring !

PR #1855, opened 2023-08-04
Fix nix build and modules

HaosGames:
This fixes the out of date nix build configuration and modules. 

Some things that still need fixing - possibly at a later date:
- creation of `.super-user` file failes due to read only file system in the store. That means that you can't use the admin ui yet. 
- installation of extensions via the ui failes because the upgrades dir is also in the read-only store


HaosGames:
Actually I will need to leave those two points for another PR. At this point I cant figure out how to fix these.

dni:
is it still a wip? or is it mergeable?

nix/modules/lnbits-service.nix
```diff
@@ -79,22 +94,32 @@ in
 
     systemd.tmpfiles.rules = [
       "d ${cfg.stateDir}                            0700 ${cfg.user} ${cfg.group} - -"
+      "d ${cfg.extensionsDir}                        0700 ${cfg.user} ${cfg.group} - -"
```
prusnak, line 97: drop extra space:
```suggestion
      "d ${cfg.extensionsDir}                       0700 ${cfg.user} ${cfg.group} - -"
```

flake.nix
```diff
@@ -7,13 +7,22 @@
...
       supportedSystems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
       forSystems = systems: f:
         nixpkgs.lib.genAttrs systems
-        (system: f system (import nixpkgs { inherit system; overlays = [ poetry2nix.overlay self.overlays.default ]; }));
+        (system: f system (import nixpkgs { 
+          inherit system; 
+          overlays = [ poetry2nix.overlay self.overlays.default ]; 
+          # This is needed for the dependency "cryptography" of cashu 
+          # which is built with python36 which is using openssl_1_1 injected by peotry2nix's defaultOverrides
+          config.permittedInsecurePackages = [
```
prusnak, line 22: Let's not deal with this here, but let's fix cashu upstream and then use the fixed version of the upstream.

flake.nix
```diff
@@ -34,7 +43,22 @@
...
+          (self: super: {
+            types-mock = super.types-mock.overridePythonAttrs
+            (
+              old: {
+                buildInputs = (old.buildInputs or [ ]) ++ [ super.setuptools ];
+              }
+            );
+            cashu = super.cashu.overridePythonAttrs (
+              old: {
+                nativeBuildInputs = (old.nativeBuildInputs or [ ]) ++ [ (import legacy-nixpkgs {inherit system;}).python3Packages.setuptools ];
```
prusnak, line 57: Let's not deal with this here, but let's fix cashu upstream and then use the fixed version of the upstream.

flake.nix
```diff
@@ -34,7 +43,22 @@
...
+              }
+            );
+          });
+          nativeBuildInputs = [(import legacy-nixpkgs {inherit system;}).python3Packages.setuptools];
```
prusnak, line 61: Please don't. If there is an issue, let's fix it, but importing old version is just nasty.

prusnak (changes requested):

prusnak:
This PR is a total mess because the main branch is no longer used for development but for a stable release.

Also https://github.com/lnbits/lnbits/pull/1979 which fixes the nix stuff has been merged.

I am closing this PR.

If something still needs fixing, please open a new PR.

PR #1859, opened 2023-08-07
[CHORE] cln.py pyln-client is not optional

dni:
some time ago we added pyln do main dependencies, its need for think like, saas or docker. so this import was dead code


motorina0 (approved):
LGTM

talvasconcelos (approved):

PR #1861, opened 2023-08-07
[BUG] dont install default extensions for tests

dni:
this is important for local testing. else you will install those extensions if you have them inside env.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,6 +18,9 @@ from tests.helpers import (
     get_real_invoice,
 )
 
+# dont install extensions for tests
+settings.lnbits_extensions_default_install = []
+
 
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -18,6 +18,9 @@
     get_real_invoice,
 )
 
+# dont install extensions for tests
+settings.lnbits_extensions_default_install = []
```
motorina0, line 22: - but this can be set in the `.env` file for tests, right?

dni: yes, but i think this is cleaner, because this actually is true for all tests

motorina0 (approved):

talvasconcelos (approved):

PR #1863, opened 2023-08-07
[REFACTOR] proper create_invoice responses for descriptions

dni:
return a exception per error it is description hash or unhashed_description

### DIFF
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -198,7 +198,7 @@ async def api_payments_create_invoice(data: CreateInvoice, wallet: Wallet):
             except ValueError:
                 raise HTTPException(
                     status_code=HTTPStatus.BAD_REQUEST,
-                    detail="'description_hash' must be a valid hex strings",
+                    detail="'description_hash' must be a valid hex string",
                 )
         if data.unhashed_description:
             try:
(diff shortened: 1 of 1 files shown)


lnbits/core/views/api.py
```diff
@@ -196,25 +196,27 @@ class CreateInvoiceData(BaseModel):
 
 
 async def api_payments_create_invoice(data: CreateInvoiceData, wallet: Wallet):
+    description_hash = None
```
motorina0, line 199: - previously the default values were `b""` and not `None`

callebtc: Agree with @motorina0, let's not switcherydoo the types like Python programmers

dni: done

lnbits/core/views/api.py
```diff
@@ -187,38 +187,37 @@ async def api_payments_paginated(
...
+                    status_code=HTTPStatus.BAD_REQUEST,
+                    detail="'description_hash' must be a valid hex strings",
+                )
+        if data.unhashed_description:
+            try:
+                unhashed_description = bytes.fromhex(data.unhashed_description)
+            except ValueError:
+                raise HTTPException(
+                    status_code=HTTPStatus.BAD_REQUEST,
+                    detail="'unhashed_description' must be a valid hex strings",
```
michael1011, line 209:
```suggestion
                    detail="'unhashed_description' must be a valid hex string",
```

arcbtc (approved):

jackstar12 (approved):

PR #1870, opened 2023-08-08
[REFACTOR] remove db.connect from create_invoice

dni:
passing through the connection is not necessary for create_invoice

### DIFF
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -236,7 +236,7 @@ async def api_payments_create_invoice(data: CreateInvoiceData, wallet: Wallet):
             webhook=data.webhook,
             internal=data.internal,
         )
-    except InvoiceFailure as exc:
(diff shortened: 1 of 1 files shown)


lnbits/core/views/api.py
```diff
@@ -236,10 +236,11 @@ async def api_payments_create_invoice(data: CreateInvoiceData, wallet: Wallet):
             webhook=data.webhook,
             internal=data.internal,
         )
-    except InvoiceFailure as e:
-        raise HTTPException(status_code=520, detail=str(e))
-    except Exception as exc:
-        raise exc
+    except InvoiceFailure as exc:
```
motorina0, line 239:
```suggestion
    except Exception as exc:
```

Otherwise all other exceptions will leak to the client.

motorina0 (approved):

PR #1871, opened 2023-08-08
[CHORE] E722 bare exception fix

dni:
remove all bare exceptions from codebase and change it in `.flake8`

Bare except will catch exceptions you almost certainly don't want to catch, including KeyboardInterrupt (the user hitting Ctrl+C) and Python-raised errors like SystemExit

If you don't have a specific exception you're expecting, at least except Exception, which is the base type for all "Regular" exceptions.

https://stackoverflow.com/questions/54948548/what-is-wrong-with-using-a-bare-except

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -133,7 +133,7 @@ async def check_funding_source() -> None:
                 f"The backend for {WALLET.__class__.__name__} isn't working properly: '{error_message}'",
                 RuntimeWarning,
             )
-        except:
+        except Exception:
             pass
 
         if settings.lnbits_admin_ui and retry_counter == timeout:
--- a/lnbits/decorators.py
+++ b/lnbits/decorators.py
@@ -173,7 +173,7 @@ async def get_key_type(
                 pass
             else:
                 raise
-        except:
+        except Exception:
             raise
     raise HTTPException(
         status_code=HTTPStatus.NOT_FOUND, detail="Wallet does not exist."
--- a/.flake8
+++ b/.flake8
@@ -10,7 +10,5 @@ ignore =
     W503,
     # F821: undefined name - should be addressed in future PR
     F821,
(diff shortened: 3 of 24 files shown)


motorina0:
@dni please check if there are other large PRs pending (c-lightning rest?). This PR might cause a lot of conflicts

dni:
pretty hard to check, before it merged, but i can easily solve all conflicts in PRs.

lnbits/decorators.py
```diff
@@ -173,7 +173,7 @@ async def get_key_type(
                 pass
             else:
                 raise
-        except:
+        except Exception:
             raise
```
motorina0, line 177: - isn't this `except` redundant?

motorina0 (approved):
LGTM. 
See the two open questions.

lnbits/decorators.py
dni: Bare except will catch exceptions you almost certainly don't want to catch, including KeyboardInterrupt (the user hitting Ctrl+C) and Python-raised errors like SystemExit

If you don't have a specific exception you're expecting, at least except Exception, which is the base type for all "Regular" exceptions.

https://stackoverflow.com/questions/54948548/what-is-wrong-with-using-a-bare-except

lnbits/app.py
```diff
@@ -133,7 +133,7 @@ def signal_handler(signal, frame):
                 f"The backend for {WALLET.__class__.__name__} isn't working properly: '{error_message}'",
                 RuntimeWarning,
             )
-        except:
+        except Exception:
```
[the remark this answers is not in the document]
dni: check descripition

lnbits/decorators.py
dni: looking at it the whole thing is suspicious. i would defer it to another PR

talvasconcelos (approved):

PR #1874, opened 2023-08-09
[CHORE] flake8 issues E402, E721 and F821

dni:
for more details look into commit messages.

could not fix all E402 issues, the last one needed following refactor.
https://github.com/lnbits/lnbits/pull/1876

E721 only popped up for python 3.9 not 3.10

### DIFF
--- a/lnbits/wallets/lnpay.py
+++ b/lnbits/wallets/lnpay.py
@@ -1,11 +1,8 @@
 import asyncio
 import hashlib
-import json
-from http import HTTPStatus
 from typing import AsyncGenerator, Dict, Optional
 
 import httpx
-from fastapi import HTTPException
 from loguru import logger
 
 from lnbits.settings import settings
@@ -65,7 +62,7 @@ class LNPayWallet(Wallet):
         memo: Optional[str] = None,
         description_hash: Optional[bytes] = None,
         unhashed_description: Optional[bytes] = None,
-        **kwargs,
+        **_,
     ) -> InvoiceResponse:
         data: Dict = {"num_satoshis": f"{amount}"}
         if description_hash:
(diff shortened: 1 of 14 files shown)


lnbits/wallets/lnpay.py
```diff
@@ -139,25 +136,26 @@ async def paid_invoices_stream(self) -> AsyncGenerator[str, None]:
...
-            raise HTTPException(status_code=HTTPStatus.NO_CONTENT)
-
-        lntx_id = data["data"]["wtx"]["lnTx"]["id"]
-        r = await self.client.get(f"/lntx/{lntx_id}?fields=settled")
-        data = r.json()
-        if data["settled"]:
-            await self.queue.put(lntx_id)
-
-        raise HTTPException(status_code=HTTPStatus.NO_CONTENT)
+        # text: str = await request.get_data()
```
motorina0, line 143: dead code?

lnbits/wallets/opennode.py
```diff
@@ -136,18 +133,20 @@ async def paid_invoices_stream(self) -> AsyncGenerator[str, None]:
...
-        charge_id = data["id"]
-        x = hmac.new(self.auth["Authorization"].encode("ascii"), digestmod="sha256")
-        x.update(charge_id.encode("ascii"))
-        if x.hexdigest() != data["hashed_order"]:
-            logger.error("invalid webhook, not from opennode")
-            raise HTTPException(status_code=HTTPStatus.NO_CONTENT)
-
-        await self.queue.put(charge_id)
-        raise HTTPException(status_code=HTTPStatus.NO_CONTENT)
+        # data = await request.form  # type: ignore
```
motorina0, line 140: - why disable code?
- will it later be re-enabled? if not, then it should be remove.

motorina0:
- why is `isinstance(something, str)` preferred to `type(something) == str` ?

jackstar12:
> - why is `isinstance(something, str)` preferred to `type(something) == str` ? 

If something wasnt an actual `str` but maybe a `claas MyStr(str)` the latter would return False while the `isinstance` check returns True. This can lead to unexpected bugs

lnbits/wallets/lnpay.py
dni: lets open a issue for those 2, this code does not work currently thats why i commented it out, at some point some need to reimplement it i guess.

lnbits/wallets/opennode.py
dni: same here

talvasconcelos (approved):
apart from the commented out code vs deleting, it looks fine

PR #1876, opened 2023-08-09
[REFACTOR] core/__init__ to not have circular import issues

dni:
This fixes `E402: module level import not at top of file` and required a bit of a refactor for core imports.

now every core.view creates its own `APIRouter` which is imported into `core/__init__.py` where its initialised inside app via `init_core_routers(app)`

tested backwards compatibility for extensions and they at least not crash with import errors.

### DIFF
--- a/lnbits/core/db.py
+++ b/lnbits/core/db.py
@@ -1,0 +1,5 @@
+from lnbits.core.models import CoreAppExtra
+from lnbits.db import Database
+
+db = Database("database")
+core_app_extra: CoreAppExtra = CoreAppExtra()
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -33,14 +33,11 @@ from lnbits.utils.cache import cache
 from lnbits.wallets import get_wallet_class, set_wallet_class
 
 from .commands import db_versions, load_disabled_extension_list, migrate_databases
-from .core import (
-    add_installed_extension,
-    core_app,
-    core_app_extra,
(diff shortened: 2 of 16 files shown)


lnbits/core/db.py
```diff
@@ -0,0 +1,5 @@
+from lnbits.core.models import CoreAppExtra
+from lnbits.db import Database
+
+db = Database("database")
+core_app_extra: CoreAppExtra = CoreAppExtra()
```
jackstar12, line 5: `core_app_extra` just doesnt really fit here. Not really sure where it might be better tough. Maybe the two methods behind it could somehow become helpers instead?

jackstar12 (approved):

lnbits/core/db.py
motorina0: - not sure if this file should exist

motorina0:
> required a bit of a refactor for core imports
- why a reordering of imports requires a refactor?

jackstar12:
> > required a bit of a refactor for core imports
> 
> * why a reordering of imports requires a refactor?

because it introduced circular dependency issues. The previous `core/__init__.py` declared its routers before importing the other apis because those import the core router aswell and declare their routes there directly.

lnbits/core/db.py
dni: without that file, you would have a circular import again

arcbtc (approved):

Issue #1111, opened 2022-11-07
[Feature request] signet support


openoms:
signet has many advantages, but most of all being more reliable than testnet as there are no block-storms and crazy reorgs.
See collected resources here: https://github.com/GaloyMoney/charts/issues/1123

I presume LNbits has invoice parsing issues like everything else
due to the extra letter in the signet LNTBS... invoice.
Tested and able to generate invoices and receive signet sats, but can send with this error:

![image](https://user-images.githubusercontent.com/43343391/200379345-e875dfd2-9347-4dd3-9f90-1cab6093ea5d.png)

example signet invoice:
```
lntbs2220p1p3kjsmtsp5pvxxvnpdv2hrdzfcflwsy09kh9qe7nmg5ynh2pu2wlh2ksnucvfqpp5ptgl7q65zuqh0llq28w38axcgyd857z9wa2p8tuyy2redhmk8qkqdqdwpshjar0gdxyuxqyjw5qcqp2rzjqttu3n0dv9zfzaravqdvkap5zcwpk0n8dv3e3r7xugl5zykdjfekyqwrpuqqqqgqqyqqqqlgqqqqqqgq9q9qyysgq04xhvgwcu7wty007dna2gg9jekzdy54486p5u7ty8z3327tvx9zx7r20tepq536kktcmae94zvdsy4d0jmuuqeq7t38y4xrnutrdqecpgwc3ll
```

see the related discussion in bitcoinjs: https://github.com/bitcoinjs/bitcoinjs-lib/issues/1820

Feel free to test on this temporary instance:
https://signet.diynodes.com/wallet?usr=7a9d00aa89f44af1b1808244d8caafef&wal=ce0e6e5827224618aa75a56ac73f2ef2

bitkarrot:
@rsafier,  @RandyMcMillan 

https://github.com/PLEBNET-PLAYGROUND/plebnet-playground-docker

rsafier:
> 

yeah we started and did seem to have issues, but based on bitcoinjs library above it won't work until that is corrected, same issues almost all apps seem to have is the underlying libraries they use tend to have missing validation code for signet.

looks like that was probably were we got stuck based on PR notes:
`
 File "/app/lnbits/./lnbits/core/views/api.py", line 153, in api_payments_create_invoice
    payment_hash, payment_request = await create_invoice(
  File "/app/lnbits/./lnbits/core/services.py", line 55, in create_invoice
[...]

PR #1881, opened 2023-08-16
[FEATURE] Adding coverage for 'lntbs' signet prefix

enaples:
This PR should add coverage for 'lntbs' prefix of signet invoices. Hence, should resolve the issue #1111 which I think can be closed.


dni:
can you do `make bundle` aswell and commit that? else i can push it to your branch aswell

dni (approved):

PR #1886, opened 2023-08-16
[REFACTOR] grpcio and protofbuf for lndgrpc is not optional

dni:
also removed dead code def get_ssl_context

### DIFF
--- a/lnbits/wallets/lndgrpc.py
+++ b/lnbits/wallets/lndgrpc.py
@@ -4,8 +4,13 @@ import hashlib
 from os import environ
 from typing import AsyncGenerator, Dict, Optional
 
+import grpc
 from loguru import logger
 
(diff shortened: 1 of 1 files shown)


callebtc:
I think you're missing the point of this code. If you don't have the libs installed, lnbits won't start up. We didn't assume everyone having to have lndgrpc installed on their machine.

motorina0 (approved):
LGTM

- jobs re-ran fine here: https://github.com/lnbits/lnbits/actions/runs/6023489337?pr=1886

motorina0:
> I think you're missing the point of this code. If you don't have the libs installed, lnbits won't start up. We didn't assume everyone having to have lndgrpc installed on their machine.

1. how is `lndgrpc` different from any other installed lib
2. generally speaking, how is it decided that a lib might not be present/installed (and use conditional imports)

jackstar12 (approved):

PR #1899, opened 2023-08-24
[CHORE] cleanup cache and exception on `make test`

dni:
moved the invalidate_forever task to proper location and use `create_permanent_task`. and there was a Task never awaited Exception when running tests


tests/core/test_cache.py
```diff
@@ -2,48 +2,37 @@
...
-from tests.conftest import pytest_asyncio
-
-
-@pytest_asyncio.fixture(scope="session")
-async def cache():
-    cache = Cache()
-
-    task = asyncio.create_task(cache.invalidate_forever(interval=0.1))
-    yield cache
-    task.cancel()
```
jackstar12, line 15: I wouldn't remove this fixture, the invalidation needs go faster so that it actually runs in the expiration test - this doesn't show because the expiration is just done by the `get` call now, but the test can be adjusted to show that

dni: thats the only way how i could get rid of the task canceled error.
can you do that?

dni: weirdly the tests passed locally for me. maybe we a function with `cache.interval(0.1)` or something like that to dynamically set the cache interval?

lnbits/utils/cache.py
```diff
@@ -17,9 +17,13 @@ class Cache:
...
     """
 
-    def __init__(self) -> None:
+    def __init__(self, interval: float = 10) -> None:
+        self.interval = interval
         self._values: dict[Any, Cached] = {}
 
+    def set_interval(self, interval: float):
+        self.interval = interval
+
```
jackstar12, line 26: I think this one isnt needed

jackstar12 (approved):

arcbtc (approved):

PR #1903, opened 2023-08-25
[FEAT] improve update_admin_settings

dni:
while working on the push notification pr i found it very hard just to update 2 settings inside the db, so i improved upon update_admin_settings. now you just need to provide a dict with key/values you want to update inside db.

also debugging the endpoints for update_settings i found despite the type of `EditableSettings` fastapi did in fact pass a dict.

import

### DIFF
--- a/lnbits/core/views/admin_api.py
+++ b/lnbits/core/views/admin_api.py
@@ -8,7 +8,6 @@ from urllib.parse import urlparse
 
 from fastapi import Depends
 from fastapi.responses import FileResponse
-from pydantic import Extra
 from starlette.exceptions import HTTPException
 
 from lnbits.core.crud import get_wallet
@@ -20,7 +19,7 @@ from lnbits.core.services import (
 )
 from lnbits.decorators import check_admin, check_super_user
 from lnbits.server import server_restart
-from lnbits.settings import AdminSettings, EditableSettings, settings
+from lnbits.settings import AdminSettings, UpdateSettings, settings
 
 from .. import core_app, core_app_extra
 from ..crud import delete_admin_settings, get_admin_settings, update_admin_settings
@@ -53,11 +52,6 @@ async def api_get_settings(
 ) -> Optional[AdminSettings]:
     admin_settings = await get_admin_settings(user.super_user)
     return admin_settings
-
-
-class UpdateSettings(EditableSettings):
-    class Config:
-        extra = Extra.forbid
 
 
 @core_app.put(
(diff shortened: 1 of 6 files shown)


motorina0:
With this change any property is allowed in the DB. Eg:
```py
update_admin_settings({"foo": "bar"})
```

I see two options:
 1. inside `update_admin_settings` also check that all keys in `data: dict` are part of `EditableSettings`
 2. leave `update_admin_settings` as is and add a new function `update_admin_setting(key: str, value: str)`
    - check here that the key is valid
    - has the disadvantage that it might have to be called multiple times

jackstar12 (changes requested):

jackstar12:
I have changed it so that the function still takes in `EditableSettings` but also has the partial update through `exclude_unset`.

jackstar12:
argh, `exclude_unset` doesnt seem to work here because `EditableSettings` is `BaseSettings` so the values are always filled in from env

tests/conftest.py
```diff
@@ -10,15 +10,16 @@
...
 from lnbits.core.models import CreateInvoice
 from lnbits.core.services import update_wallet_balance
 from lnbits.core.views.api import api_payments_create_invoice
 from lnbits.db import Database
 from lnbits.settings import settings
 from tests.helpers import get_hold_invoice, get_random_invoice_data, get_real_invoice
 
-# dont install extensions for tests
+# override settings for tests
+settings.lnbits_admin_ui = True
```
jackstar12, line 22: i guess this could lead to uncomfortable situations in local testing, where the tests now use values from your db while you'd like the tests to use your env

dni: no, i think that i compleltly fine, when using `pytest` without `make test` we are lacking those `.env` we set in the make command. so i actually think conftest is the place to put them!

lnbits/core/views/admin_api.py
```diff
@@ -54,13 +55,16 @@ async def api_get_settings(
     return admin_settings
 
 
+class UpdateSettings(EditableSettings):
```
motorina0, line 58: - is this needed now, extra fields will be ignored, right?

jackstar12: its intentionally, so that if you try to send invalid settings it will error instead of ignoring. I didn't want that for `EditableSettings` though because it can be annoying in development when you have some settings from another branch in your db and now you get errors on startup because they dont exist in current branch

motorina0: ok, this class should declared be next to `EditableSettings` (not in the api file)

motorina0 (approved):

jackstar12 (approved):

PR #1907, opened 2023-08-28
[FEAT] release / docker ci pipeline

dni:
the goal of this PR is to run these action on a new tag:
- create a github release and autogenerate release notes on tagging
- release a dockerimage on dockerhub for that tag and latest
- can manually run docker publish on custom tag

## BREAKING
in the future we will use `docker pull lnbits/lnbits` instead of `lnbitsdocker/lnbits-legend`


motorina0 (approved):

.github/workflows/docker.yml
```diff
@@ -2,19 +2,52 @@ name: docker
...
+        uses: docker/build-push-action@v4
+        with:
+            context: .
+            push: true
+            tags: ${{ secrets.DOCKER_USERNAME }}/lnbits:${{ inputs.tag }}
+            platforms: [ linux/amd64, linux/arm64 ]
+            cache-from: type=local,src=/tmp/.buildx-cache
+            cache-to: type=local,dest=/tmp/.buildx-cache
+
       - name: Build and push
```
motorina0, line 45:
```suggestion
      - name: Build and push latest tag
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
+name: release
+
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
```
motorina0, line 6: are we still planning to support tags of the form `v0.11.rc0`?

motorina0:
The image on https://hub.docker.com/u/lnbits was `Updated 4 moths ago`.

Can we get the flow to run for this branch?
 - rebase dev
 - relax tag regex
 - create trial tag on this branch. Ex: `dec-20-2023`?
 - revert tag relax regex

.github/workflows/release.yml
dni: we could consider switching to `v` as it is the default on github

v0.0.0
or 0.0.0

dni:
> The image on https://hub.docker.com/u/lnbits was `Updated 4 moths ago`.
> 
> Can we get the flow to run for this branch?
> 
>     * rebase dev
> 
>     * relax tag regex
> 
>     * create trial tag on this branch. Ex: `dec-20-2023`?
> 
>     * revert tag relax regex

i dont really think that adds much, i would just switch the docker with the `0.12.0` release

if we want we could manually run older tags via github actions GUI and push them like that.

.github/workflows/release.yml
motorina0: I meant the `rc0` at the end. The regex only supports digits.

motorina0:
> i dont really think that adds much, i would just switch the docker with the `0.12.0` release
> 
> if we want we could manually run older tags via github actions GUI and push them like that.

Its good to test before we merge.

.github/workflows/release.yml
dni: currently we don't have it in on-tag. we could add it. but i think rc could be an expection where we do it manually in the GUI

arcbtc (approved):

PR #1914, opened 2023-09-06
[BUG] self payments for the same wallet

dni:
make it possible to pay to yourself on fakewallet.

adds failing test
and a commit fixing it


callebtc:
One problem in this line in `crud.py` was that we were checking for the existence of a previous payment using the `payment_hash` and not the `checking_id`. The `checking_id` would not collide for an internal payment, because we add `internal_` to it. 

The second problem is that we looked up payments only from the active wallet. We want to avoid having the same payment twice in the entire db though.

arcbtc (approved):

PR #1917, opened 2023-09-07
[FEAT] move out extensions from /lnbits

dni:
wip

we would need to remove the "app" StaticFiles inside all extensions to make this work, atm


jackstar12:
> we would need to remove the "app" StaticFiles inside all extensions to make this work, atm

couldnt we just check if the static files path starts with `lnbits/extensions`, strip it away and replace it with the right one for now?

dni:
https://github.com/lnbits/lnbits/pull/1940

PR #1929, opened 2023-09-11
[TEST] use test data as mockdata

jackstar12:
the `mock_data.zip` is and the migration workflow seems to fail randomly with mismatching db versions.

generating the mock data through a script makes sure the db is up to date


dni (approved):

lnbits/commands.py
```diff
@@ -104,6 +113,27 @@ async def db_versions():
         return await get_dbversions(conn)
 
 
+@db.command("mock")
+def database_mock():
```
motorina0, line 117: - still want to keep this?

Makefile
```diff
@@ -48,18 +48,16 @@ test-real-wallet:
...
 
 test-migration:
-	rm -rf ./migration-data
-	mkdir -p ./migration-data
-	unzip tests/data/mock_data.zip -d ./migration-data
+	make test
 	HOST=0.0.0.0 \
 	PORT=5002 \
-	LNBITS_DATA_FOLDER="./migration-data" \
+	LNBITS_DATA_FOLDER="./tests/data" \
```
motorina0, line 54: - are we sure there is no clean-up hook that cleans up the data?

lnbits/commands.py
jackstar12: i dont think so

jackstar12:
removed the script, https://github.com/lnbits/lnbits/pull/1929/commits/0c69332fe77dbcb2009c18965850a928304202d6 could be its own pr now, but I think we can leave it here

arcbtc (approved):

PR #1932, opened 2023-09-11
[FEAT] cleanup GET /wallet endpoint, add wallet api routes

dni:
this removes the functionalitiy to create accounts and wallets via the GET /wallet endpoint in generic.py

it add endpoints inside the api.py for it and the frontend is modified to use the api endpoints

this also simplifies for the `feat/login` for the route.

### DIFF
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -77,7 +77,7 @@ async def get_user(user_id: str, conn: Optional[Connection] = None) -> Optional[
                 SELECT balance FROM balances WHERE wallet = wallets.id
             ), 0) AS balance_msat
             FROM wallets
-            WHERE "user" = ?
+            WHERE "user" = ? and wallets.deleted = false
             """,
             (user_id,),
         )
--- a/lnbits/core/static/js/wallet.js
+++ b/lnbits/core/static/js/wallet.js
@@ -702,7 +702,7 @@ new Vue({
 
       LNbits.api
         .authLnurl(this.g.wallet, this.parse.lnurlauth.callback)
-        .then(response => {
+        .then(_ => {
           dismissAuthMsg()
           this.$q.notify({
             message: `Authentication successful.`,
@@ -728,27 +728,34 @@ new Vue({
     updateWallet: function (data) {
       LNbits.api
         .request('PATCH', '/api/v1/wallet', this.g.wallet.adminkey, data)
-        .then(res => {
+        .then(_ => {
           this.$q.notify({
             message: `Wallet updated.`,
             type: 'positive',
             timeout: 3500
           })
-          LNbits.href.updateWallet(
-            res.data.name,
-            this.user.id,
-            this.g.wallet.id
-          )
         })
         .catch(err => {
           LNbits.utils.notifyApiError(err)
         })
     },
-    deleteWallet: function (walletId, user) {
+    deleteWallet: function () {
       LNbits.utils
         .confirmDialog('Are you sure you want to delete this wallet?')
         .onOk(() => {
-          LNbits.href.deleteWallet(walletId, user)
+          LNbits.api
+            .deleteWallet(this.g.wallet)
+            .then(_ => {
+              this.$q.notify({
+                timeout: 3000,
+                message: `Wallet deleted!`,
+                spinner: true
+              })
+            })
+            .catch(err => {
+              this.paymentsTable.loading = false
+              LNbits.utils.notifyApiError(err)
+            })
         })
     },
     fetchPayments: function (props) {
--- a/lnbits/core/static/js/wallet.js
+++ b/lnbits/core/static/js/wallet.js
@@ -734,6 +734,7 @@ new Vue({
             type: 'positive',
             timeout: 3500
           })
+          window.location.reload()
         })
         .catch(err => {
           LNbits.utils.notifyApiError(err)
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -152,7 +152,7 @@ async def api_update_wallet(
     return await update_wallet(wallet.wallet.id, name, currency)
 
 
-@api_router.delete("/api/v1/wallet", response_model=Wallet)
+@api_router.delete("/api/v1/wallet")
 async def api_delete_wallet(
     wallet: WalletTypeInfo = Depends(require_admin_key),
 ) -> None:
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -36,6 +36,7 @@ from lnbits.core.models import (
     CreateInvoice,
     CreateLnurl,
     CreateLnurlAuth,
+    CreateWallet,
     CreateWebPushSubscription,
     DecodePayment,
     Payment,
@@ -75,11 +76,14 @@ from lnbits.utils.exchange_rates import (
 from ..crud import (
     DateTrunc,
     add_installed_extension,
+    create_account,
     create_tinyurl,
+    create_wallet,
     create_webpush_subscription,
     delete_dbversion,
     delete_installed_extension,
     delete_tinyurl,
+    delete_wallet,
     delete_webpush_subscription,
     drop_extension_db,
     get_dbversions,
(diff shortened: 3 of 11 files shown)


lnbits/static/js/base.js
```diff
@@ -64,9 +64,29 @@ window.LNbits = {
         callback
       })
     },
+    createAccount: function (name) {
+      return this.request('post', '/api/v1/wallet', null, {
```
motorina0, line 68: - the function is called `createAccount` but it creates a wallet

dni: you are right! that is abug

lnbits/core/views/generic.py
```diff
@@ -162,77 +161,46 @@ async def extensions_install(
...
-    description="""
-just **wallet_name**: create a new user, then create a new wallet
-                      for user with wallet_name
-just **user_id**: return the first user wallet or create one if none found
-                  (with default wallet_name)
-**user_id** and **wallet_name**: create a new wallet for user with wallet_name
-**user_id** and **wallet_id**: return that wallet if user is the owner
-nothing: create everything
-""",
+    description="show wallet page",
```
motorina0, line 164: I hope people are not using this HTML serving endpoint as an REST API one

motorina0: Maybe a similar POST endpoint should exist

lnbits/core/static/js/wallet.js
```diff
@@ -723,27 +723,34 @@ new Vue({
...
       LNbits.api
         .request('PATCH', '/api/v1/wallet', this.g.wallet.adminkey, data)
-        .then(res => {
+        .then(_ => {
           this.$q.notify({
             message: `Wallet updated.`,
             type: 'positive',
             timeout: 3500
           })
-          LNbits.href.updateWallet(
```
motorina0, line 732: - this used to work as a redirect `window.location.href`. Is it no longer needed?

lnbits/static/js/base.js
```diff
@@ -64,9 +64,29 @@ window.LNbits = {
...
     },
+    createAccount: function (name) {
+      return this.request('post', '/api/v1/account', null, {
+        name: name
+      })
+    },
     getWallet: function (wallet) {
       return this.request('get', '/api/v1/wallet', wallet.inkey)
     },
+    createWallet: function (name, wallet) {
```
motorina0, line 75: - param order: `(wallet, name)` feels more natural

tests/core/views/test_api.py
```diff
@@ -25,10 +25,40 @@
...
-    response = await client.get("/")
+async def test_create_account(client):
+    response = await client.post("/api/v1/account", json={"name": "test"})
+    assert response.status_code == 200
+    result = response.json()
+    assert "name" in result
+    assert result["name"] == "test"
+    assert "balance_msat" in result
+    assert "id" in result
+    assert "user" in result
```
motorina0, line 38: - add test for `patch` (include currency update)

motorina0:
Please `make bundle` and push

motorina0:
- delete wallet does not refresh UI

https://github.com/lnbits/lnbits/assets/2951406/2906e69c-7766-4f2d-a6ea-889d76f59068

arcbtc (approved):
LFGM. Well spotted!

tests/core/views/test_api.py
dni: TODO: add test for patch

lnbits/core/views/generic.py
dni: we have to annouce it a pbreaking change for `0.11.0`

tests/core/views/test_api.py
```diff
@@ -25,10 +25,40 @@
...
+    assert "balance_msat" in result
+    assert "id" in result
+    assert "adminkey" in result
+    response = await client.delete(
+        "/api/v1/wallet",
+        headers={
+            "X-Api-Key": result["adminkey"],
+            "Content-type": "application/json",
+        },
+    )
```
callebtc, line 61: Maybe the test can try to create an invoice that is expected to fail from this deleted wallet.

callebtc (approved):

lnbits/core/static/js/wallet.js
dni: it is basically just a reload to change the name, i think its unecessary

lnbits/static/js/base.js
```diff
@@ -64,9 +64,29 @@ window.LNbits = {
...
+      return this.request('post', '/api/v1/wallet', wallet.adminkey, {
+        name: name
+      }).then(res => {
+        window.location = '/wallet?usr=' + res.data.user + '&wal=' + res.data.id
+      })
+    },
+    updateWallet: function (name, wallet) {
```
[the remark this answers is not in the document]
dni: there is a test for that.

screenshot-1694524840

motorina0:
**Steps**:
 - delete all wallets of a user

**Expected Result**:
 - a new default wallet is created for the user
 - the user is still valid and can see its extensions and data

**Actual Result**:
 - the user can no longer access the wallet page (or any other page)
 - even if the user has no wallet it might still have some important data in some of the extensions

image

dni:
we could make the `?wal` optional and redirect to the first wallet. do you think its a good idea? and what if there are no wallets left, should we delete the user aswell?

lnbits/core/static/js/wallet.js
dni: fixed

lnbits/core/crud.py
```diff
@@ -77,7 +77,7 @@ async def get_user(user_id: str, conn: Optional[Connection] = None) -> Optional[
                 SELECT balance FROM balances WHERE wallet = wallets.id
             ), 0) AS balance_msat
             FROM wallets
-            WHERE "user" = ?
```
motorina0, line 80: - why was this done in this PR, it supposed to be a refactoring only
- this should be done in a different PR

lnbits/core/views/generic.py
```diff
@@ -166,18 +166,25 @@ async def extensions_install(
...
         return template_renderer().TemplateResponse(
             "error.html", {"request": request, "err": "User does not exist."}
         )
 
+    if not wal:
+        if len(user.wallets) == 0:
+            return template_renderer().TemplateResponse(
+                "error.html", {"request": request, "err": "User has no wallets."}
+            )
+        return RedirectResponse(url=f"/wallet?usr={user_id}&wal={user.wallets[0].id}")
```
motorina0, line 184: - `user.wallets[0]` can now be `None` as the `user always has a wallet` rule is no longer valid

lnbits/core/views/api.py
```diff
@@ -148,6 +152,30 @@ async def api_update_wallet(
     return await update_wallet(wallet.wallet.id, name, currency)
 
 
+@api_router.delete("/api/v1/wallet")
+async def api_delete_wallet(
+    wallet: WalletTypeInfo = Depends(require_admin_key),
+) -> None:
+    await delete_wallet(
```
motorina0, line 159: Suggestion:
- if user has zero wallets after the `delete_wallet` operation then create a default empty wallet

lnbits/core/views/generic.py
dni: i catch that about with `len(user.wallets) == 0`

lnbits/core/views/api.py
dni: i do not want to create stuff on `GET` or `DELETE` request

dni: i would rather delete the account when the last wallet is deleted

motorina0: > i do not want to create stuff on `GET` or `DELETE` request

- I understand for `GET` but for `DELETE` write operations are OK

lnbits/core/views/generic.py
```diff
@@ -162,77 +161,52 @@ async def extensions_install(
...
         )
 
-    logger.debug(
-        "Access "
-        f"{'user '+ user.id + ' ' if user else ''} "
-        f"{'wallet ' + wallet_name if wallet_name else ''}"
-    )
+    if not wal:
+        if len(user.wallets) == 0:
+            wallet = await create_wallet(user_id=user.id)
```
motorina0, line 181: - the wallet can **now** be deleted via the API also
- the default wallet should be checked and created at the deletion point
- this page might never be accessed if someone uses LNbits exclusively via the API

lnbits/core/crud.py
dni: else deleted wallets will show up on the wallet page. i directly uses the get_user crud to render it into jinja html template.

PR #1942, opened 2023-09-13
do not show deleted wallets

### LINKED ISSUE #1939: [BUG] deleted wallets still show up in frontend
screenshot-1694590265

wallet is still visible, when deleted, discoverd it here.

https://github.com/lnbits/lnbits/pull/1932/files#r1323243651

i think we have to put it into the CRUD aswell

talvasconcelos:
Adds the deleted flag to the CRUD for `get_user`, so it doesn't send deleted wallets to the UI.

Closes #1939

### DIFF
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -62,7 +62,9 @@ async def get_account(
 
 
 async def get_user(
-    user_id: str, conn: Optional[Connection] = None, deleted: Optional[bool] = False
+    user_id: str,
+    conn: Optional[Connection] = None,
+    include_deleted_wallets: Optional[bool] = False,
 ) -> Optional[User]:
     user = await (conn or db).fetchone(
         "SELECT id, email FROM accounts WHERE id = ?", (user_id,)
@@ -72,7 +74,7 @@ async def get_user(
         clauses = ["user = ?"]
         values: List[Any] = [user_id]
 
-        if not deleted:
+        if not include_deleted_wallets:
             clauses.append("deleted = ?")
             values.append(False)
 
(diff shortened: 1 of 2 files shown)


lnbits/core/crud.py
```diff
@@ -61,25 +61,35 @@ async def get_account(
...
+) -> Optional[User]:
     user = await (conn or db).fetchone(
         "SELECT id, email FROM accounts WHERE id = ?", (user_id,)
     )
 
     if user:
+        clauses = ["user = ?"]
+        values: List[Any] = [user_id]
+
+        if not deleted:
```
motorina0, line 75: - if `deleted == True` then the clause will not be added and both deleted and non-deleted wallets will be returned

lnbits/core/models.py
```diff
@@ -29,7 +29,7 @@ class Wallet(BaseModel):
     inkey: str
     currency: Optional[str]
     balance_msat: int
-    deleted: bool
+    deleted: Optional[bool] = False
```
motorina0, line 32: - why does it have to be `Optional`?

arcbtc (approved):

lnbits/core/crud.py
talvasconcelos: Doesn't that change the existing behaviour of the function? Wouldn't we have to track all calls to the function and change to `get_user(....., deleted = False)` ?

lnbits/core/models.py
talvasconcelos: Removed Optional. My DB was faulty...

PR #1943, opened 2023-09-18
Startup optimization: nonblocking expiry check

callebtc:


lnbits/tasks.py
```diff
@@ -149,37 +148,36 @@ async def check_pending_payments():
...
+            since=(int(time.time()) - 60 * 60 * 24 * 15),  # 15 days ago
+            complete=False,
+            pending=True,
+            outgoing=outgoing,
+            incoming=incoming,
+            exclude_uncheckable=True,
+        )
+        for payment in pending_payments:
+            await payment.check_status()
+            await asyncio.sleep(0.1)  # to avoid complete blocking
```
callebtc, line 166: Consider removing

dni (approved):

lnbits/tasks.py
```diff
@@ -149,37 +148,36 @@ async def check_pending_payments():
...
 
+        logger.info(
+            f"Task: pending check finished for {len(pending_payments)} payments"
+            f" (took {time.time() - start_time:0.3f} s)"
+        )
+        # we delete expired invoices once upon the first pending check
+        if incoming:
```
motorina0, line 173: - the `incoming` logic should either be outside this loop (before the while), or
- `check_pending_payments` should be split in two different functions:
   - `check_pending_outgoing_payments`
   - `check_pending_incoming_payments`

lnbits/tasks.py
```diff
@@ -149,37 +148,36 @@ async def check_pending_payments():
...
-            for payment in pending_payments:
-                await payment.check_status(conn=conn)
+        logger.info(
+            f"Task: checking all pending payments (incoming={incoming},"
+            f" outgoing={outgoing}) of last 15 days"
+        )
+        start_time = time.time()
+        pending_payments = await get_payments(
+            since=(int(time.time()) - 60 * 60 * 24 * 15),  # 15 days ago
+            complete=False,
```
motorina0, line 158: - `completed` is `False` by default

lnbits/tasks.py
```diff
@@ -149,37 +148,36 @@ async def check_pending_payments():
...
-            )
-            for payment in pending_payments:
-                await payment.check_status(conn=conn)
+        logger.info(
+            f"Task: checking all pending payments (incoming={incoming},"
+            f" outgoing={outgoing}) of last 15 days"
+        )
+        start_time = time.time()
+        pending_payments = await get_payments(
+            since=(int(time.time()) - 60 * 60 * 24 * 15),  # 15 days ago
```
motorina0, line 157: - consider optimizing the `since` value
```py
since=(int(time.time()) - 60 * 60 * 24 * 15), # 15 days ago
while True:
    # ...
        last_check_time = int(time.time())
        pending_payments = await get_payments(
            since=since,
            # ...
        )
        since = last_check_time
        # ...
     # ...
```

lnbits/tasks.py
callebtc: > * `completed` is `False` by default

[explicit > implicit](https://peps.python.org/pep-0020/)

lnbits/tasks.py
callebtc: To be safe, I won't be changing the logic of this loop in this PR, maybe some other time!

lnbits/tasks.py
callebtc: The 15 days are meant to cover the maximum expiry time for bolt11 invoices. Invoices checked once can be paid when checked another time so we can't change this.

arcbtc (approved):

PR #1957, opened 2023-09-22
[Hotfix] don't ask wallet key for GET on tinyurl

talvasconcelos:
Remove the request for wallet key on "GET" tinyurl

### DIFF
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -924,7 +924,12 @@ async def api_get_tinyurl(tinyurl_id: str):
     try:
         tinyurl = await get_tinyurl(tinyurl_id)
         if tinyurl:
-            return tinyurl
+            return {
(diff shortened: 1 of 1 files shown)


lnbits/core/views/api.py
```diff
@@ -929,14 +929,16 @@ async def api_create_tinyurl(
...
-            if tinyurl.wallet == wallet.wallet.inkey:
-                return tinyurl
+            return {
+                "id": tinyurl.id,
+                "url": tinyurl.url,
+                "endless": tinyurl.endless,
+                "time": tinyurl.time,
+            }
         raise HTTPException(
             status_code=HTTPStatus.FORBIDDEN, detail="Wrong key provided."
```
dni, line 943: that exceptions does not make a lot of sense, should be 404

arcbtc (approved):

PR #1940, opened 2023-09-13
Support custom path for installed extensions

motorina0:
### Summary
Long discussion here: https://github.com/lnbits/lnbits/issues/1579

By default extensions are installed in a hard-coded directory: `lnbits/extensions`.
This PR allow users to specify a custom directory where the extensions will be installed by specifying an [environment variable](https://github.com/lnbits/lnbits/blob/2023c32b9dce0481b865d524b5ebebc1f62433cb/.env.example#L45).

The reason for having a custom extension path is that some users do not want to perform write operations in their `lnbits` directory (maybe they are using Nix which installs programs to immutable locations).

**Required changes for extensions**:
 - this PR is backwards compatible. If the custom extensions path is not used then there is no need for extensions to be updated
 - if the custom extensions path is used then extensions must be updated (see [comment](https://github.com/lnbits/lnbits/pull/1940#issuecomment-1717352903) below)

**Config Steps**:
 - update the `.env` file by configuring `LNBITS_EXTENSIONS_PATH`. This location must allow WRITE operations.
```properties
LNBITS_EXTENSIONS_PATH="/Users/moto/Documents/GitHub/motorina0/lnbits/data/code"
```
 - installed extensions MUST be updated in order to support this custom path feature

### Testing
**Backwards compatible**
 - test that the PR is backwards compatible:
   - do not use the `LNBITS_EXTENSIONS_PATH`
   - install/uninstal/upgrade extensions
 
**Extension Install**
 - set `LNBITS_EXTENSIONS_PATH` to a valid path
 - make sure the extension to be installed has been updated (see [comment](https://github.com/lnbits/lnbits/pull/1940#issuecomment-1717352903) below)
 - install extension
 - check that the extension works as expected
 - upgrade extension and check that the new features are present
 - check that the `lnbits/extensions` directory is empty/does not exist
 - check that the `LNBITS_EXTENSIONS_PATH` has an `extensions` folder that contains the installed extension
 - install older version of the extension and check that the changes are rolled back

**Test data**:
 - manifest file: `https://raw.githubusercontent.com/motorina0/watchonly/main/manifest.json`
 - there are two releases:
    - [v0.1-custom-path-ready](https://github.com/motorina0/watchonly/releases/tag/v0.1-custom-path-ready) - contains the updated paths for the extension
    - [v0.2-ui-api-updates](https://github.com/motorina0/watchonly/releases/tag/v0.2-ui-api-updates) - contains code changes for the extension (see screenshot 1)

### Screenshots
[...]

PR #1962, opened 2023-09-24
Updated dev user guide for extension development

blackcoffeexbt:
Updated the dev guide for extension development to work with the current extension management process

### DIFF
--- a/docs/devs/extensions.md
+++ b/docs/devs/extensions.md
@@ -6,13 +6,13 @@ nav_order: 2
 ---
 
 
-Making extensions
+Extension set up
 =================
 
-Start off by cloning the [example extension](https://github.com/lnbits/example) into your `lnbits/extensions` folder and renaming it to `mysuperplugin`:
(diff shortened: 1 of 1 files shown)


dni (approved):

motorina0:
- the instructions are correct, but they give the impression that for each change in `mysuperplugin` the developer has to make a new release and re-install. Which is not correct. The custom extension path can be used (in the future) or the  symlink command can be used (now)

- we should not use both the `plugin` (as in `mysuperplugin`) and `extension` terms. It will create confusion.

motorina0:
> What should be the recommended approach for devs?
- clone the extension folder outside of the `lnbits` git repo (it can create problems for repo inside repo)
- Use one of these approaches:
  1. symlink command: `ln -s /path/to/myextension /other/path/to/lnbits/extensions` 
  1. set custom extension path  (after this PR is merged https://github.com/lnbits/lnbits/pull/1940/)

arcbtc (approved):

PR #1963, opened 2023-09-24
[FEAT] add extension functionality to lnbits-cli

dni:
`lnbits-cli extensions`

update, install, uninstall, list extensions from CLI

screenshot-1695503917

### DIFF
--- a/lnbits/commands.py
+++ b/lnbits/commands.py
@@ -28,6 +28,13 @@ def db():
     """
 
 
+@lnbits_cli.group()
+def extensions():
+    """
+    Extensions related commands
+    """
+
+
 def get_super_user() -> str:
     """Get the superuser"""
     with open(Path(settings.lnbits_data_folder) / ".super_user", "r") as file:
@@ -118,6 +125,35 @@ async def load_disabled_extension_list() -> None:
     """Update list of extensions that have been explicitly disabled"""
     inactive_extensions = await get_inactive_extensions()
     settings.lnbits_deactivated_extensions += inactive_extensions
+
+
+@extensions.command("list")
+def extensions_list():
+    """Show currently installed extensions"""
+    click.echo("Installed extensions:")
+
+
+@extensions.command("upgrade")
+def extensions_upgrade():
+    """Update all extension"""
+
+
+@extensions.command("update")
+@click.argument("extension")
+def extensions_update(extension: str):
+    """Update a extension"""
+
+
+@extensions.command("install")
+@click.argument("extension")
+def extensions_install(extension: str):
+    """Install a extension"""
+
+
+@extensions.command("uninstall")
+@click.argument("extension")
+def extensions_uninstall(extension: str):
+    """Uninstall a extension"""
 
 
 def main():
--- a/lnbits/commands.py
+++ b/lnbits/commands.py
@@ -214,7 +214,6 @@ def extensions_upgrade(
         return
 
     async def wrap():
-        await check_admin_settings()
         if not await _can_run_operation(url):
             return
 
@@ -510,6 +509,7 @@ async def _call_uninstall_extension(
 
 
 async def _can_run_operation(url) -> bool:
+    await check_admin_settings()
     if await _is_lnbits_started(url):
         if not url:
             click.echo("LNbits server is started. Please either:")
--- a/lnbits/commands.py
+++ b/lnbits/commands.py
@@ -53,22 +53,36 @@ def extensions():
     """
 
 
-def get_super_user() -> str:
+def get_super_user() -> Optional[str]:
     """Get the superuser"""
-    with open(Path(settings.lnbits_data_folder) / ".super_user", "r") as file:
+    superuser_file = Path(settings.lnbits_data_folder, ".super_user")
+    if not superuser_file.exists() or not superuser_file.is_file():
+        raise ValueError(
+            "Superuser id not found. Please check that the file "
+            + f"'{superuser_file.absolute()}' exists and has read permissions."
+        )
+    with open(superuser_file, "r") as file:
         return file.readline()
 
 
 @lnbits_cli.command("superuser")
 def superuser():
     """Prints the superuser"""
-    click.echo(get_super_user())
+    try:
+        click.echo(get_super_user())
+    except ValueError as e:
+        click.echo(str(e))
 
 
 @lnbits_cli.command("superuser-url")
 def superuser_url():
     """Prints the superuser"""
-    click.echo(f"http://{settings.host}:{settings.port}/wallet?usr={get_super_user()}")
+    try:
+        click.echo(
+            f"http://{settings.host}:{settings.port}/wallet?usr={get_super_user()}"
+        )
+    except ValueError as e:
+        click.echo(str(e))
 
 
 @lnbits_cli.command("delete-settings")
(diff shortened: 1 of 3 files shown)


prusnak:
I don't like the `upgrade` command. How about adding `-a`/`--all` option to the `update` command?

``` python
@extensions.command("update")
@click.option("-a", "--all", is_flag=True)
@click.argument("extensions", nargs=-1)
def extensions_update(all: bool, extensions: list[str]):
    """Update extensions"""
    if all:
        # update all extensions
    else
        for extension in extensions:
            click.echo(f"Updating {extension}...")
```

lnbits/core/models.py
```diff
@@ -276,7 +276,7 @@ def from_row(cls, row: Row):
 
 
 class CoreAppExtra:
-    register_new_ext_routes: Callable
+    register_new_ext_routes: Callable = logger.debug
```
dni, line 279: i dont get this?

lnbits/commands.py
```diff
@@ -120,10 +145,372 @@ async def load_disabled_extension_list() -> None:
...
+@click.option(
+    "-i", "--repo-index", help="Select the index of the repository to be used."
+)
+@click.option(
+    "-s", "--source-repo", help="Provide the repository URL to be used for upgrading."
+)
+@click.option(
+    "-u",
+    "--url",
+    help="Use this option to update a runing server. Eg: 'http://localhost:5000'.",
```
dni, line 175:
```suggestion
    help="Use this option to update a running server. Eg: 'http://localhost:5000'.",
```

lnbits/commands.py
```diff
@@ -120,10 +145,372 @@ async def load_disabled_extension_list() -> None:
...
+)
+@click.option(
+    "-u",
+    "--url",
+    help="Use this option to update a runing server. Eg: 'http://localhost:5000'.",
```
dni, line 249:
```suggestion
    help="Use this option to update a running server. Eg: 'http://localhost:5000'.",
```

lnbits/commands.py
```diff
@@ -120,10 +145,372 @@ async def load_disabled_extension_list() -> None:
...
+
+    _run_async(wrap)
+
+
+@extensions.command("uninstall")
+@click.argument("extension")
+@click.option(
+    "-u",
+    "--url",
+    help="Use this option to update a runing server. Eg: 'http://localhost:5000'.",
```
dni, line 276:
```suggestion
    help="Use this option to update a running server. Eg: 'http://localhost:5000'.",
```

lnbits/commands.py
```diff
@@ -120,10 +145,372 @@ async def load_disabled_extension_list() -> None:
...
+
+    async def wrap():
+        if not await _can_run_operation(url):
+            return
+        try:
+            user = User(id=get_super_user(), super_user=True)
+            await _call_uninstall_extension(extension, user, url)
+            click.echo(f"Extension '{extension}' uninstalled.")
+        except HTTPException as ex:
+            click.echo(f"Faield to uninstall '{extension}' Error: '{ex.detail}'.")
```
dni, line 294:
```suggestion
            click.echo(f"Failed to uninstall '{extension}' Error: '{ex.detail}'.")
```

lnbits/commands.py
```diff
@@ -120,10 +145,372 @@ async def load_disabled_extension_list() -> None:
...
+            return False, ex.detail
+        except Exception as ex:
+            click.echo(f"Faield to uninstall '{extension}': {str(ex)}.")
```
dni, line 297:
```suggestion
            click.echo(f"Failed to uninstall '{extension}': {str(ex)}.")
```

prusnak:
I don't have a superuser set (= file `data/.super_user` does not exist). The CLI still wants to read that file, even though I am trying to use the CLI without the LNbits running and it fails.

It should work even when `data/.super_user` does not exist:
- in offline mode this should work without warnings/errors
- in online mode this should show a reasonable error message (=not `data/.super_user does not exist`)

prusnak (changes requested):
I don't have a superuser set (= file `data/.super_user` does not exist). The CLI still wants to read that file, even though I am trying to use the CLI without the LNbits running and it fails.

It should work even when `data/.super_user` does not exist:
- in offline mode this should work without warnings/errors
- in online mode this should show a reasonable error message (=not `data/.super_user does not exist`)

lnbits/commands.py
```diff
@@ -120,10 +145,367 @@ async def load_disabled_extension_list() -> None:
...
+    _run_async(wrap)
+    return
+
+
+@extensions.command("install")
+@click.argument("extension")
+@click.option("--repo-index", help="Select the index of the repository to be used.")
+@click.option(
+    "--source-repo", help="Provide the repository URL to be used for installing."
+)
```
prusnak, line 240: provide short-options too (same as upgrade command)

```suggestion
@click.option("-i", "--repo-index", help="Select the index of the repository to be used.")
@click.option(
    "-s", "--source-repo", help="Provide the repository URL to be used for installing."
)
```

lnbits/core/models.py
prusnak: @motorina0 is this intentional?

motorina0:
> I don't have a superuser set (= file `data/.super_user` does not exist). The CLI still wants to read that file, even though I am trying to use the CLI without the LNbits running and it fails.
> ...

Fixed:
 - added `--admin-user` for allowing the user to specify an admin user id
 - `.super_user` file only required if online mode and `--admin-user` not provided.
 - `.super_user` file not required in offline mode

lnbits/core/models.py
motorina0: - `register_new_ext_routes` must not be `None`
- under normal conditions (when the LNbits server is started) this value is initialized. However when starting from `commands` the value was `None`.
- `logger.debug` was used as a default value because it allowed any arguments, but that was a bad idea
- now an explicit empty `_do_nothing(*_)` function is declared and used

motorina0: fixed in commit: https://github.com/lnbits/lnbits/pull/1963/commits/7adcf1545d482ac76f876bb012830c8c5594a998

motorina0 (approved):

prusnak (changes requested):
Great effort!

One nit-pick, though: I think we should be using the word `update` not `upgrade`.

> "A software update is typically a release containing enhancements to the current version. An upgrade is a whole new version of software that represents a significant change or major improvement."

If you don't have the time to update the PR, let me know, I will do it and then we can merge.

motorina0:
> If you don't have the time to update the PR, let me know, I will do it and then we can merge.

I'll do it

motorina0:
> One nit-pick, though: I think we should be using the word `update` not `upgrade`.

Done!

prusnak (approved):

callebtc:
Nit: 

This PR makes extensive use of a `run_async` wrapper but click can be wrapped by a `@coro` decorator to simply allow for async methods. 

```python
from functools import wraps
import asyncio

def coro(f):
    @wraps(f)
    def wrapper(*args, **kwargs):
        return asyncio.run(f(*args, **kwargs))

    return wrapper
    
@lnbits_cli.command("cmd")
@coro
async def cmd():
   ...
```

motorina0:
> This PR makes extensive use of a `run_async` wrapper but click can be wrapped by a `@coro` decorator to simply allow for async methods.

Done!

PR #1964, opened 2023-09-25
[FEAT] add cache busting via static_url_for and settings.cache_version

### LINKED ISSUE #1954: [Feature request] Prevent hard refresh requirement on new release
Whenever LNbits has a new release, lots of users see only parts of the GUI or none at all / blank page.
Even the advanced users fail to understand immediately, that a hard refresh is needed.

Cache busting on new release

Possible solution could be to give a version to static files (e.g. /static/js/script.js?v=1.0) on release-build.

dni:
closes #1954
this PR add cache busting to `/static`
additionally i combined `lnbits/core/static` with `lnbits/static`, it was not necessary and added a lot of duplicate code for cache busting now you have to include all static files inside the html files with `{{ static_url_for('static', 'app.css' ) }}`

screenshot-1695625845


arcbtc:
Hi @prusnak, I added you as I was concerned this may impact nix stuff. We originally stopping static file building as nix didnt like lnbits changing its own code

dni:
its created a `lnbits/static/_digest` direcotry with all the renamed files and serves them from there.

prusnak:
> We originally stopping static file building as nix didnt like lnbits changing its own code

the change does not change any code, it just creates new files - which is a good point for nix

> its created a `lnbits/static/_digest` direcotry with all the renamed files and serves them from there.

this location shouldn't be used because on nix this would be read-only, maybe we should generate the files not into `lnbits/static/_digest` but into `LNBITS_DATA_FOLDER` or some other configurable location with write enabled?

dni:
> I am wondering whether we shouldn't just introduce `static_url_for(X)` which returns `f"/static/{X}?v={CACHE_VERSION}"` (providing we introduce `CACHE_VERSION` to Python as well) and not use the `fastapi-static-digest` magic.

yep, seems reasonable aswell, ill have another try

dni:
it feels like that was a bit too easy :)

screenshot-1696361814

lnbits/settings.py
```diff
@@ -290,6 +291,8 @@ class EnvSettings(LNbitsSettings):
     lnbits_extensions_path: str = Field(default="lnbits")
     super_user: str = Field(default="")
     version: str = Field(default="0.0.0")
+    # new cache version each time you restart the server
+    cache_version: int = Field(default=time())
```
prusnak, line 295: Neat trick!

prusnak (approved):
I haven't tested this, but I do like this approach much more than introducing fastapi-static-digest dependency.

motorina0:
- I have ran `make bundle` and pushed changes
- `vendor.json` needs to update paths
- see commit https://github.com/lnbits/lnbits/pull/1964/commits/d86472c586eb3ec43290227b8cb79fd2f36d5a60

```diff
-    "/static/vendor/moment.js",
-    "/static/vendor/underscore.js",
...

+    "vendor/moment.js",
+    "vendor/underscore.js",
...

```

lnbits/app.py
```diff
@@ -81,12 +81,10 @@ def create_app() -> FastAPI:
...
     setattr(core_app_extra, "register_new_ratelimiter", register_new_ratelimiter(app))
 
-    app.mount("/static", StaticFiles(packages=[("lnbits", "static")]), name="static")
-    app.mount(
-        "/core/static",
-        StaticFiles(packages=[("lnbits.core", "static")]),
-        name="core_static",
-    )
+    # register static files
+    static_path = Path("lnbits/static")
```
motorina0, line 85: - do not use OS dependent path separators
```suggestion
    static_path = Path("lnbits", "static")
```

lnbits/core/views/generic.py
```diff
@@ -41,7 +41,7 @@
 
 @generic_router.get("/favicon.ico", response_class=FileResponse)
 async def favicon():
-    return FileResponse("lnbits/core/static/favicon.ico")
+    return FileResponse("lnbits/static/favicon.ico")
```
motorina0, line 44: - do not use OS dependent path separator
```suggestion
    return FileResponse(Path("lnbits", "static", "favicon.ico"))
```
- also must add `from pathlib import Path`

lnbits/core/views/generic.py
```diff
@@ -329,7 +329,7 @@ async def lnurlwallet(request: Request):
 
 @generic_router.get("/service-worker.js", response_class=FileResponse)
 async def service_worker():
-    return FileResponse("lnbits/core/static/js/service-worker.js")
+    return FileResponse("lnbits/static/js/service-worker.js")
```
motorina0, line 332:
```suggestion
    return FileResponse(Path("lnbits", "static", "js", "service-worker.js")
```

lnbits/settings.py
motorina0: - this means the client will re-download the files even if nothing has changed in the codebase (only as server restart took place). A bit sub-optimal.

motorina0: - this field (if it will be kept) should have a name that represents its value:
```suggestion
    server_startup_time: int = Field(default=time())
```
- this value could then also be used to for example to show the uptime of the server (on the admin page)

dni: > * this means the client will re-download the files even if nothing has changed in the codebase (only as server restart took place). A bit sub-optimal.

i agree, that is my concern aswell. but i think the tradeoff is worth it. you should try to minimize restarting to only after updates.

lnbits/settings.py
```diff
@@ -299,6 +300,8 @@ class EnvSettings(LNbitsSettings):
     enable_log_to_file: bool = Field(default=True)
     log_rotation: str = Field(default="100 MB")
     log_retention: str = Field(default="3 months")
+    # new cache version each time you restart the server
```
motorina0, line 303: - outdated, the field name now speaks for itself
```suggestion
```

motorina0 (approved):

Issue #1966, opened 2023-09-25
[BUG] Change of denomination crashes GUI


arbadacarbaYK:
When editing "Denomination of FakeWallet" from sats to EUR LNbits shows a blanc page. I would to a screenshot of the menue but .. :P Where to fix ?

arbadacarbaYK:
fixed with https://github.com/lnbits/lnbits/pull/1973

arbadacarbaYK:
fixed with https://github.com/lnbits/lnbits/pull/1973

PR #1973, opened 2023-09-27
[BUG] Fix bad transation call

motorina0:
### Summary
Closes: https://github.com/lnbits/lnbits/issues/1966
Small but very annoying bug. It is a blocker, the user cannot access the `wallet` page.
The translation function call was wrong `:label="$('amount')` instead of `:label="$t('amount')`
This results in an error:
```
bundle.min.js:14 TypeError: $ is not a function
    at Proxy.eval (eval at nl (bundle.min.js:14:113045), <anonymous>:3:21046)
....
    at new ei (bundle.min.js:14:39365)
```

**Steps**:
 - go to Manage Server > Theme and set `Denomination` to EUR
 - go to wallet page

**Actual Result**:
 - the `/wallet` page does not load
 - other pages like `/admin` still work fine. The value can be changed back to `sat`

**Expected Result**:
 - changing the denomination does not affect the wallet page


lnbits/core/static/js/service-worker.js-e
```diff
@@ -0,0 +1,87 @@
...
+    })
+  )
+})
+
+// User can click on the notification message to open wallet
+// Installed app will open when `url_handlers` in web app manifest is supported
+self.addEventListener('notificationclick', function (event) {
+  event.notification.close()
+  event.waitUntil(clients.openWindow(event.notification.data.url))
+})
```
dni, line 87: remoe this file

dni:
`lnbits/static/bundle.min.js`
and serviceworker.js are not needed you are only changing a `html` file. bundle is crrently broken though.

here the pr https://github.com/lnbits/lnbits/pull/1973

dni (approved):

PR #1977, opened 2023-09-29
feat: install wizard on first launch

arcbtc:
adds a install form on first LNbits launch and lets you configure it initially. 
also shows you the super_user -> useful for bundlers

login


motorina0:
Some tests are failing.

talvasconcelos:
> Some tests are failing.

Yes. The middleware is overpowering everything! Don't know how to fix it. Maybe set the `settings.first_install` to False, on test start. @dni ?

dni (approved):
LGTM!

manually tested!

talvasconcelos (approved):
Manually tested also!

lnbits/middleware.py
```diff
@@ -210,3 +210,18 @@ async def block_allow_ip_middleware(request: Request, call_next):
...
+        if request.url.path != "/api/v1/auth/first_install":
+            if (
+                settings.first_install
+                and request.url.path != "/first_install"
+                and not request.url.path.startswith("/static")
+            ):
+                return RedirectResponse("/first_install")
+        return await call_next(request)
+
+    app.middleware("http")(first_install_middleware)
```
motorina0, line 227: We should be careful with middlewares. Their logic is executed for each HTTP request, so they can increase the response time.
In this particular case, the logic is relevant only once in the life of the installation.

lnbits/core/views/generic.py
```diff
@@ -52,6 +52,18 @@ async def home(request: Request, lightning: str = ""):
...
 
+@generic_router.get("/first_install", response_class=HTMLResponse)
+async def first_install(request: Request):
+    if not settings.first_install:
+        return template_renderer().TemplateResponse(
+            "error.html", {"request": request, "err": "Page not available!"}
+        )
+    return template_renderer().TemplateResponse(
+        "core/first_install.html",
+        {"request": request, "super_user": settings.super_user},
```
motorina0, line 63: Why is the `settings.super_user` sent to the front end. This is risky.

lnbits/middleware.py
```diff
@@ -210,3 +210,18 @@ async def block_allow_ip_middleware(request: Request, call_next):
...
+def add_first_install_middleware(app: FastAPI):
+    @app.middleware("http")
+    async def first_install_middleware(request: Request, call_next):
+        if request.url.path != "/api/v1/auth/first_install":
+            if (
+                settings.first_install
+                and request.url.path != "/first_install"
+                and not request.url.path.startswith("/static")
+            ):
+                return RedirectResponse("/first_install")
```
dni, line 224:
```suggestion
        if (
            settings.first_install
            and request.url.path != "/api/v1/auth/first_install"
            and request.url.path != "/first_install"
            and not request.url.path.startswith("/static")
        ):
            return RedirectResponse("/first_install")
```

lnbits/middleware.py
dni: the if could be combined but i dont think a check in memory `settings.first_install ` is expensive

lnbits/settings.py
```diff
@@ -360,6 +360,7 @@ class SaaSSettings(LNbitsSettings):
 
 
 class PersistenceSettings(LNbitsSettings):
+    first_install: bool = Field(default=False)
```
motorina0, line 363: This is not a `PersistenceSettings` (related to how data is persisted).
Maybe `TransientSettings` is more appropriate.:

```py
class TransientSettings(InstalledExtensionsSettings):
    # Transient Settings:
    #  - are initialized, updated and used at runtime
    #  - are not read from a file or from the `settings` table
    #  - are not persisted in the `settings` table when the settings are updated
    #  - are cleared on server restart
```

tests/conftest.py
```diff
@@ -50,6 +50,7 @@ async def app():
     clean_database(settings)
     app = create_app()
     await app.router.startup()
+    settings.first_install = False
```
motorina0, line 53: isn't the default value `False`?

lnbits/middleware.py
motorina0: - do we have underscore in path in other places?
     - `/api/v1/auth/first_install`
 - suggestion:
      - `/api/v1/auth/init`

lnbits/core/views/auth_api.py
```diff
@@ -251,6 +251,25 @@ async def update(
         raise HTTPException(HTTP_500_INTERNAL_SERVER_ERROR, "Cannot update user.")
 
 
+@auth_router.put("/api/v1/auth/first_install")
+async def first_install(data: UpdateUserPassword) -> JSONResponse:
+    if not settings.first_install:
+        raise HTTPException(HTTP_401_UNAUTHORIZED, "This is not your first install")
+    try:
+        assert data.username, "Missing username."
+        await update_account(data.user_id, username=data.username)
```
motorina0, line 260:
```suggestion
        data.user_id = settings.super_user
        await update_account(data.user_id, username=data.username)
```

motorina0: we should never send the `super_user` to the front-end

lnbits/core/models.py
```diff
@@ -141,6 +141,7 @@ class UpdateUserPassword(BaseModel):
     password: str = Query(default=..., min_length=8, max_length=50)
     password_repeat: str = Query(default=..., min_length=8, max_length=50)
     password_old: Optional[str] = Query(default=None, min_length=8, max_length=50)
+    username: Optional[str] = Query(default=..., min_length=2, max_length=20)
```
motorina0, line 144: We should have a new cleaner model instead of `UpdateUserPassword`.

```py
class UpdateSuperuserPassword(BaseModel):
    username: Optional[str] = Query(default=..., min_length=2, max_length=20)
```

lnbits/core/views/auth_api.py
```diff
@@ -251,6 +251,25 @@ async def update(
         raise HTTPException(HTTP_500_INTERNAL_SERVER_ERROR, "Cannot update user.")
 
 
+@auth_router.put("/api/v1/auth/first_install")
+async def first_install(data: UpdateUserPassword) -> JSONResponse:
```
motorina0, line 255:
```suggestion
async def first_install(data: UpdateSuperuserPassword) -> JSONResponse:
```

tests/conftest.py
dni: yes bit it will run the admin ui setup inside the tests which makes it beeing set to True.
this is needed here else the Redirects happen inside the tests

lnbits/settings.py
dni: the comment sounds fitting but why is it inherited from `InstalledExtensionsSettings`?

talvasconcelos: How should I treat this?

motorina0: > How should I treat this?

Just move it under `TransientSettings`

lnbits/core/views/auth_api.py
```diff
@@ -251,6 +252,34 @@ async def update(
...
+            user_id=user_id,
+            password=data.password,
+            password_repeat=data.password_repeat,
+            username=data.username,
+        )
+        await update_user_password(super_user)
+        settings.first_install = False
+        return _auth_success_response(
+            username=super_user.username, user_id=super_user.user_id
+        )
```
motorina0, line 273:
```suggestion
        return _auth_success_response(
            username=super_user.username
        )
```
Let's not put the `user_id` in the auth token yet. The `user_id` only auth option might still be enabled.

lnbits/core/services.py
```diff
@@ -656,6 +656,7 @@ async def init_admin_settings(super_user: Optional[str] = None) -> SuperSettings
     if super_user:
         account = await get_account(super_user)
     if not account:
+        settings.first_install = True
```
motorina0, line 659: This logic is executed only if `lnbits_admin_ui` is true. Is that expected?

motorina0: This only works for the **first** server startup.
**Scenario**:
 - start the server
 - open the browser -> the `first_install` page is served
 - do not enter new username/pass for `super_admin`
 - restart the server
 - open the browser -> `Page not available!` is shown (NOK)
 - expected:  the `first_install` page is served even after a server restart

lnbits/core/templates/core/first_install.html
```diff
@@ -0,0 +1,139 @@
+{% extends "public.html" %} {% block page %}
```
motorina0, line 1: nitpick: should submit form on `Enter`

lnbits/core/services.py
motorina0: Fixed in this commit: [c903d1d92adc076779acfff3865c1ce138046116](https://github.com/lnbits/lnbits/pull/1977/commits/c903d1d92adc076779acfff3865c1ce138046116)

lnbits/core/views/auth_api.py
motorina0: fixed with this commit: https://github.com/lnbits/lnbits/pull/1977/commits/0e0439acb0ff66dea1b98aa40da0aeaa5065c4ff

lnbits/core/models.py
```diff
@@ -141,6 +145,13 @@ class UpdateUserPassword(BaseModel):
...
+
+
+class UpdateSuperuserPassword(BaseModel):
+    username: Optional[str] = Query(default=..., min_length=2, max_length=20)
```
motorina0, line 152: should `username` this be optional?

lnbits/core/views/auth_api.py
```diff
@@ -251,6 +252,36 @@ async def update(
         raise HTTPException(HTTP_500_INTERNAL_SERVER_ERROR, "Cannot update user.")
 
 
+@auth_router.put("/api/v1/auth/first_install")
+async def first_install(data: UpdateSuperuserPassword) -> JSONResponse:
+    if not settings.first_install:
+        raise HTTPException(HTTP_401_UNAUTHORIZED, "This is not your first install")
+    try:
+        assert data.username, "Missing username."
```
motorina0, line 260: if `UpdateSuperuserPassword.username` is made not optional, then this `assert` is refundant

lnbits/core/views/generic.py
```diff
@@ -52,6 +52,18 @@ async def home(request: Request, lightning: str = ""):
     )
 
 
+@generic_router.get("/first_install", response_class=HTMLResponse)
+async def first_install(request: Request):
+    if not settings.first_install:
+        return template_renderer().TemplateResponse(
+            "error.html", {"request": request, "err": "Page not available!"}
```
motorina0, line 59: nitpick: use a more friendly message, eg: `Super user account has already been configured.`

lnbits/core/views/auth_api.py
motorina0: fixed here: https://github.com/lnbits/lnbits/pull/1977/commits/8a25f65e8f62543a58190a59923c929e3145520b

lnbits/core/models.py
motorina0: fixed here: https://github.com/lnbits/lnbits/pull/1977/commits/8a25f65e8f62543a58190a59923c929e3145520b

motorina0 (approved):
tACK

PR #1979, opened 2023-10-02
nix: fix nixosModule and build

MatthewCroughan:
There wasn't too much bitrot over time, this is a minimal PR to update potree2nix, get lnbits building again and passing the existing vmTest which is buildable via `nix build .#checks.x86_64-linux.vmTest -L` or `nix flake check` for example.

I've tested this as working on x86_64-linux and aarch64-linux.

It would be really nice if we could explain how to use the nixosModule in the Documentation rather than telling people to just `nix build` and run the program. The correct way to use the Nix work is via `services.lnbits.enable = true`, but this can be done in a later PR, and I'll endeavor to write that document.

### DIFF
--- a/nix/modules/lnbits-service.nix
+++ b/nix/modules/lnbits-service.nix
@@ -90,7 +90,6 @@ in
         LNBITS_DATA_FOLDER = "${cfg.stateDir}";
         LNBITS_EXTENSIONS_PATH = "${cfg.stateDir}/extensions";
         LNBITS_PATH = "${cfg.package.src}";
-        LNBITS_COMMIT = "${cfg.package.meta.rev}";
       };
       serviceConfig = {
         User = cfg.user;
--- a/nix/modules/lnbits-service.nix
+++ b/nix/modules/lnbits-service.nix
@@ -88,6 +88,9 @@ in
       after = [ "network-online.target" ];
       environment = {
         LNBITS_DATA_FOLDER = "${cfg.stateDir}";
+        LNBITS_EXTENSIONS_PATH = "${cfg.stateDir}/extensions";
+        LNBITS_PATH = "${cfg.package.src}";
+        LNBITS_COMMIT = "${cfg.package.meta.rev}";
       };
       serviceConfig = {
         User = cfg.user;
--- a/flake.lock
+++ b/flake.lock
@@ -1,17 +1,41 @@
 {
   "nodes": {
     "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
       "locked": {
-        "lastModified": 1676283394,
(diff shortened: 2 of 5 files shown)


prusnak:
If https://github.com/lnbits/lnbits/pull/1980 goes in first (and it should), we could remove the `lnbits_commit` magic from this PR.

arcbtc:
Added you as reviewer @prusnak  so we can merge LFGM! https://github.com/lnbits/lnbits/pull/1980

nix/modules/lnbits-service.nix
```diff
@@ -88,6 +88,9 @@ in
       after = [ "network-online.target" ];
       environment = {
         LNBITS_DATA_FOLDER = "${cfg.stateDir}";
+        LNBITS_EXTENSIONS_PATH = "${cfg.stateDir}/extensions";
+        LNBITS_PATH = "${cfg.package.src}";
+        LNBITS_COMMIT = "${cfg.package.meta.rev}";
```
prusnak, line 93: no longer needed:

```suggestion
```

prusnak (changes requested):

prusnak (approved):

motorina0:
@MatthewCroughan thanks for updating these `nix` files. 
The `nix` syntax and file structure is not a straight forward one. In order to be able to easily maintain & update the `nix` configs in the future can you please:
  - [ ] add as much comments as possible inside the `.nix` files
    - links to to https://nix.dev/tutorials/ or other resource inside the `.nix` files are OK
  - [ ] add a README with the nix commands to be ran
    - `nix-shell ...`, `nix-build ...`, etc
    - for example `flakes` requires the experimental feature flag. We should not search how to do this each time a new person tries to run lnbits with nix
  - [ ] is `flakes` still supported and used (see https://nix.dev/concepts/flakes#why-are-flakes-controversial)? 
```
This led to a situation where the stable interface was only sparsely maintained for multiple years, and repeatedly suffered breakages due to ongoing development. Meanwhile, the new interface was adopted widely enough for evolving its design without negatively affecting users to become very challenging.
```

MatthewCroughan:
Are you going to block the update until I add comments? I'm going to refrain from doing that as I don't have the time right now, and I was making this PR so that I could give a talk at bitcoin++. As far as nix code goes, this is a very minimal Flake, you just have to learn Nix. You cannot expect to be able to read a language you do not know, no amount of comments will fix that.

I will endeavour as I make PRs in the future to upgrade the docs, I already mentioned this intention in the PR text. but this PR is not for that.

MatthewCroughan:
@motorina0 the build commands are as follows:

`nix build` in the root of the repo to build lnbits
`nix flake check` in the root of the repo to run the VM Tests which run a full integration test of lnbits

PR #1981, opened 2023-10-03
[CHORE] cleanup ip block middleware workaround for older fastapi

dni:
removes a workaround and TODO left in the code.

### DIFF
--- a/lnbits/middleware.py
+++ b/lnbits/middleware.py
@@ -2,7 +2,7 @@ from http import HTTPStatus
 from typing import Any, List, Tuple, Union
 from urllib.parse import parse_qs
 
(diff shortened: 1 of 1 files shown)


motorina0:
- wouldn't this change bring back a lot of exceptions in the log?

dni:
there should not be any exceptions in the first place, it was a bug in fastapi which is solved in newer versions. 
https://stackoverflow.com/questions/71222144/runtimeerror-no-response-returned-in-fastapi-when-refresh-request

motorina0 (approved):

prusnak (approved):
untested ACK

PR #1982, opened 2023-10-03
fix: admin_ui js error

dni:
undefined formData.lnbits_allowed_currencies JS error in console

### DIFF
--- a/lnbits/core/templates/admin/_tab_server.html
+++ b/lnbits/core/templates/admin/_tab_server.html
@@ -75,7 +75,7 @@
             clearable
             hint="Default currency for accounting"
             label="Currency"
-            :options="formData.lnbits_allowed_currencies.length ? formData.lnbits_allowed_currencies : {{ currencies }}"
+            :options="formData.lnbits_allowed_currencies && formData.lnbits_allowed_currencies.length ? formData.lnbits_allowed_currencies : {{ currencies }}"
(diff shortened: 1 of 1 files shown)


lnbits/core/templates/admin/_tab_server.html
```diff
@@ -75,7 +75,7 @@ <h6 class="q-my-none">Server Management</h6>
             clearable
             hint="Default currency for accounting"
             label="Currency"
-            :options="formData.lnbits_allowed_currencies.length ? formData.lnbits_allowed_currencies : {{ currencies }}"
+            :options="formData.lnbits_allowed_currencies && formData.lnbits_allowed_currencies.length ? formData.lnbits_allowed_currencies : {{ currencies }}"
```
motorina0, line 78: - why not use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)?
```js
formData.lnbits_allowed_currencies?.length
```

dni, line 78:
```suggestion
            :options="formData.lnbits_allowed_currencies?.length ? formData.lnbits_allowed_currencies : {{ currencies }}"
```

like this?

motorina0: yes

PR #1990, opened 2023-10-05
refactor: move admin js into .js file

dni:
just some simple cleanup for better readability


motorina0 (approved):

prusnak (approved):
I confirmed the code removed from admin/index.html is the same as the one added to js/admin.js

PR #1992, opened 2023-10-05
refactor: split admin_api and superuser_api

dni:
adds dependencies `check_admin` and `check_super_user` to APIRouter. 
goal of this pr is to make it clearer which api endpoint has which permissions.

### DIFF
--- a/lnbits/core/__init__.py
+++ b/lnbits/core/__init__.py
@@ -8,6 +8,7 @@ from .views.api import api_router
 from .views.generic import generic_router, update_user_extension
 from .views.node_api import node_router, public_node_router, super_node_router
 from .views.public_api import public_router
+from .views.superuser_api import superuser_router
 
 # backwards compatibility for extensions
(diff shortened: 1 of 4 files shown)


jackstar12 (approved):

motorina0:
- I think having a path based on permissions is not a good idea (eg: `'/superuser/api/v1/restart/?usr='`)
  - the permissions come and go. What if in the future we add permissions based on `ACL`s (Access Control List)
  - the `admin` segment of the path refers to `"Administration of the server"`, not to the `admin` user level
  - the changes are not backwards compatible (some users might have built API calls on the `/admin/...` path)

dni:
hm, i think there is a clear difference between a super user and an admin user, no matter what ACL we might introduce?

motorina0:
> hm, i think there is a clear difference between a super user and an admin user, no matter what ACL we might introduce?

- even so, these differences should not leak in the REST path

PR #1998, opened 2023-10-05
fix: Closed channels affecting the stats

iWarpBTC:
So I've updated to 0.11.0, rushed the new node stats and it just doesn't seem right to me. Closed channels are being included. Here I'm offering my humble hot fix. 

Maybe also inactive channels should be included in some/all of those stats? Let me know what you think.

### DIFF
--- a/lnbits/nodes/base.py
+++ b/lnbits/nodes/base.py
@@ -70,7 +70,7 @@ class ChannelStats(BaseModel):
 
         return cls(
             counts=counts,
-            avg_size=int(sum(active_channel_sizes) / len(channels)),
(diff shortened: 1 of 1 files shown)


jackstar12:
> Maybe also inactive channels should be included in some/all of those stats? Let me know what you think.

I think only including active ones is fine

jackstar12 (approved):

iWarpBTC (commented):
LGTM, thanks!

prusnak (approved):

dni (approved):

motorina0 (approved):

prusnak:
4 acks - merging

PR #1999, opened 2023-10-05
fix: use correct field from response (old one got removed)

jackstar12:


dni (approved):

jackstar12:
> * give some context please and some testing steps

cln removed the `value` field from the `listfunds` rpc in v23, which wasnt caught because regtest ran on v22

motorina0:
> cln removed the `value` field from the `listfunds` rpc somewhere in v23, which wasnt caught because regtest ran on v22

- when was `"amount_msat"` field first introduced? Pre `"amount_msat"` clightning versions will fail.

jackstar12:
> > cln removed the `value` field from the `listfunds` rpc somewhere in v23, which wasnt caught because regtest ran on v22
> 
>     * when was `"amount_msat"` field first introduced? Pre `"amount_msat"` clightning versions will fail.

I know that it was already present and working in v22

motorina0:
> I know that it was already present and working in v22
- ok, but if `"amount_msat"` was introduced, lets say, in `v15` then it means `clightning` `v14` will not work with this PR
- note: `v15` is just an example, it can be anything between `1` and `22`
- as an alternative the PR can check for `"value"` and if not present use `"amount_msat"`

dni:
it was `msatoshi` before

https://github.com/ElementsProject/lightning/commit/cd7e784d6f503553548654ca056c8faaa091348d

dni:
looking at cln `0.12.1` i also the `amount_msat` here https://github.com/ElementsProject/lightning/blob/fda3904d8c4d657985568f34444b3f59b2527aae/cln-rpc/src/model.rs#L1168

motorina0 (approved):
LGTM, with this observation from @dni :
```
everyone running cln < 0.12. is basically suicidal
```
- should be added to the readme ☝️

PR #2005, opened 2023-10-06
fix: make bolt11 optional in `NodePayment`

jackstar12:
not every payment has a bolt11 present

### DIFF
--- a/lnbits/nodes/base.py
+++ b/lnbits/nodes/base.py
@@ -104,7 +104,7 @@ class NodePayment(BaseModel):
     fee: Optional[int] = None
     memo: Optional[str] = None
     time: int
-    bolt11: Optional[str]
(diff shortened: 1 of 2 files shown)


dni (approved):

lnbits/nodes/base.py
```diff
@@ -104,7 +104,7 @@ class NodePayment(BaseModel):
     fee: Optional[int] = None
     memo: Optional[str] = None
     time: int
-    bolt11: str
+    bolt11: Optional[str]
```
prusnak, line 107: shouldn't this add `= None` at the end like other optional fields above and below?

```suggestion
    bolt11: Optional[str] = None
```

prusnak (approved):

PR #2011, opened 2023-10-07
dependency cleanup

prusnak:
- async-timeout is not used anywhere (nor core neither extensions)
- cerberus is not used anywhere (nor core neither extensions)
- move embit and secp256k1 to end of the list
  - these are not used by core
  - document which extensions use them


jackstar12 (approved):

callebtc (approved):

dni (approved):
love it! especially the comments!
LGTM

PR #2017, opened 2023-10-08
add possibility to turn off new registrations 

prusnak:
with `LNBITS_CREATE_ACCOUNT=false` while keeping existing users functional (i.e. no allowlist for existing users)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,6 +21,9 @@ LNBITS_ADMIN_USERS=""
 # Extensions only admin can access
 LNBITS_ADMIN_EXTENSIONS="ngrok, admin"
 
+# Disable account creation for new users
+# LNBITS_ALLOW_NEW_ACCOUNTS=false
+
 # Enable Admin GUI, available for the first user in LNBITS_ADMIN_USERS if available
 # Warning: Enabling this will make LNbits ignore this configuration file. Your settings will
 # be stored in your database and you will be able to change them only through the Admin UI.
--- a/lnbits/core/templates/core/index.html
+++ b/lnbits/core/templates/core/index.html
@@ -3,6 +3,7 @@
 {% endblock %} {% block page %}
 <div class="row q-col-gutter-md justify-center">
   <div class="col-12 col-md-7 col-lg-6 q-gutter-y-md">
+    {% if lnurl or LNBITS_NEW_ACCOUNTS_ALLOWED %}
     <q-card>
       <q-card-section>
         {% if lnurl %}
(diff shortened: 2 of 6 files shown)


dni (approved):

lnbits/core/views/api.py
```diff
@@ -172,7 +172,7 @@ async def api_create_wallet(
 
 @api_router.post("/api/v1/account", response_model=Wallet)
 async def api_create_account(data: CreateWallet) -> Wallet:
-    if len(settings.lnbits_allowed_users) > 0:
+    if not settings.lnbits_create_account or len(settings.lnbits_allowed_users) > 0:
```
motorina0, line 175: - nitpick: add property to settings:
```py
@property
def allow_new_user_creation(self):
   return settings.lnbits_create_account and len(settings.lnbits_allowed_users) == 0
```

lnbits/helpers.py
```diff
@@ -50,6 +50,9 @@ def template_renderer(additional_folders: Optional[List] = None) -> Jinja2Templa
     t.env.globals["SITE_DESCRIPTION"] = settings.lnbits_site_description
     t.env.globals["LNBITS_THEME_OPTIONS"] = settings.lnbits_theme_options
     t.env.globals["LNBITS_VERSION"] = settings.version
+    t.env.globals["LNBITS_CREATE_ACCOUNT"] = (
+        settings.lnbits_create_account and len(settings.lnbits_allowed_users) == 0
```
motorina0, line 54: - use `settings. allow_new_user_creation` property

.env.example
```diff
@@ -21,6 +21,9 @@ LNBITS_ADMIN_USERS=""
 # Extensions only admin can access
 LNBITS_ADMIN_EXTENSIONS="ngrok, admin"
 
+# Disable account creation for new users
+# LNBITS_CREATE_ACCOUNT=false
```
motorina0, line 25: - the variable name sounds more like an action
```suggestion
# LNBITS_ALLOW_ACCOUNT_CREATION=false
```

lnbits/core/templates/core/index.html
```diff
@@ -33,6 +34,7 @@
         {% endif %}
```
motorina0, line 34: - maybe show an warning that new user creation is not allowed on this instance
- the read-only landing page can create confusion

image

lnbits/core/views/api.py
prusnak: addressed in b9e6b915c477abdf5d3267ac4a5fa3b221c0ffaf

lnbits/helpers.py
prusnak: addressed in b9e6b915c477abdf5d3267ac4a5fa3b221c0ffaf

lnbits/core/templates/core/index.html
prusnak: I prefer to keep the design clean on the front page and I feel it is nicer not to put any warning there.

motorina0: OK, with the Login PR https://github.com/lnbits/lnbits/pull/1653  there is going to be a Login form (even if new users are not allowed)

PR #2025, opened 2023-10-10
run optipng on logos + remove old logo from docs

prusnak:
https://docs.lnbits.org still uses the old logo

this PR fixes this and removes old logos from the repository

this PR also runs `optipng -o 7` on LNbits logos to reduce their size (especially when we now include logo in the QR code)

---

the old logo is also used on main github page here: https://github.com/lnbits

and also in the telegram group here: https://t.me/lnbits

but these two places need to be changes by administrators


jackstar12 (approved):

dni:
screenshot-1697436596
screenshot-1697435123

svg and png have this weird border? i have a look at them

dni:
can you run the optipng, again on the 2 PNG's?

should me maybe consider reducing the size off 900px x 900px if the icon, i seems very  excessive, i think 200x200 is enough.  or we include an additional lnbits-small.svg for the QR

PR #2026, opened 2023-10-10
add "nix build" to GitHub CI

prusnak:
builds on top of https://github.com/lnbits/lnbits/pull/2027 which is needed for the successful nix build

The `nix build` action in the CI [takes around **~36** minutes](https://github.com/lnbits/lnbits/actions/runs/6473111117/job/17575129639?pr=2026), so it is run only if run only when poetry/flake files are changed

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,28 @@
+name: LNbits CI / nix
+
+on:
+  push:
+    branches:
+        - main
+        - dev
+    paths:
+      - 'flake.nix'
+      - 'flake.lock'
+      - 'pyproject.toml'
+      - 'poetry.lock'
+  pull_request:
+    paths:
+      - 'flake.nix'
+      - 'flake.lock'
+      - 'pyproject.toml'
+      - 'poetry.lock'
+
+jobs:
+  nix:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v23
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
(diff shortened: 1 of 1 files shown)


dni:
WTF, it takes 36 minutes... ?

LOL, i dont have any words for that

dni (approved):

.github/workflows/nix.yml
```diff
@@ -0,0 +1,25 @@
+name: LNbits CI / nix
+
+on:
+  push:
+    paths:
```
dni, line 5:
```suggestion
    branches:
        - main
        - dev
    paths:
```

jackstar12 (approved):

.github/workflows/nix.yml
```diff
@@ -0,0 +1,28 @@
...
+    branches:
+      - main
+      - dev
+    paths:
+      - 'flake.nix'
+      - 'flake.lock'
+      - 'pyproject.toml'
+      - 'poetry.lock'
+  pull_request:
+    paths:
```
motorina0, line 14: If I understand correctly:
- the flow will only run if one of these `path` files is changed
- it will not run when the 'lnbits' logic is changed
- it only checks that the new dependency or new versions of dependencies are nixable

Suggestion:
 - call the file `check-nix-dependencies.yml`

motorina0 (approved):

.github/workflows/nix.yml
prusnak: > If I understand correctly:

Yes

> call the file check-nix-dependencies.yml

I prefer to keep the filename, because in the future this workflow file can contain more nix specific stuff

PR #2027, opened 2023-10-11
downgrade rpds-py to 0.10.3 which is not broken on nix

prusnak:
extracted the first commit from https://github.com/lnbits/lnbits/pull/2026

so we can fix nix install for lnbits-0.11.1


motorina0 (approved):

dni (changes requested):
should it be locked inside pyproject.toml aswell?
else it will be upgraded again on any poetry lock or add, or not?

dni (approved):

PR #2039, opened 2023-10-17
fix: explicitly specify the DB name when migrating

motorina0:
### Summary
Each extension has a DB version stored in the `dbversions` table. When an extension is upgraded this version can change if there are new scripts in `migration.py` file.

The bug is on extension upgrade. The temp upgraded module name was used for the `dbversions` table:
image

The fix uses the explicit DB name.

**Test Steps**:
 - install an extension with a lower version (eg: satspay v0.2.3)
 - upgrade to a higher version (eg: satspay v0.2.5)

**Actual Result**:
 - two entries for `satspay` are now in the `dbversions` (see screenshot)

**Expected Result**:
 - the `satspay` entry gets updated to version `7`
 - no new `satspay` entry in `dbversions`

**Note**:
 - I suspect this bug was introduced with this PR https://github.com/lnbits/lnbits/pull/1940

### DIFF
--- a/lnbits/commands.py
+++ b/lnbits/commands.py
@@ -89,7 +89,7 @@ async def migrate_databases():
 
         current_versions = await get_dbversions(conn)
         core_version = current_versions.get("core", 0)
-        await run_migration(conn, core_migrations, core_version)
+        await run_migration(conn, core_migrations, "core", core_version)
 
     for ext in get_valid_extensions():
         current_version = current_versions.get(ext.code, 0)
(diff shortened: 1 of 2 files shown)


dni (approved):

jackstar12 (approved):
> * I suspect this bug was introduced with this PR [Support custom path for installed extensions #1940](https://github.com/lnbits/lnbits/pull/1940)

Just checked, it happened here:
https://github.com/lnbits/lnbits/pull/1940/files#diff-82afd4105ca2e11a7f9683427475878700edea7b4fb0d15321b386970e4a89d0L185-R197

The update module name broke the `[-2]` index, which was pretty dirty.

PR #2044, opened 2023-10-17
add python-bitcoinlib to dependencies (required by cashu)

prusnak:
here cashu requires python-bitcoinlib, but it is not currently in lnbits pyproject.toml

https://github.com/lnbits/cashu/blob/0.3.0/lib/cashu/core/script.py#L9-L18

this PR fixes this problem


prusnak:
other and probably more preferred way would be to update `cashu` extension that does not need a new dependency, but rather `embit` which is already used by boltz, lnurldevice, watchonly extensions

prusnak:
Closed in favor of https://github.com/lnbits/cashu/issues/44

PR #2050, opened 2023-10-18
Add service fee to specific wallet

callebtc:
Adds a service fee to all transactions. Will consume an env variable `lnbits_service_fee_wallet` to which service fees are deducted to. 

env `lnbits_service_fee_ignore_internal` disables fees for internal payments


callebtc:
`service_fee` should probably be included in `fee_reserve` so that external modules can call it to get the actual fee required now. `fee_reserve` itself should probably include a check whether a payment is internal or not. otherwise it's going to get more complicated for an extension (say splitpayment) to figure out things like: 

- is this an internal payment?
- if yes, is there a fee reserve?
- is there a service fee?
- if there a service fee for internal payments?

All this should be abstracted away from an external module.

.env.example
```diff
@@ -78,7 +78,13 @@ LNBITS_EXTENSIONS_DEFAULT_INSTALL="tpos"
 LNBITS_DATA_FOLDER="./data"
 # LNBITS_DATABASE_URL="postgres://user:password@host:port/databasename"
 
+# Value between 0 and 1, 0 means no fee, 1 means 100% fee
 LNBITS_SERVICE_FEE="0.0"
+# Max fee to charge, 0 will leave no max and use LNBITS_SERVICE_FEE %
```
prusnak, line 83:
```suggestion
# Max fee to charge per transaction (in satoshi), 0 will leave no max and use LNBITS_SERVICE_FEE %
```

lnbits/core/services.py
```diff
@@ -502,6 +521,23 @@ def fee_reserve(amount_msat: int) -> int:
     return max(int(reserve_min), int(amount_msat * reserve_percent / 100.0))
 
 
+def service_fee(amount_msat: int) -> int:
+    service_fee_percent = settings.lnbits_service_fee
```
prusnak, line 525: The comment in `.env.example` says that "1.0" is "100% fee", so the fee is not specified in percent.

We need to either fix the code or the comment.

prusnak (changes requested):

prusnak (approved):
not fully understood the math in `lnbits/core/services.py` but the rest looks OK, untested ACK

dni (approved):

Issue #2055, opened 2023-10-22
[BUG] transactions tab broken for CLN keysend


kallerosenbaum:
**Describe the bug**
The "transactions" tab in the "manage node" page, gives red error message:

'amount_msat'
500 INTERNAL SERVER ERROR

I'm running Core Lightning 23.08

**Desktop (please complete the following information):**

- LNbits version: 0.11.1
- Database sqlite

**Additional context**
Here's the last few lines of the stack trace:
```
Oct 22 16:30:49 wilfrid poetry[839641]:   File "/mnt/bitcoin/.cache/pypoetry/virtualenvs/lnbits-xoNZnh6n-py3.10/lib/python3.10/site-packages/starlette/routing.py", line 66, in app
Oct 22 16:30:49 wilfrid poetry[839641]:     response = await func(request)
Oct 22 16:30:49 wilfrid poetry[839641]:   File "/mnt/bitcoin/.cache/pypoetry/virtualenvs/lnbits-xoNZnh6n-py3.10/lib/python3.10/site-packages/fastapi/routing.py", line 273, in app
Oct 22 16:30:49 wilfrid poetry[839641]:     raw_response = await run_endpoint_function(
Oct 22 16:30:49 wilfrid poetry[839641]:   File "/mnt/bitcoin/.cache/pypoetry/virtualenvs/lnbits-xoNZnh6n-py3.10/lib/python3.10/site-packages/fastapi/routing.py", line 190, in run_endpoint_function
Oct 22 16:30:49 wilfrid poetry[839641]:     return await dependant.call(**values)
Oct 22 16:30:49 wilfrid poetry[839641]:   File "/mnt/bitcoin/git/lnbits-legend/lnbits/core/views/node_api.py", line 141, in api_get_invoices
Oct 22 16:30:49 wilfrid poetry[839641]:     return await node.get_invoices(filters)
Oct 22 16:30:49 wilfrid poetry[839641]:   File "/mnt/bitcoin/git/lnbits-legend/lnbits/nodes/cln.py", line 46, in wrapper
Oct 22 16:30:49 wilfrid poetry[839641]:     return await f(*args, **kwargs)
Oct 22 16:30:49 wilfrid poetry[839641]:   File "/mnt/bitcoin/git/lnbits-legend/lnbits/nodes/cln.py", line 311, in get_invoices
Oct 22 16:30:49 wilfrid poetry[839641]:     data=[
[...]

PR #2056, opened 2023-10-22
CLN: Handle optional invoice fields

kallerosenbaum:
Fixes https://github.com/lnbits/lnbits/issues/2055

Never coded python before, so I had to wing it, but at least it works.

### DIFF
--- a/lnbits/nodes/cln.py
+++ b/lnbits/nodes/cln.py
@@ -310,8 +310,11 @@ class CoreLightningNode(Node):
         return Page(
             data=[
                 NodeInvoice(
-                    bolt11=invoice.get("bolt11") if invoice.get("bolt11") != None else invoice.get("bolt12"),
-                    amount=invoice.get("amount_msat") if invoice.get("amount_msat") != None else invoice.get("amount_received_msat"),
+                    bolt11=invoice.get("bolt11") or invoice.get("bolt12"),
+                    amount=(
+                        invoice.get("amount_msat")
+                        or invoice.get("amount_received_msat")
+                    ),
                     preimage=invoice.get("payment_preimage"),
                     memo=invoice.get("description"),
                     paid_at=invoice.get("paid_at"),
--- a/lnbits/nodes/cln.py
+++ b/lnbits/nodes/cln.py
@@ -310,10 +310,10 @@ class CoreLightningNode(Node):
         return Page(
             data=[
                 NodeInvoice(
-                    bolt11=invoice["bolt11"],
-                    amount=invoice["amount_msat"],
+                    bolt11=invoice.get("bolt11") if invoice.get("bolt11") != None else invoice.get("bolt12"),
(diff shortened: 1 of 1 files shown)


lnbits/nodes/cln.py
```diff
@@ -310,10 +310,10 @@ async def get_invoices(
         return Page(
             data=[
                 NodeInvoice(
-                    bolt11=invoice["bolt11"],
-                    amount=invoice["amount_msat"],
+                    bolt11=invoice.get("bolt11") if invoice.get("bolt11") != None else invoice.get("bolt12"),
```
kallerosenbaum, line 313: Unfortunately NodeInvoice has a bolt11 field, but no bolt12 field. Now I shoe-horn a bolt12 onto a field named "bolt11".

I suggest renaming NodeInvoice.bolt11 to NodeInvoice.bolt1x or something. Another approach is to use two fields in NodeInvoice, but that unnecessarily complicates things, I think.

jackstar12: I think a seperate field would be the best way. We'd also have to consider how to show that in the ui aswell

prusnak: There are so many places where the code expects `bolt11` to be set, I found it extremely difficult to refactor it now. I think that as a short-term/mid-term fix this will have to do.

lnbits/nodes/cln.py
```diff
@@ -310,10 +310,10 @@ async def get_invoices(
...
+                    amount=invoice.get("amount_msat") if invoice.get("amount_msat") != None else invoice.get("amount_received_msat"),
```
prusnak, line 314: This is more idiomatic and captures better what we want to do (if value is set, we want to use it, otherwise throw an exception, we do not want to silently pass `None`).

```suggestion
                    bolt11=invoice["bolt11"] if "bolt11" in invoice else invoice["bolt12"],
                    amount=invoice["amount_msat"] if "amount_msat" in invoice else invoice["amount_received_msat"],
```

prusnak (changes requested):

lnbits/nodes/cln.py
prusnak: @kallerosenbaum can you please test whether this suggested code works for you?

jackstar12: > This is more idiomatic and captures better what we want to do (if value is set, we want to use it, otherwise throw an exception, we do not want to silently pass `None` because neither `bolt11` nor `amount` are optional fields).

these changes are not required as pydantic will throw an exception if we pass `None` to either of these fields. We can actually simplify it to
```python
bolt11=invoice.get("bolt11") or invoice.get("bolt12"),
```

kallerosenbaum:
Can confirm https://github.com/lnbits/lnbits/pull/2056/commits/eef03f225296dccc9c9ae1a1842931ed327fa41c works for me. I had actual (not just theoretical) issues with description, bolt11, and amount_msat being unset, and now it works.

prusnak (approved):

jackstar12 (approved):

PR #2059, opened 2023-10-24
[main/fix] fix docs/guide/installation

prusnak:
Fix install instructions after #2052 and #2060

Also these have been merged to `main`, not to `dev`, so this is also targeted to `main`. We will then merge `main` into `dev`.


dni (changes requested):
its still based on main?

PR #2062, opened 2023-10-26
LNDRestWallet: catch http errors

callebtc:
This PR catches the errors in `lndrest.py` that can happen when for example the LND certificate has expired. 

This can lead to stuck payments.

### DIFF
--- a/lnbits/wallets/lndrest.py
+++ b/lnbits/wallets/lndrest.py
@@ -114,7 +114,12 @@ class LndRestWallet(Wallet):
                 hashlib.sha256(unhashed_description).digest()
             ).decode("ascii")
 
-        r = await self.client.post(url="/v1/invoices", json=data)
+        try:
+            r = await self.client.post(url="/v1/invoices", json=data)
+        except Exception as exc:
+            return InvoiceResponse(
+                False, None, None, f"LndRestWallet create_invoice error: {str(exc)}"
+            )
 
         if r.is_error:
             error_message = r.text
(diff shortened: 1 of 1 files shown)


lnbits/wallets/lndrest.py
```diff
@@ -168,10 +183,9 @@ async def get_payment_status(self, checking_id: str) -> PaymentStatus:
         """
         # convert checking_id from hex to base64 and some LND magic
         try:
-            checking_id = base64.urlsafe_b64encode(bytes.fromhex(checking_id)).decode(
-                "ascii"
-            )
-        except ValueError:
+            checking_id = base64.b64encode(bytes.fromhex(checking_id)).decode("ascii")
+        except Exception as exc:
+            logger.error(f"LndRestWallet error get_payment_status: {str(exc)}")
```
motorina0, line 188: - is `str(exc)` required? If yes, then it should be present on line `170` also

lnbits/wallets/lndrest.py
```diff
@@ -136,24 +141,34 @@ async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse
...
-            timeout=None,
-        )
+        try:
+            r = await self.client.post(
+                url="/v1/channels/transactions",
+                json={"payment_request": bolt11, "fee_limit": lnrpcFeeLimit},
+                timeout=None,
+            )
+        except Exception as exc:
+            return PaymentResponse(
```
motorina0, line 151: - the error should also be logged

lnbits/wallets/lndrest.py
```diff
@@ -114,7 +114,12 @@ async def create_invoice(
                 hashlib.sha256(unhashed_description).digest()
             ).decode("ascii")
 
-        r = await self.client.post(url="/v1/invoices", json=data)
+        try:
+            r = await self.client.post(url="/v1/invoices", json=data)
+        except Exception as exc:
```
motorina0, line 119: - the error should also be logged. We cannot rely on the catcher that will log it (maybe it just displays on the UI)

lnbits/wallets/lndrest.py
prusnak: No, it is not needed.

See:

```
>>> ex = Exception("foo bar")
>>> str(ex)
'foo bar'
>>> f"{ex}"
'foo bar'
>>> f"{str(ex)}"
'foo bar'
```

prusnak (changes requested):

lnbits/wallets/lndrest.py
prusnak, line 188: Yes, not needed:

```suggestion
            logger.error(f"LndRestWallet error get_payment_status: {exc}")
```

prusnak: Yes, not needed:

```suggestion
            logger.error(f"LndRestWallet error get_payment_status: {exc}")
```

PR #2065, opened 2023-10-26
add comment why CLightningWallet alias is needed

prusnak:
There is an extra empty line because black wants it that way and ruff does not complain.


dni (approved):

callebtc:
This is intentional to keep backwards compatibility for old configs that called it `CLightningWallet`.

prusnak:
Reworked to add comment why this alias is needed.

dni (approved):

PR #2066, opened 2023-10-26
make example paths in .env.example less confusing

prusnak:


.env.example
```diff
@@ -120,15 +119,15 @@ LNBITS_KEY=LNBITS_ADMIN_KEY
...
+LND_GRPC_CERT="/home/bob/.lnd/tls.cert"
+LND_GRPC_MACAROON="/home/bob/.lnd/data/chain/bitcoin/mainnet/admin.macaroon"  # or HEXSTRING
 # To use an AES-encrypted macaroon, set
 # LND_GRPC_MACAROON="eNcRyPtEdMaCaRoOn"
 
 # LndRestWallet
 LND_REST_ENDPOINT=https://127.0.0.1:8080/
-LND_REST_CERT="/home/bob/.config/Zap/lnd/bitcoin/mainnet/wallet-1/data/chain/bitcoin/mainnet/tls.cert"
-LND_REST_MACAROON="/home/bob/.config/Zap/lnd/bitcoin/mainnet/wallet-1/data/chain/bitcoin/mainnet/admin.macaroon or HEXSTRING"
+LND_GRPC_CERT="/home/bob/.lnd/tls.cert"
```
dni, line 129: i guess this should be `LND_REST_CERT` :)

prusnak: fixed in 7461f5505846bb69e3858f0957e03ed747ff847b

dni (approved):

dni (approved):

motorina0:
Comment (`#`) in the middle of the line is not interpreted as commment:

image

Though [the spec says](https://pypi.org/project/python-dotenv/) it should be:
### File format

```
... Values can be followed by a comment. ...
```

prusnak:
> Comment (`#`) in the middle of the line is not interpreted as commment:

Hmm, I remember explicitly testing for that and it worked for me. Will try again.

prusnak:
I was testing with dotenv/environs directly and it works:

```
$ python3 
Python 3.11.5 (main, Aug 24 2023, 15:09:45) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from dotenv import dotenv_values
>>> env = dotenv_values(".env.example")
>>> env["CORELIGHTNING_REST_MACAROON"]
'/path/to/clnrest/access.macaroon'
>>> env["LND_GRPC_MACAROON"]
'/home/bob/.lnd/data/chain/bitcoin/mainnet/admin.macaroon'
>>> env["LND_REST_MACAROON"]
'/home/bob/.lnd/data/chain/bitcoin/mainnet/admin.macaroon'
```

How exactly are the values from `.env` loaded to the settings classes? It seems they are using some different way?

prusnak:
Also I tried `LNBITS_SITE_TITLE="LNbits X" # testiing` and this worked just fine, so maybe this is an issue specific only to some fields? /me puzzled

motorina0:
> Also I tried `LNBITS_SITE_TITLE="LNbits X" # testiing` and this worked just fine, so maybe this is an issue specific only to some fields? /me puzzled

If `LNBITS_ADMIN_UI=true` then the values are loaded from the DB. Only the `EnvSettings` values are still loaded from the `.env` file in this case.

prusnak:
I am not using Admin UI

prusnak:
I also tried the Settings class (which is filled in via `pydantic`):

```
$ cat .env | grep LNBITS_SITE_TITLE
LNBITS_SITE_TITLE="LNbits X" # foo

$ python3
>>> from lnbits.settings import Settings 
>>> x=Settings()
>>> x.lnbits_site_title
'LNbits X'
```

There is no `foo` in title.

motorina0 (approved):

Issue #2031, opened 2023-10-13
[BUG] Web rounding milisatoshis up


AugustoResende:
**Describe the bug**
balance: 586 in API shows as 1 satoshi in web, but this is not true, and if i try to pay another lnbits invoice 1 sat (zero fee, same node), it fails because no sufficient balance.

**To Reproduce**

1. Deposit any amount
2. Pay anything cheap that gives you milisatoshis (because of fees)
3. Create another lnbits wallet
4. Transfer everything but 1 sat deducted
5. See response of wallet request return an msats amount less than 1 satoshi and interface shows 1 full satoshi.

**Expected behavior**
Show 0 satoshis or show the milisatoshis.

**Screenshots**
![Captura de tela de 2023-10-13 16-08-12](https://github.com/lnbits/lnbits/assets/10731727/66f81ff3-c104-4064-b6d4-a3f36607ea1b)

prusnak:
I think this is the fix (untested yet):

``` patch
diff --git a/lnbits/core/static/js/wallet.js b/lnbits/core/static/js/wallet.js
index d72996bf5..a80505fcf 100644
--- a/lnbits/core/static/js/wallet.js
+++ b/lnbits/core/static/js/wallet.js
@@ -791,7 +791,7 @@ new Vue({
     },
     fetchBalance: function () {
       LNbits.api.getWallet(this.g.wallet).then(response => {
-        this.balance = Math.round(response.data.balance / 1000)
+        this.balance = Math.floor(response.data.balance / 1000)
         EventHub.$emit('update-wallet-balance', [
           this.g.wallet.id,
           this.balance
diff --git a/lnbits/static/js/base.js b/lnbits/static/js/base.js
index 0b676e3dd..73e61271d 100644
--- a/lnbits/static/js/base.js
+++ b/lnbits/static/js/base.js
@@ -196,7 +196,7 @@ window.LNbits = {
         currency: data.currency
       }
       newWallet.msat = data.balance_msat
-      newWallet.sat = Math.round(data.balance_msat / 1000)
+      newWallet.sat = Math.floor(data.balance_msat / 1000)
       newWallet.fsat = new Intl.NumberFormat(window.LOCALE).format(
         newWallet.sat
       )
```

I will wait until https://github.com/lnbits/lnbits/pull/1964 is merged tomorrow (it reorganizes the file structure), then will apply this change, test and send a PR.

prusnak:
[...]

PR #2070, opened 2023-10-27
use floor() instead of round() to convert msat to sat in frontend

prusnak:
I haven't tested this yet, but this hopefully fixes https://github.com/lnbits/lnbits/issues/2031


dni:
screenshot-1698410051

not working for me, subsat value `950` but still showing 1 sat on the balance, i stepped through with the debugger, and it seems the activeBalance value passed is 0. i dont really have a clue where the 1 comes in.

motorina0:
tACK

- api call:
  - image


- UI on this branch:
  - image
- UI on dev branch:
  -  image

motorina0 (approved):

prusnak:
@motorina0 any idea why this works for you and not for @dni ?

@dni maybe you used the cashed version of the javascript and you didn't test the new code?

prusnak:
Rebased on top of current dev and force pushed.

motorina0:
> @motorina0 any idea why this works for you and not for @dni ?

I only had one issue while testing: in `Manage Server` > `Theme` the currency was set to `sat` instead of `sats` (from previous tests) and the formatting in the front end was different.

image


For the `millisats` value I have simply changed the value in my local DB:
```sql
update apipayments set amount = 1750 where checking_id = '...'
```

Note: 
- a value of `1750` (/1000) will been rounded to `2`.
- a value of `1050` (/1000) will been rounded to `1`, not to `2`.

dni:
the issue was, that 750 millisats should be rounded to 0

dni:
same. 
screenshot-1699252009

dni:
ok, i figured it out, it wasnt actually coming from the `/wallet` api call, it somewhere set in the `this.g.wallet`, i dont even know where it comes from, it seems extremely complicated for me how the wallet info is fetched. 

but after setting the milisats in the db, it works. thanks @motorina0

dni (approved):

Issue #2045, opened 2023-10-18
[BUG] Fronend does not poll invoice state if backend does not support `paid_invoices_stream`


callebtc:
**Describe the bug**
Create invoice with a backend that does not support `paid_invoices_stream` like LNPay, OpenNode, ... and pay it. The frontend does not recognize the incoming payment. Browser tab needs refresh. 

It would be better if the frontend knew whether that method is available or not and if not, use the old polling mechanism for the invoice state if the stream is not available.

jackstar12:
What about emulating the behaviour for wallets that dont support it? The base wallet could provide such implementation which would be overriden by wallet that support it. 

Something like:
```python
    async def paid_invoices_stream(self) -> AsyncGenerator[str, None]:
        while True:
            for invoice in self.pending_invoices:
                try:
                    status = await self.get_invoice_status(invoice)
                    if status.paid:
                        yield invoice
                        self.pending_invoices.remove(invoice)
                except Exception as exc:
                    logger.error(f"could not get status of invoice {invoice}: '{exc}'")
            await asyncio.sleep(5)
```
Where `pending_invoices` would be a list that is appended to when `create_invoice` is being called.

prusnak:
> Something like:

Sounds nice! 

@jackstar12 will you come up with a PR that implements this?

hdlslm:
i'm running an lnbits instance with CoreLightningRestWallet on the backend (separate node) and i'm getting "| ERROR | Error in paid_invoices_stream:" is this issue related to my error? thanx in advance!!

prusnak:
> i'm running an lnbits instance with CoreLightningRestWallet on the backend (separate node) and i'm getting "| ERROR | Error in paid_invoices_stream:" is this issue related to my error? thanx in advance!!

No. Please open a new issue.

motorina0:
Done here: https://github.com/lnbits/lnbits/pull/2071
Thank you @jackstar12

PR #2071, opened 2023-10-28
feat: emulate behaviour of paid_invoices_stream for wallets that dont support it

jackstar12:
Possible solution to #2045.

Instead of optionally polling on the frontend we can just emulate the behaviour of the invoices stream by regulary polling the currently pending invoices. It shouldnt be too big of a difference in regards to rate limiting because whether the browser or the invoice task polls doesnt matter.

### DIFF
--- a/lnbits/wallets/base.py
+++ b/lnbits/wallets/base.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
+import asyncio
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, AsyncGenerator, Coroutine, NamedTuple, Optional
+
+from loguru import logger
 
 if TYPE_CHECKING:
     from lnbits.nodes.base import Node
@@ -93,6 +96,9 @@ class Wallet(ABC):
 
     __node_cls__: Optional[type[Node]] = None
 
+    def __init__(self) -> None:
+        self.pending_invoices: list[str] = []
+
     @abstractmethod
     async def cleanup(self):
         pass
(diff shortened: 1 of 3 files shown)


prusnak:
I think we could also rework `FakeWallet` and `VoidWallet` to use the same mechanism.

callebtc:
Very cool but we discussed this approach before and concluded that this would not work in practice as soon as a few people use an instance.

> It shouldnt be too big of a difference in regards to rate limiting because whether the browser or the invoice task polls doesnt matter.

That's not true. You would end up polling every unpaid invoice ever created by everyone until expiry.

lnbits/wallets/base.py
```diff
@@ -91,9 +97,19 @@ def get_payment_status(
     ) -> Coroutine[None, None, PaymentStatus]:
         pass
 
-    @abstractmethod
-    def paid_invoices_stream(self) -> AsyncGenerator[str, None]:
-        pass
+    async def paid_invoices_stream(self) -> AsyncGenerator[str, None]:
+        while True:
+            for invoice in self.pending_invoices:
```
motorina0, line 102: this loop is non-blocking, the while loop will be CPU intensive

jackstar12: Why would it be cpu intensive? There is a sleep at the bottom.

The original queue was never put into. A better overall approach would probably be creating a task for each invoice which only queries the invoice status for a certain time and then stops, similar to the frontend.

motorina0: > Why would it be cpu intensive? There is a sleep at the bottom.

My bad. I thought the sleep is in the catch block.

motorina0 (approved):

dni (approved):

PR #2075, opened 2023-11-01
feat: logger success when wallet is connected

dni:
this should be a success log, see screenshot.

screenshot-1698851921

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -162,7 +162,7 @@ async def check_funding_source() -> None:
 
     signal.signal(signal.SIGINT, original_sigint_handler)
 
-    logger.info(
+    logger.success(
         f"✔️ Backend {WALLET.__class__.__name__} connected "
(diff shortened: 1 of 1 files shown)


lnbits/app.py
```diff
@@ -162,7 +162,7 @@ def signal_handler(signal, frame):
 
     signal.signal(signal.SIGINT, original_sigint_handler)
 
-    logger.info(
+    logger.success(
         f"✔️ Backend {WALLET.__class__.__name__} connected "
```
motorina0, line 166: why is there no `+`?

motorina0 (approved):

lnbits/app.py
dni: you mean for string concatenation? its not needed when inside `()` and if a fn has only one argument

motorina0: TIL
Thanks!

prusnak (approved):

PR #2091, opened 2023-11-08
feat: payment updates fallback

jackstar12:
closes: #2045 

This PR makes the sse endpoint return a 503 status if the wallet backend does not support sse, which the frontend detects. 

I also simplified the internal invoice structure by removing the `api_invoice_listeners` layer, which was a bit confusing to me. `SseListenersDict` also seems unnecessary if it only has one usage now.

I'm also not sure if we really need the count of listeners in the frontend, I think that could be removed aswell.

### DIFF
--- a/lnbits/core/tasks.py
+++ b/lnbits/core/tasks.py
@@ -1,5 +1,4 @@
 import asyncio
-from typing import Dict
 
 import httpx
 from loguru import logger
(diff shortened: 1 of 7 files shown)


jackstar12:
@motorina0 has there been an other fix?

dni:
i think i will incooprerate this, im working some invoice listener stuff. @jackstar12

PR #2092, opened 2023-11-10
 [FEAT] Auth, Login, OAuth, create account with username and password #1653 

motorina0:
### Summary
Add login functionality. Supported methods:
 - `user-id-only` 
    - the current (legacy) method
    - the `usr` (user id) from the url params is automatically removed
 - `username-password`
    - allow user to register * login with a username and password
    - an email is not required, but the user can configure one via SSO
 - `google-auth` - use google SSO
 - `github-auth` - use github SSO

The auth fields are documented in `.env.example`.

Based on the `AUTH_ALLOWED_METHODS` different Login/Register options will be available to the end user.

**Login Page (all options)**
image

**Login Page (no `usr` allowed)**
image

**Google SSO**:
https://github.com/lnbits/lnbits/assets/2951406/5f7ce5a7-7a9a-479b-b948-71c5cfbb39d4

On the **Account Settings** page the user can verify its email address via SSO (google and github at the moment).
image

### Test Scenarios

**Scenario 1 (backwards compatibility)**
 - access with `usr` in URL query params
 - test `Login with user ID`
 - test `Create New Wallet`
    - image

**Scenario 2**
 - Register (username & password)
 - Login with the above username and password

**Scenario 3**
 - Logout
 
**Scenario 4**
 - Login with Google
 - Login with Github

**Scenario 5 (Account Settings page)**
 - set `username` (username cannot be updated)
 - update other values (first name, last name, picture, etc)
   - image

**Scenario 6 (Change password)**
 - on `Account Settings` click `Change Password`
 - if the user was registered via SSO the the old password is not required (only the new one)
 - otherwise the old password is required

**Scenario 7 (add email)**
 - an account that was created via the `usr` or `Register` flow has no email
 - it is possible for the user to attach a validated email using SSO:
   - go to `Account Settings`
   - if no email is present then the SSO options are available
   - select one of them and authenticate
   - at the end of the flow, the user should have the validated email configured
   - emails cannot be changed or manually edited
   -  image

**Scenario 8 (token expire)**
 - set a low value (1 minute) for `AUTH_TOKEN_EXPIRE_MINUTES`
 - login and refresh after 1 min
 - the user should be redirected to the login page and a info message displayed

### Dependencies
 - `itsdangerous` is required by `from starlette.middleware.sessions import SessionMiddleware`

### Notes
Stuff that is still open but the PR can be reviewed without:
 - i18n
 - warning message on login/logout (if email not set)
 - find place for long description text on Login page
 - CI failures: nix and a regtest call


.env.example
```diff
@@ -21,6 +21,12 @@ LNBITS_RATE_LIMIT_UNIT="minute"
 LNBITS_ALLOWED_IPS=""
 LNBITS_BLOCKED_IPS=""
 
+# Auth Configurations
+SECRET_KEY="" # will default to the superuser value
+ALGORITHM = "HS256"
```
prusnak, line 26: Why would you want to make this configurable?

lnbits/core/migrations.py
```diff
@@ -393,3 +393,13 @@ async def m015_create_push_notification_subscriptions_table(db):
         );
     """
     )
+
+
+async def m016_add_username_column_to_acounts(db):
```
prusnak, line 398: `acounts` -> `accounts`

lnbits/core/templates/core/index.html
```diff
@@ -3,7 +3,6 @@
 {% endblock %} {% block page %}
 <div class="row q-col-gutter-md justify-center">
   <div class="col-12 col-md-7 col-lg-6 q-gutter-y-md">
-    {% if lnurl or LNBITS_NEW_ACCOUNTS_ALLOWED %}
```
prusnak, line 6: If you remove this and `LNBITS_NEW_ACCOUNTS_ALLOWED` is set to `false`, then this would create an empty box at the top.

prusnak (changes requested):
1. file `.coverage.Vlads-MacBook-Pro.local.79917.036134` does not belong here
2. file `lnbits/static/js/service-worker.js-e` does not belong here
3. I am not super happy this PR adds so many new dependencies. Surely we can do better?
4. Is `python-multipart` used anywhere?

motorina0:
@prusnak this is a draft PR based on https://github.com/lnbits/lnbits/pull/1653
Many things are unfinished or inherited from the previous PR.
Please wait until I take it out of draft.

.env.example
zoop-btc: I would need at least SECRET_KEY to be configurable so I can put lnbits behind an auth-proxy configured with the same SECRET_KEY for example.

prusnak: Yeah, I was asking about `ALGORITHM` field.

lnbits/decorators.py
```diff
@@ -306,9 +306,9 @@ async def _get_account_from_token(access_token):
     try:
         payload = jwt.decode(access_token, settings.auth_secret_key, "HS256")
         if "sub" in payload and payload.get("sub"):
-            return await get_account_by_username(payload.get("sub"))
+            return await get_account_by_username(str(payload.get("sub")))
```
zoop-btc, line 309: "sub" in a jwt token is usually the user-id and not a username

motorina0: Yes, I'm still in doubt here due to this scenario:
 1. `user_id` (`usr`) can be used as login method (current state, but for backwards compatibility it will still be allowed)
 1. sometimes `access-tokens` are shared, lost, stolen, etc. That is why an access token has an `expire` value (becomes invalid after a while and a new one must be obtained)

If the `user_id` is present in the token and **both** `1.` and `2.` are true then it is not possible to revoke that token anymore.

lnbits/core/helpers.py
```diff
@@ -58,14 +60,13 @@ async def stop_extension_background_work(ext_id: str, user: str):
     async with httpx.AsyncClient() as client:
         try:
             url = f"http://{settings.host}:{settings.port}/{ext_id}/api/v1?usr={user}"
```
dni, line 62: is this `?usr=` needed here?

lnbits/core/models.py
```diff
@@ -79,14 +79,25 @@ class WalletTypeInfo:
     wallet: Wallet
 
 
+class UserConfig(BaseModel):
+    email_verified: Optional[bool] = False
+    first_name: Optional[str] = None
+    last_name: Optional[str] = None
+    display_name: Optional[str] = None
+    picture: Optional[str] = None
+    provider: Optional[str] = "lnbits"  # auth provider
```
dni, line 88: is there a reason why you added this into a separate model/object `config:` . i would rather have it flat in the user table/model so you could do a simple `SELECT * from users WHERE email_verified=true`. the less json in the db the better, imho

pyproject.toml
```diff
@@ -42,6 +43,12 @@ secp256k1 = "0.14.0"
 environs = "9.5.0"
 # needed for scheduler extension
 python-crontab = "3.0.0"
+python-jose = "^3.3.0"
+passlib = "^1.7.4"
+types-python-jose = "^3.3.4.8"
+types-passlib = "^1.7.7.13"
```
dni, line 49: types can be dev dependencies

motorina0: done!

lnbits/core/models.py
motorina0: > i would rather have it flat in the user table/model
- if there is a scenario where one of this fields is required for querying then it can be extracted. At the moment there is none so we should not over-generalize

> the less json in the db the better, imho
- if the data is not needed for querying then it is fine in a JSON, no need to have data migrations each time a new field is added to the model

lnbits/core/helpers.py
motorina0: Good point. 
I've kept `usr` in case the install/uninstall endpoints are called via the API and no `access-token` is provided. One such scenario is the `lnbits-cli` calls.

jackstar12:
I just started from a fresh db, ran `poetry run lnbits-cli superuser-url` and opened the link, got the following:

image

It immediately redirects to the /wallet url without the `usr` param, but adding the `wal` param. If I know manually add the `usr` back to the url it works.

dni (approved):

prusnak:
Question: @motorina0 When `LNBITS_ALLOW_NEW_ACCOUNTS` is set to `False`:
1. does it prevent creation of new users with username and password (`username-password` method)?
2. does it prevent creation of new users via Google SSO and Github SSO (`google-auth` and `github-auth` methods)?

motorina0:
> Question: @motorina0 When `LNBITS_ALLOW_NEW_ACCOUNTS` is set to `False`:
> 
> 1. does it prevent creation of new users with username and password (`username-password` method)?
> 2. does it prevent creation of new users via Google SSO and Github SSO (`google-auth` and `github-auth` methods)?

- 1: yes
- 2: yes

lnbits/helpers.py
```diff
@@ -166,3 +170,20 @@ def update_query(table_name: str, model: BaseModel, where: str = "WHERE id = ?")
     """
     query = ", ".join([f"{field} = ?" for field in model.dict().keys()])
     return f"UPDATE {table_name} SET {query} {where}"
+
+
+def is_valid_email_address(email: str) -> bool:
+    email_regex = r"\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,7}\b"
```
prusnak, line 176: This is too strict, there are new TLDs such as `.foundation` which do not match it - the TLD can be maximum [63 chars](https://stackoverflow.com/questions/9238640/how-long-can-a-tld-possibly-be).
Also I think we do not need boundaries since we are using the fullmatch.

```suggestion
    email_regex = r"[A-Za-z0-9\._%+-]+@[A-Za-z0-9\.-]+\.[A-Za-z]{2,63}"
```

lnbits/helpers.py
```diff
@@ -166,3 +170,20 @@ def update_query(table_name: str, model: BaseModel, where: str = "WHERE id = ?")
...
+    return re.fullmatch(email_regex, email) is not None
+
+
+def is_valid_username(username: str) -> bool:
+    username_regex = r"\b(?=[a-zA-Z0-9._]{2,20}$)(?!.*[_.]{2})[^_.].*[^_.]\b"
```
prusnak, line 181: I am clueless. Where does this come from and why is it so complicated?

prusnak (changes requested):

lnbits/helpers.py
motorina0: Adapted from here: https://stackoverflow.com/questions/12018245/regular-expression-to-validate-username

```
- allowed characters: a-zA-Z0-9._
- username is 2-20 characters long
- no __ or _. or ._ or .. inside
- no _ or . at the beginning
- no _ or . at the end
```

PR #2096, opened 2023-11-13
hotfix toolbar title

talvasconcelos:
fix button wrapping on toolbar

button

Make the button (when existing) independent. For example, the QR button on TPoS extension is inside the logo button.


motorina0:
Not sure what is the "Expected Result" or the "Actual Result"

talvasconcelos:
> Not sure what is the "Expected Result" or the "Actual Result"

On public pages, like TPoS for example, we have now a share button, that displays a QR dialog. That share button is wrapped inside the logo button. The PR just moves the start of the button tag `<q-btn ... >` inside the block, so it's not wrapping the children, when in TPoS i add a button to the `{% block toolbar_title %}` it's not wrapped inside the logo button! it's a UI nitpick...

motorina0 (approved):

arcbtc (approved):

PR #2098, opened 2023-11-13
FakeWallet: provide payment preimage

callebtc:
FakeWallet used to create invalid payment secrets. Now it does it in a bolt11 compliant way. 

FakeWallet now also keeps the payment secrets and returns them upon a successful payment. This should fix the bug reported by Alby users who complained that they were only receiving "00000...".

FakeWallet can now also report the status of a paid invoice by keeping a set of them.


dni (approved):

arcbtc (approved):

prusnak (approved):
utACK

PR #2112, opened 2023-11-17
Added extra field to wallets db

arcbtc:
I need for the goofy animations pr, but I imagine is useful for things like people using LNbits in their stack, if apipayments, wallets and accounts have "extra" fields.
Note: apipayments already has and accounts will get "extra" with Vlads login pr.

### DIFF
--- a/lnbits/core/migrations.py
+++ b/lnbits/core/migrations.py
@@ -393,3 +393,7 @@ async def m015_create_push_notification_subscriptions_table(db):
         );
     """
     )
+
+
+async def m016_add_extra_wallets(db):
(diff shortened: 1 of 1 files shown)


motorina0:
Following the `Payment` model, the `Wallet` model should also have an `extra` field and persist/restore it.

```py
class Payment(FromRowModel):
    ...
    extra: Dict = {}
    ...

    @classmethod
    def from_row(cls, row: Row):
        return cls(
            ...
            extra=json.loads(row["extra"] or "{}"),
            ...
```

dni (commented):
i dont think this is useful, we should go for  columns instead of a generic extra field.

PR #1897, opened 2023-08-23
feat: add Breez SDK wallet

callebtc:
Set config in `.env`:

```
LNBITS_BACKEND_WALLET_CLASS=BreezSdkWallet

BREEZ_GREENLIGHT_SEED = "..."
BREEZ_API_KEY = "..."
BREEZ_GREENLIGHT_INVITE_CODE = "..."
```

PR #2117, opened 2023-11-23
feat: add libwally for liquid support

dni:
this is needed for liquid support in the new boltz extension, sadly i could not get it to work with embit library, liquid support is not quite ready there and no documentation or tests available

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -38,6 +38,7 @@
           meta.rev = self.dirtyRev or self.rev;
           overrides = pkgs.poetry2nix.overrides.withDefaults (final: prev: {
             ruff = prev.ruff.override { preferWheel = true; };
+            wallycore = prev.wallycore.override { preferWheel = true; };
             fastapi = prev.fastapi.overridePythonAttrs (old: {
               postPatch = ''
                 substituteInPlace pyproject.toml \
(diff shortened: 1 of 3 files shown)


prusnak:
I think we should start using optional group dependencies for the project.
Especially when we are introducing binary dependencies. Same goes for https://github.com/lnbits/lnbits/pull/1897

That is, add this dependency using `poetry add wallycore --group=liquid` and then requiring user to install via `poetry install -E liquid` instead of just `poetry install`.

Also boltz should import wallycore conditionally, so everything works just fine (modulo liquid support) when wallycore is not installed.

What do you think, @dni?

motorina0:
> We then require user to install via `poetry install -E liquid` instead of just `poetry install`. Or is this too much to ask from admins?

I think this is a very good solution.
If the admin are planning to run the `blotz` extension then they are technical enough to use the `-E` flag.

prusnak:
I pushed the poetry extra change to commit 30137d596d520a26f7551def271a7290cfd0bbef

dni:
if we want to do this, i think we should do that for the other dependencies aswell.

prusnak (approved):

motorina0 (approved):

PR #2120, opened 2023-11-24
fix: redirect for all paths of the extension (not only for `/api`)

motorina0:
### Summary
There is a middleware called `InstalledExtensionMiddleware` and this middleware is great! But it had a bug.
A middleware (in general) intercepts all HTTP calls and does some pre/post processing on the request.

The main job of this middleware is to:
 - check if an extension is upgraded
 - if so, then redirect the calls to the latest upgraded version by re-writing the HTTP path

Static resources do not require redirect so in the old version only the paths for the `/api` calls were redirected.

However, the public pages (like http://127.0.0.1:5000/tpos/9BKt3SCEtLMmwXSh6sxjsp) are not `/api` and are not static resources, therefore they were not redirected (this is the bug).

The changes in this PR simplify the middleware by:
 - explicitly check if the path is a static resource (and skip redirect)
 - explicitly check if the extension is upgraded. If true then redirect for all paths (except `static`)

### Testing Scenarios
**Scenario 1 (regression bugs):**
- smoke test the core functionality (wallet, admin) pages. All HTTP calls go trough the middleware.
- smoke test a random extension

**Scenario 2 (ext upgrade):**
- install an extension. A good example is TPOS (version 0.4.3)
- create some data in the extension
- upgrade the extension (TPOS 0.5.1)
- check that the UI is upgraded both on the main page `/tpos` and on the public pages `/tpos/{some_id}`
  - previously the public page failed

**Scenario 3 (static fiels)**
- the TPOS ext serves no static files
- install `watchonly` extension and upgrade it (it has many static files)
- check that it works fine

**Notes**
When an extension is installed all the files (`.py`, `.html`, `.js`, etc) in `lnbits/extensions/myext` are replaced by the new version. 
For the static resources, the client will be served the new (updated) files. These resources are loaded via [static_url_for()](https://github.com/lnbits/lnbits/blob/dev/lnbits/helpers.py#L32) function).  The `"static"` path for static files is configured [here](https://github.com/lnbits/lnbits/blob/dev/lnbits/app.py#L84-L87).
On the other hand the logic in the python modules is not automatically reloaded when the files are updated, and the redirecting is needed.

### DIFF
--- a/lnbits/middleware.py
+++ b/lnbits/middleware.py
@@ -25,7 +25,8 @@ class InstalledExtensionMiddleware:
         self.app = app
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        if "path" not in scope:
+        path = scope.get("path", "/")
+        if path == "/":
             await self.app(scope, receive, send)
             return
 
--- a/lnbits/middleware.py
+++ b/lnbits/middleware.py
@@ -29,14 +29,7 @@ class InstalledExtensionMiddleware:
             await self.app(scope, receive, send)
             return
 
-        path_elements = scope["path"].split("/")
-        if len(path_elements) > 2:
-            _, path_name, path_type, *rest = path_elements
-        else:
-            _, path_name = path_elements
-            path_type = None
-            rest = []
-
+        path_name, *rest = [p for p in scope["path"].split("/") if p]
         headers = scope.get("headers", [])
 
         # block path for all users if the extension is disabled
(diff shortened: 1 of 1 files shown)


dni (approved):
ntACK

prusnak (approved):
tested ACK

I tested all 3 scenarios and they work as expected.

I also confirm I was able to reproduce the issue with the latest in `dev` and that this PR fixes the issue.

PR #2119, opened 2023-11-24
i18n: add Korean language

prusnak:
Supersedes https://github.com/lnbits/lnbits/pull/2118

PR #2121, opened 2023-11-24
 feat: manage extensions navbar

dni:
screenshot-1700858476
screenshot-1700858471

create `Manage` spacer in sidebar and add manage extensions to it


prusnak (changes requested):
1. the PR needs a `make bundle` commit

2. This change does not work correctly if admin UI is disabled (`LNBITS_ADMIN_UI=false`)
Since there is no management section, there is also no manage extensions button.

Screenshot 2023-11-24 at 23 17 21

before the change it looks like this:

Screenshot 2023-11-24 at 23 18 27

dni:
i actually did that it was visible for non admins? maybe you used the bundle?
if you look at my screenshot above

dni:
bundled

prusnak:
1. I pushed the `make bundle` commit which was missing
2. I still do not see the extension menu.

Steps to reproduce (2):
1. copy `.env.example` to `.env`
2. start LNbits - create new user - copy usr id
3. add usr id to `LNBITS_ADMIN_USERS` in `.env`
4. restart LNbits
5. observe there is no extension menu

prusnak:
I discovered duplicit strings in localization related to the manage change, I also added Korean translation from https://github.com/lnbits/lnbits/pull/2119 which I fixed to use the new manage strings too.

prusnak (approved):

motorina0 (approved):
tACK

PR #2123, opened 2023-11-26
use awk to bump CACHE_VERSION (drop sed usage)

prusnak:
since we are already using awk, we might as well use it exclusively and drop sed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,8 +97,8 @@ bundle:
 	npm run vendor_bundle_js
 	npm run vendor_minify_js
 	# increment serviceworker version
-	sed -i -e "s/CACHE_VERSION =.*/CACHE_VERSION = $$(awk '/CACHE_VERSION =/ { print 1+$$4 }' lnbits/static/js/service-worker.js)/" \
-		lnbits/static/js/service-worker.js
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -97,8 +97,8 @@ bundle:
 	npm run vendor_bundle_js
 	npm run vendor_minify_js
 	# increment serviceworker version
-	sed -i -e "s/CACHE_VERSION =.*/CACHE_VERSION = $$(awk '/CACHE_VERSION =/ { print 1+$$4 }' lnbits/static/js/service-worker.js)/" \
-		lnbits/static/js/service-worker.js
+	awk '/CACHE_VERSION =/ {sub(/[0-9]+$$/, $$NF+1)} 1' lnbits/static/js/service-worker.js > lnbits/static/js/service-worker.js.new
+	mv lnbits/static/js/service-worker.js.new lnbits/static/js/service-worker.js
```
dni, line 101:
```suggestion
	awk -i inplace '/CACHE_VERSION =/ {sub(/[0-9]+$$/, $$NF+1)} 1' lnbits/static/js/service-worker.js
```

prusnak: -i inplace does not work with all versions of awk

Actually only modern distros have this option (Ubuntu 20.04 doesn't and macOS neither)

dni (approved):

jackstar12 (approved):

PR #2125, opened 2023-11-27
chore: update to version 0.12.0

dni:


motorina0 (approved):

prusnak:
How do we now whether the next version is 0.12.0 or 0.11.3?

I think we should postpone merging of this PR until it is obvious?

prusnak (approved):
Since we now have the new Login thingie in dev, we should update the version to 0.12.0 in dev. ACK

PR #2136, opened 2023-11-30
feat: add currency amount to lnurl/lnaddress payments

### LINKED ISSUE #2135: [Feature request] make lnurl payable with a fiat amount
https://t.me/decentraleyes: `when i pay xyz@mydomain.com via pASTE REQUEST/READ  it asks for sats however i want to be able to enter a FIAT amt.
can someone point me to the area of code that i could chane to make that happen? or could i Pay someone to make that change? If someone needs 1000 pesos its too darn hard to figure out how many sats taht is`

there is a bounty. :)

dni:
closes #2135


prusnak:
Hah, `checkbundle` CI check worked! 🎉

lnbits/core/templates/core/wallet.html
```diff
@@ -660,7 +660,7 @@ <h6 v-else class="q-my-none">
               <p v-if="parse.lnurlpay.fixed" class="q-my-none text-h6">
                 <b>{{ parse.lnurlpay.domain }}</b> is requesting {{
                 parse.lnurlpay.maxSendable | msatoshiFormat }}
-                {{LNBITS_DENOMINATION}}
+                <span v-text="parse.data.unit"></span>
```
prusnak, line 663: Revert this line change

When I change the currency, the dialog claims the LNaddress requests amount in that currency (between 1 and 1,000,000 CZK) which is not the case. The title of the header window should be still kept in sats.

Screenshot 2023-11-30 at 9 52 06

lnbits/core/templates/core/wallet.html
```diff
@@ -672,7 +672,7 @@ <h6 v-else class="q-my-none">
                 between
                 <b>{{ parse.lnurlpay.minSendable | msatoshiFormat }}</b> and
                 <b>{{ parse.lnurlpay.maxSendable | msatoshiFormat }}</b>
-                {% endraw %} {{LNBITS_DENOMINATION}} {% raw %}
+                <span v-text="parse.data.unit"></span>
```
prusnak, line 675: Same here

prusnak (changes requested):

talvasconcelos (approved):
Seems good

prusnak:
When I set EUR in the dialog once and then go to pay for another invoice, the unit is set to EUR again. Maybe we should reset it to `sat` every time the dialog is open?

prusnak:
> When I set EUR in the dialog once and then go to pay for another request, the unit is set to EUR again. Maybe we should reset it to `sat` every time the dialog is open?
> 
> But if you say that it is preferred that way, to keep the preference persistent, I am OK with that as well.

@dni What is your take on this? If you want to keep it persistent, I will approve and merge.

prusnak:
Resolved conflicts, rebased on top of the current `dev`.

dni:
hey @prusnak! i look it at again and after i tried it i thought that when i selected it once, i probably want to keep going in that currency and not have to select it again at least for the current pageload. i dont think not reseting to sat is a problem.

prusnak (approved):

PR #2139, opened 2023-11-30
feat: usermanager

dni:
add usermanager to admin_ui. show user accounts and wallets with transaction count and balance

screenshot-1702563067

- adds `manage user section`
- move topup to users
- enable / disable admins

PR #2140, opened 2023-11-30
feat: add group_by to fetch_page

dni:
feature is required by #2139

### DIFF
--- a/lnbits/db.py
+++ b/lnbits/db.py
@@ -179,7 +179,7 @@ class Connection(Compat):
         values: Optional[List[str]] = None,
         filters: Optional[Filters] = None,
         model: Optional[Type[TRowModel]] = None,
-        group_by: Optional[str] = "",
+        group_by: Optional[str] = None,
     ) -> Page[TRowModel]:
         if not filters:
             filters = Filters()
@@ -197,7 +197,7 @@ class Connection(Compat):
             f"""
             {query}
             {clause}
-            {group_by}
+            {group_by or ""}
             {filters.order_by()}
             {filters.pagination()}
             """,
@@ -211,7 +211,7 @@ class Connection(Compat):
                     SELECT COUNT(*) FROM (
                         {query}
                         {clause}
-                        {group_by}
+                        {group_by or ""}
                     ) as count
                     """,
                     parsed_values,
--- a/lnbits/db.py
+++ b/lnbits/db.py
@@ -179,7 +179,7 @@ class Connection(Compat):
         values: Optional[List[str]] = None,
         filters: Optional[Filters] = None,
         model: Optional[Type[TRowModel]] = None,
-        group_by: Optional[str] = None,
+        group_by: Optional[str] = "",
     ) -> Page[TRowModel]:
         if not filters:
             filters = Filters()
@@ -192,8 +192,6 @@ class Connection(Compat):
             ):
                 raise ValueError("Value for GROUP BY is invalid")
             group_by = f"GROUP BY {group_by}"
-        else:
-            group_by = ""
 
         rows = await self.fetchall(
             f"""
--- a/lnbits/db.py
+++ b/lnbits/db.py
@@ -179,25 +179,27 @@ class Connection(Compat):
         values: Optional[List[str]] = None,
         filters: Optional[Filters] = None,
         model: Optional[Type[TRowModel]] = None,
-        group_by: Optional[str] = None,
+        group_by: Optional[List[str]] = None,
     ) -> Page[TRowModel]:
         if not filters:
             filters = Filters()
         clause = filters.where(where)
         parsed_values = filters.values(values)
 
-        if group_by:
-            if not re.fullmatch(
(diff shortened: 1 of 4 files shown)


lnbits/db.py
```diff
@@ -173,16 +173,19 @@ async def fetch_page(
         values: Optional[List[str]] = None,
         filters: Optional[Filters] = None,
         model: Optional[Type[TRowModel]] = None,
+        group_by: Optional[str] = None,
     ) -> Page[TRowModel]:
         if not filters:
             filters = Filters()
         clause = filters.where(where)
         parsed_values = filters.values(values)
+        group_by = f"GROUP BY {group_by}" if group_by else ""
```
motorina0, line 182: SQL injection risk

jackstar12: cant easily be avoided here. this is stuff which orms / query builders would take care of

motorina0: a regex check of this (roughly) form should do it: `^[a-zA-Z_][a-zA-Z0-9_]*$`

prusnak: yeah, we should probably add:

``` python
assert re.fullmatch(r"[a-zA-Z_][a-zA-Z0-9_]*", group_by)
```

motorina0:
Can we please have one small test for `GROUP BY`.

dni:
added tests

lnbits/db.py
```diff
@@ -173,16 +173,26 @@ async def fetch_page(
...
         clause = filters.where(where)
         parsed_values = filters.values(values)
 
+        if group_by:
+            assert re.fullmatch(
+                r"[a-zA-Z_][a-zA-Z0-9_]*(\.[a-zA-Z_][a-zA-Z0-9_]*)?", group_by
+            ), "GROUP BY regex failed"
+            group_by = f"GROUP BY {group_by}"
+        else:
+            group_by = ""
```
motorina0, line 189: nitpick: 
 - if the param declaration is `group_by: Optional[str] = "",` then this `else` is not required anymore

tests/core/test_db_fetch_page.py
```diff
@@ -0,0 +1,67 @@
...
+        model=TestModel,
+        group_by="name",
+    )
+    assert row
+    assert row.total == 4
+
+
+@pytest.mark.asyncio
+async def test_db_fetch_page_group_by_evil(fetch_page, db):
+    with pytest.raises(AssertionError):
```
motorina0, line 62: Can the assertion message be checked? Other failing assertion can give a false positive for this test

tests/helpers.py
```diff
@@ -9,13 +9,19 @@
...
 from loguru import logger
+from pydantic import BaseModel
 from sqlalchemy.engine.url import make_url
 
 from lnbits import core
 from lnbits.db import DB_TYPE, POSTGRES
 from lnbits.wallets import get_wallet_class, set_wallet_class
 
 
+class TestModel(BaseModel):
```
motorina0, line 20: nitpick:
```suggestion
class TestPerson(BaseModel):
```

tests/core/test_db_fetch_page.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+
+    assert row
+    assert row.total == 5
+    assert len(row.data) == 5
+
+
+@pytest.mark.asyncio
+async def test_db_fetch_page_group_by(fetch_page, db):
+    row = await db.fetch_page(
```
motorina0, line 51: I get the following (expected) error on local `make test`. Why does the CI pass?
```
E       sqlalchemy.exc.ProgrammingError: (psycopg2.errors.GroupingError) column "test_db_fetch_page.id" must appear in the GROUP BY clause or be used in an aggregate function
E       LINE 2:             select * from test_db_fetch_page
E                                  ^
E       
E       [SQL: 
E                   select * from test_db_fetch_page
E                   
E                   GROUP BY name
E                   
E                   
E                   ]
```

tests/core/test_db_fetch_page.py
prusnak: Addressed in 3d51a30ccc17f32d786d99c9895861681f8385d5

tests/helpers.py
dni: i disagree, it is meant as a generic model not a person

tests/core/test_db_fetch_page.py
dni: couldn't reproduce what sqllite version do you have?

prusnak (approved):

lnbits/db.py
```diff
@@ -179,16 +179,27 @@ async def fetch_page(
         values: Optional[List[str]] = None,
         filters: Optional[Filters] = None,
         model: Optional[Type[TRowModel]] = None,
+        group_by: Optional[List[str]] = None,
```
motorina0, line 182: The users need to know that they have to add an aggregator function (`max`, `count`, etc) when using `group_by`. Otherwise the query execution will fail.

Can this requirement be documented somewhere?

motorina0 (approved):

PR #2141, opened 2023-11-30
feat: add password to `<lnbits-dynamic-fields>`

dni:


motorina0 (approved):

motorina0:
LGTM
Nitpick: the html is inside a string therefore `make format` does not touch it. Can you please format it independently.

PR #2151, opened 2023-12-05
bug: cln paymentstate check on error

callebtc:
Payment state check should *only* return False if the payment 100% failed. Uncheckable entries should not enter the db in the first place, which we will be working on next.

### DIFF
--- a/lnbits/core/models.py
+++ b/lnbits/core/models.py
@@ -280,11 +280,18 @@ class Payment(FromRowModel):
                 f"expired {expiration_date}"
             )
             await self.delete(conn)
+        # wait at least 15 minutes before deleting failed outgoing payments
         elif self.is_out and status.failed:
-            logger.info(
-                f"Deleting outgoing failed payment {self.checking_id}: {status}"
-            )
-            await self.delete(conn)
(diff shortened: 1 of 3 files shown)


prusnak (dismissed):
CI failure: https://github.com/lnbits/lnbits/actions/runs/7099157921/job/19322647760?pr=2151

dni (changes requested):
i am scared, i think also the debug log when deleting was removed,and that actually saved us this time discovering the issue at all. so at least the log should be readded.

prusnak:
> i am scared, i think also the debug log when deleting was removed

It was re-added in 7dfbabf

lnbits/core/models.py
```diff
@@ -223,16 +223,10 @@ async def check_status(
             await self.delete(conn)
         # wait at least 15 minutes before deleting failed outgoing payments
```
dni, line 224:
```suggestion
```

dni (approved):

prusnak:
Squashed and rebased

talvasconcelos (approved):

motorina0 (approved):

PR #2154, opened 2023-12-05
refactor: move tinyurl to its own router

dni:
small little refactor getting to a cleaner `views/api.py`

PR #2155, opened 2023-12-05
refactor: move webpush to its own router

dni:
refactor to make `views/api.py` cleaner

### DIFF
--- a/lnbits/core/views/webpush_api.py
+++ b/lnbits/core/views/webpush_api.py
@@ -1,0 +1,60 @@
+import base64
+import json
+from http import HTTPStatus
+from urllib.parse import unquote, urlparse
+
+from fastapi import (
+    APIRouter,
+    Depends,
+    Request,
+)
+
+from lnbits.core.models import (
+    CreateWebPushSubscription,
+    WebPushSubscription,
+)
+from lnbits.decorators import (
+    WalletTypeInfo,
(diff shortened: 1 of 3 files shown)


prusnak (approved):
utACK

lnbits/core/views/webpush_api.py
```diff
@@ -0,0 +1,60 @@
...
+    require_admin_key,
+)
+
+from ..crud import (
+    create_webpush_subscription,
+    delete_webpush_subscription,
+    get_webpush_subscription,
+)
+
+webpush_router = APIRouter(prefix="/api/v1/webpush", tags=["webpush"])
```
motorina0, line 27: **Nitpick**: a bit inconsistent reported to `tinyurl_router` (which does not have a prefix):
```py
tinyurl_router = APIRouter()
```

Both `tinyurl_router` and `webpush_router` should [either have or not have a `prefix` ](https://github.com/lnbits/lnbits/pull/2154/files#diff-f2fbe70e0e85aafee5ab16b5ffa59e11acbafd33d2e7a606f94ca745018dd722R22)

I like with `prefix` better: normally the version is updated for all the paths in the file.

motorina0: oh, I see 👍

motorina0 (approved):
utACK

PR #2156, opened 2023-12-05
chore: update lock file and poetry 1.7.0 in CI

dni:
fixed 2 github security alerts.

https://github.com/lnbits/lnbits/security/dependabot/40
https://github.com/lnbits/lnbits/security/dependabot/37

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1689068808,
-        "narHash": "sha256-6ixXo3wt24N/melDWjq70UuHQLxGV8jZvooRanIHXw0=",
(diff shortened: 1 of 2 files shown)


dni:
i am fine with that !

prusnak:
I found the root cause of the nix build failure. There are two issues and both need some time until they are fixed in upstream:
- https://github.com/nix-community/poetry2nix/pull/1447 (needs to be merged)
- https://github.com/NixOS/nixpkgs/pull/270452 (already merged, but it will take some time until staging branch is merged to master)

prusnak:
I pushed another commit which hopefully fixes the build on nix 🎉 

It certainly fixes the build for me locally.

PR #2157, opened 2023-12-06
add fiat balance to wallet

### LINKED ISSUE #2149: [Feature request] Balance of wallet in fiat amount too
LNBits has "Fiat tracking" for wallets. But balance of wallet (top amount in "sats") is displayed always in sats only
May be if a wallet has setting of fiat currency to display there a balance in fiat amount too (just for a set fiat currency only or for default global fiat currency if a wallet has not a set fiat unit)?

I as user would like to see there fiat amount too. I think it will be a good UX...

talvasconcelos:
also fixed payments table fiat denomination for consistency

Closes #2149

### DIFF
--- a/lnbits/core/templates/core/wallet.html
+++ b/lnbits/core/templates/core/wallet.html
@@ -55,6 +55,13 @@
                 </q-popup-edit>
               </q-btn>
             </h3>
+            <div v-if="g.wallet.currency">
+              <span
+                class="text-h5 text-italic"
+                v-text="formattedFiatBalance"
+                style="opacity: 0.75"
+              ></span>
+            </div>
           </q-card-section>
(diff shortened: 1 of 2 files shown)


prusnak (approved):
tested ACK

motorina0 (approved):
tACK

dni (approved):

Perlover:
I updated the LNBits from v0.12.2 to v0.12.4 and now I don't see balances of wallets in fiat balances. Is it bug of v0.12.4?

dni:
yes, its just the initial load, if you switch pages it comes up, its fixed and will be in 0.12.5. https://github.com/lnbits/lnbits/pull/2346

PR #2161, opened 2023-12-08
Handles intents

arcbtc:
If lightning/lnurl param is present in /wallet, it will pass to the decode dialogue

### DIFF
--- a/lnbits/core/templates/core/wallet.html
+++ b/lnbits/core/templates/core/wallet.html
@@ -225,7 +225,7 @@
                   </q-td>
                 </q-tr>
 
-                <q-dialog v-model="props.expand" :props="props">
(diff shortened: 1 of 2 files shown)


motorina0:
Suggestion:
 - skip the `Read` step in this case (the text of the invoice/lnurl is not of interest to the user)

image

motorina0 (approved):

motorina0:
Added `position="top"` for all dialogs on `wallet` page (so that mobile keyboard does not overlap with dialog buttons).

### Before
telegram-cloud-photo-size-4-6033070853828230895-y

### After
telegram-cloud-photo-size-4-6033070853828230896-y

prusnak (approved):
utACK

dni (approved):

PR #2163, opened 2023-12-10
Add nixos module option to pass additional environment variables

HaosGames:
This adds an additional option to the nixos module so you can pass additional environment variables to lnbits. 
These variables are merged with the variables set in the systemd service.


prusnak (approved):
tested ACK

dni (approved):

PR #2170, opened 2023-12-12
update nix ci action to use cachix to cache results

prusnak:
- this should reduce the build time needed in the CI
- this should reduce the build time needed when deploying LNbits - users just need to call `cachix use lnbits` before calling `nix build`


prusnak:
nix build phase that took 25 minutes now takes 21 seconds 😅

dni (approved):

motorina0 (approved):

PR #2171, opened 2023-12-12
Login improve UI config

motorina0:
Various UI improvements and bug fixes:

- [x] UI config for Auth (except `auth_secret` which is env only)
  - image
- [x] fix hard to click on `My Account`
   -  image
- [x] show warning on logout
   -  image
- [x]  update login warning message
   -  image
- [x] put long text somewhere
   - image
- [x] check login page on mobile
   - image
- [x] hide user ID by default on account page
   - image
- [x] fix user id first login account button
- [x] i18n account page
- [x] "Check email using:" is confusing when no auth provider
- [x] remove `Display Name`  field


dni (approved):

lnbits/core/templates/core/account.html
```diff
@@ -139,7 +155,7 @@ <h4 class="q-my-none">Account Settings</h4>
                   :src="'{{ static_url_for('static', 'images/google-logo.png') }}'"
                 ></q-img>
               </q-avatar>
-              <div>Google</div>
+              <div><span v-text="$t('google')"></span></div>
```
prusnak, line 158: Do you expect that "Google" is called something else in some languages?

If not, I think we should not translate this string. Same for "GitHub".

motorina0: Maybe in Pirate :)
Reverted in this commit https://github.com/lnbits/lnbits/pull/2171/commits/6785846d1cfbe720287cd53fb11348d61251aafc

talvasconcelos (approved):

PR #2176, opened 2023-12-13
feat: `cleanup-wallets` cli command

dni:
`lnbits-cli db cleanup-wallets` removes all wallets that never had an transaction. this helps against spammers creating random wallets, eventually slowing down the db.

### DIFF
--- a/lnbits/commands.py
+++ b/lnbits/commands.py
@@ -162,7 +162,7 @@ async def db_versions():
 
 
 @db.command("cleanup-wallets")
-@click.argument("delta", required=False)
+@click.argument("delta", type=int, required=False)
 @coro
 async def database_cleanup_wallets(delta: Optional[int] = None):
     """Delete all wallets that never had any transaction"""
@@ -179,7 +179,7 @@ async def database_cleanup_deleted_wallets():
 
 
 @db.command("cleanup-accounts")
-@click.argument("delta", required=False)
+@click.argument("delta", type=int, required=False)
 @coro
 async def database_cleanup_accounts(delta: Optional[int] = None):
     """Delete all accounts that have no wallets"""
--- a/lnbits/commands.py
+++ b/lnbits/commands.py
@@ -162,12 +162,12 @@ async def db_versions():
 
 
 @db.command("cleanup-wallets")
-@click.argument("cleanup_wallets_days", type=int, required=False)
+@click.argument("days", type=int, required=False)
 @coro
-async def database_cleanup_wallets(cleanup_wallets_days: Optional[int] = None):
+async def database_cleanup_wallets(days: Optional[int] = None):
     """Delete all wallets that never had any transaction"""
     async with core_db.connect() as conn:
-        delta = cleanup_wallets_days or settings.cleanup_wallets_days
+        delta = days or settings.cleanup_wallets_days
         delta = delta * 24 * 60 * 60
         await delete_unused_wallets(delta, conn)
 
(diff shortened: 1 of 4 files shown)


lnbits/core/crud.py
```diff
@@ -451,6 +451,17 @@ async def delete_wallet(
     )
 
 
+async def delete_unused_wallets(conn: Optional[Connection] = None) -> None:
+    await (conn or db).execute(
+        """
+        DELETE FROM wallets
```
motorina0, line 457: There should be an age here (older than one month for example), otherwise wallets created 1 minute ago will be deleted .

lnbits/core/crud.py
```diff
@@ -451,6 +462,10 @@ async def delete_wallet(
     )
 
 
+async def delete_deleted_wallets(conn: Optional[Connection] = None) -> None:
```
motorina0, line 465:
```suggestion
async def remove_deleted_wallets(conn: Optional[Connection] = None) -> None:
```

prusnak:
I think we should wait for the PR https://github.com/lnbits/lnbits/pull/2177 merge and then update his PR to use `created_at` to delete only old wallets/accounts.

prusnak:
> I think we should wait for the PR #2177 merge and then update his PR to use `created_at` to delete only old wallets/accounts.

#2177 has been merged now

lnbits/core/crud.py
motorina0: This can be updated now: https://github.com/lnbits/lnbits/pull/2177

lnbits/settings.py
```diff
@@ -347,6 +347,8 @@ class EnvSettings(LNbitsSettings):
     log_rotation: str = Field(default="100 MB")
     log_retention: str = Field(default="3 months")
     server_startup_time: int = Field(default=time())
+    # 3 months, 60 * 60 * 24 * 30 * 3
+    cleanup_wallets_delta: int = Field(default=7776000)
```
motorina0, line 351: please add this variable to `.env.example`

lnbits/commands.py
```diff
@@ -159,6 +162,42 @@ async def db_versions():
         return await get_dbversions(conn)
 
 
+@db.command("cleanup-wallets")
```
motorina0, line 165:
```suggestion
@coro
```
use `@coro` and avoid the loop

prusnak:
I am thinking whether delta in seconds is not too nerdy and maybe 99% admins just need delta in days?
If you agree, we can rename `CLEANUP_WALLETS_DELTA` to `CLEANUP_WALLETS_DAYS` and adjust the math accordingly. What do you think @dni?

dni:
yeah cool idea, lets use days

dni:
done

lnbits/commands.py
```diff
@@ -151,16 +154,41 @@ async def migrate_databases():
...
+@db.command("cleanup-accounts")
+@click.argument("days", type=int, required=False)
+@coro
+async def database_cleanup_accounts(days: Optional[int] = None):
+    """Delete all accounts that have no wallets"""
     async with core_db.connect() as conn:
-        return await get_dbversions(conn)
+        delta = days or settings.cleanup_wallets_days
+        delta = delta * 24 * 60 * 60
+        await delete_unused_wallets(delta, conn)
```
prusnak, line 191: I think this should be `delete_accounts_no_wallets`?

prusnak (changes requested):

lnbits/commands.py
dni: yep, copy paste error :o

prusnak (approved):

motorina0 (approved):

PR #2179, opened 2023-12-13
Update the GET /wallet endpoint to correctly handle an empty state

gkrizek:
In this PR: https://github.com/lnbits/lnbits/pull/1932

It broke the situation where someone just load ups the `/wallet` endpoint and it would create a new wallet. We would get the error:

```
Error: [{'loc': ('query', 'usr'), 'msg': 'field required', 'type': 'value_error.missing'}]
```

This is because it's expecting the User to be defined which it isn't in this case of just loading the page. So we need to add back in the create user call if the user isn't defined.

### DIFF
--- a/lnbits/core/views/generic.py
+++ b/lnbits/core/views/generic.py
@@ -165,11 +165,16 @@ async def extensions_install(
 )
 async def wallet(
     request: Request,
-    usr: UUID4 = Query(...),
+    usr: UUID4 = Query(None),
     wal: Optional[UUID4] = Query(None),
 ):
-    user_id = usr.hex
-    user = await get_user(user_id)
+    if not usr:
+        new_user = await create_account()
+        user = await get_user(new_user.id)
+        user_id = new_user.id
+    else:
+        user_id = usr.hex
+        user = await get_user(user_id)
 
     if not user:
         return template_renderer().TemplateResponse(


prusnak:
> So we need to add back in the create user call if the user isn't defined.

The whole point of PR https://github.com/lnbits/lnbits/pull/1932 was to remove this behaviour. Why don't you create an user via API first?

gkrizek:
Oh I didn't read it as such
> and the frontend is modified to use the api endpoints

It made it sound like it was just a refactor with the same behavior 

What's the reason for removing the functionality? It's just additional complexity to do something via the API that has always existed in LNbits. So I was preferring to utilize the LNbits page calling to the APIs instead of my own service calling the APIs.

prusnak:
> What's the reason for removing the functionality?

I assume it was making the design cleaner ("Explicit is better than implicit."). But I'll let @dni answer that question since he did the change.

gkrizek:
Ok yeah will be interested in what the expectations were, because there's code in there to create a wallet if one isn't defined. Just not a user.

dni:
we also have to handle if the admin enables LNBITS_ALLOWED_USERS.

i also think now that we have https://github.com/lnbits/lnbits/pull/2092 how to create accounts changed.

motorina0:
> @motorina0 what do you think?

I see this PR is against `main` can you please update it to `dev`.
Meanwhile on `dev` the login PR brought some changes for this part of the code:
```py
async def wallet(
    request: Request,
    user: User = Depends(check_user_exists),
    wal: Optional[UUID4] = Query(None),
):
    if wal:
        wallet_id = wal.hex
    elif len(user.wallets) == 0:
        wallet = await create_wallet(user_id=user.id)
        user = await get_user(user_id=user.id) or user
        wallet_id = wallet.id
    else:
        wallet_id = user.wallets[0].id

    user_wallet = user.get_wallet(wallet_id)
```

> we also have to handle if the admin enables LNBITS_ALLOWED_USERS.
 - the `check_user_exists` decorator should handle this

motorina0:
> Ok yeah will be interested in what the expectations were, because there's code in there to create a wallet if one isn't defined. Just not a user.

That is correct. There is no problem in creating the default wallet (if one does not exist) for a user. The operation is idempotent (OK for a `GET`).
However creating a new user on each `GET` is NOK, a `POST /api/v1/account` endpoint exists for this.

prusnak:
> I see this PR is against `main` can you please update it to `dev`.

Changed the base branch here in the PR. Now the PR branch has to be rebased and conflicts resolved:

```
git checkout dev
git pull
git checkout gkrizek/fix-wallet-page
git rebase dev
```

gkrizek:
Ok thanks for the updates here. So I'm a little confused on where to go. 

@motorina0, the intention of this PR was to restore the functionality of `GET /wallet` that was broken in a previous PR. Are you saying that is no longer the intended behavior and thus I should just close this PR? I ran the `dev` branch and it does fail with a better error (`Missing user ID or access token`), but still has the changed functionality.

prusnak:
> Are you saying that is no longer the intended behavior and thus I should just close this PR?

I think so. Can you please confirm @dni @motorina0?

motorina0:
> Are you saying that is no longer the intended behavior and thus I should just close this PR?

That is correct. Closing PR.

PR #2184, opened 2023-12-15
fix make checkbundle which should not bump the serviceworker version

prusnak:


dni:
it is comparing the bundle files not the service worker. 

is that fix for running it locally?

prusnak:
> it is comparing the bundle files not the service worker. 

Can you rephrase where is the problem? I don't understand what is the issue.

dni (approved):

motorina0 (approved):

PR #2185, opened 2023-12-16
fix the url for js libs

talvasconcelos:
This fixes, for example, printing LNURLw vouchers.

### DIFF
--- a/lnbits/templates/print.html
+++ b/lnbits/templates/print.html
@@ -34,7 +34,7 @@
     </q-layout>
 
     {% for url in INCLUDED_JS %}
-    <script src="{{ url }}"></script>
+    <script src="{{ static_url_for('static', url) }}"></script>
     {% endfor %}
     <!---->
     {% block scripts %}{% endblock %}


talvasconcelos:
Closes lnbits/withdraw#22

prusnak (changes requested):
I think we also need to fix the CSS:

``` patch
diff --git a/lnbits/templates/print.html b/lnbits/templates/print.html
index ef6beb4a..4fde0878 100644
--- a/lnbits/templates/print.html
+++ b/lnbits/templates/print.html
@@ -3,7 +3,7 @@
 <html lang="en">
   <head>
     {% for url in INCLUDED_CSS %}
-    <link rel="stylesheet" type="text/css" href="{{ url }}" />
+    <link rel="stylesheet" type="text/css" href="{{ static_url_for('static', url) }}" />
     {% endfor %}
     <style>
       @page {
```

prusnak (commented):
CSS fixed in 07c01ab08169ea229bf38726bc03546a0e7c3772

prusnak (approved):

dni (approved):

PR #2186, opened 2023-12-18
[draft] feat: set `samesite="none", secure=True` for cookies

motorina0:
### Summary
 - this change is in order to allow the LNbits site to be **securely** embedded as an iFrame in 3rd party sites
 - this is required for browser extension but can be used in other contexts as well
 

https://github.com/lnbits/lnbits/assets/2951406/65fc6527-a8de-4868-a4b7-811d7df388f2

The main question is if `samesite="none"` will introduce any vulnerability/risk (given that `httponly=True` and `secure=True` are set).

It is important for the cookie containing the auth-token not be accessible by 3rd party sites. This cookie now (with this PR) has these 3 attributes: `httponly=True, samesite="none", secure=True`

More details about `HttpOnly` and `Secure` attributes [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies). `SameSite attribute` described on the same page below (no anchor available)

Other Reading:
 - https://vercel.com/guides/understanding-the-samesite-attribute
 - https://www.invicti.com/learn/cookie-security-flags/
 - https://0xn3va.gitbook.io/cheat-sheets/web-application/cookie-security

### DIFF
--- a/lnbits/core/views/auth_api.py
+++ b/lnbits/core/views/auth_api.py
@@ -292,7 +292,9 @@ def _auth_success_response(
         data={"sub": username or "", "usr": user_id, "email": email}
     )
     response = JSONResponse({"access_token": access_token, "token_type": "bearer"})
-    response.set_cookie("cookie_access_token", access_token, httponly=True)
+    response.set_cookie(
+        "cookie_access_token", access_token, httponly=True, samesite="none", secure=True
+    )
     response.set_cookie(
         "is_lnbits_user_authorized", "true", samesite="none", secure=True
     )
@@ -304,7 +306,9 @@ def _auth_success_response(
 def _auth_redirect_response(path: str, email: str) -> RedirectResponse:
     access_token = create_access_token(data={"sub": "" or "", "email": email})
     response = RedirectResponse(path)
-    response.set_cookie("cookie_access_token", access_token, httponly=True)
+    response.set_cookie(
+        "cookie_access_token", access_token, httponly=True, samesite="none", secure=True
+    )
     response.set_cookie(
         "is_lnbits_user_authorized", "true", samesite="none", secure=True
     )


motorina0:
Useful Answer:
https://stackoverflow.com/questions/70745635/how-does-httponly-prevent-from-malicious-package-to-steal-the-content-in-it

HTTP requests carry with it only cookies that are applicable to the domain being requested, not all of the browser's cookies. If you have a browser with cookies for bank.com, shop.com, and evil-site.com, a request to evil-site.com would only send evil-site.com cookies with the request. Not so useful.

Now say a malicious script has infected bank.com and has placed itself on a page in that domain. Now that script is running in the context of bank.com and the currently viewing user. It can now read bank.com's non-HttpOnly cookies using [document.cookie](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie) and [send them to evil-site.com with a simple script](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#security). This means that if you logged in to bank.com and viewed that infected page, your login cookies can now be stolen.

Marking a cookie as HttpOnly tells the browser not to expose the cookie to JavaScript, i.e. any script, legit or not, cannot read the cookie's value from document.cookie. So if bank.com made their login cookie HttpOnly, this cookie would not be readable by any script on the page. However, the cookie is still passed back and forth between browser and bank.com in requests and responses while the cookies are present and valid.

HttpOnly is just one of the many measures to prevent cookie theft and should be complemented by other security features. Secure makes sure the cookie is only ever sent through HTTPS connections. SameSite defines when the cookie is allowed to cross sites. HTTPS connections prevent reading the request over the network.

dni:
has to consequences for local development running http?

dni (approved):

motorina0:
> has to consequences for local development running http?

Not if you are on localhost:
`
A cookie with the Secure attribute is only sent to the server with an encrypted request over the HTTPS protocol. It's never sent with unsecured HTTP (except on localhost), which means [man-in-the-middle](https://developer.mozilla.org/en-US/docs/Glossary/MitM) attackers can't access it easily. Insecure sites (with http: in the URL) can't set cookies with the Secure attribute.
`

motorina0:
Closing for now.

PR #2187, opened 2023-12-19
Fix: LNbits backend payment state and catch errors

callebtc:


lnbits/wallets/lnbits.py
```diff
@@ -121,10 +121,12 @@ async def get_invoice_status(self, checking_id: str) -> PaymentStatus:
             return PaymentStatus(None)
 
     async def get_payment_status(self, checking_id: str) -> PaymentStatus:
-        r = await self.client.get(url=f"/api/v1/payments/{checking_id}")
+        try:
+            r = await self.client.get(url=f"/api/v1/payments/{checking_id}")
+            r.raise_for_status()
+        except Exception:
+            return PaymentStatus(None)
```
prusnak, line 128: CI fails - shouldn't this be

```suggestion
            return PaymentStatus(False)
```

?

prusnak (dismissed):

lnbits/wallets/lnbits.py
callebtc: Nope, that's the bug I'm fixing here.

dni (approved):

motorina0 (approved):

PR #2188, opened 2023-12-20
fix: position of method invocation

motorina0:
By mistake the call to `await this.checkUsrInUrl()` was placed inside the ` if (window.extensions) {` conditional block.

This PR moves it outside of it.


prusnak (approved):
utACK

checkbundle CI check passes ✅

arcbtc (approved):

PR #2191, opened 2023-12-22
move API docs as last item of wallet actions

prusnak:
API is an advanced (though integral) feature of a wallet. Let's move it to the end of the action list.

Also lots of people have trouble finding the access keys, so maybe we should change the title from "API docs" to "API keys and API docs"

Before:

Screenshot 2023-12-22 at 10 24 24

After:

Screenshot 2023-12-22 at 13 09 47


dni:
the API docs, includes the most important part of the whole accordion, which is wallet_id, admin_key and invoice_key.  so i think its kind important. but i don't like it that it is hidden there tbh. also showing it immediately would be bad security, maybe immediately and blurred or similar?

prusnak:
> includes the most important part of the whole accordion

Maybe important for you, but for 99% it is not important. They care only about the QR code which is now the first item.

dni (approved):

prusnak:
After chat with @dni I am also pushing the change which changes the icon and wording of the item, so the API keys are discoverable easier:

Screenshot 2023-12-22 at 12 24 12

dni (approved):

motorina0 (approved):

PR #2198, opened 2024-01-09
revert: migration

dni:
took to long on legend to run, now taking current date as default


arcbtc (approved):

talvasconcelos (approved):

motorina0 (approved):

prusnak:
I wonder how long it would take if you replaced

``` sql
SELECT foo FROM ... WHERE ... ORDER BY foo ASC LIMIT 1
```

with

``` sql
SELECT MIN(foo) FROM ... WHERE ...
```

PR #2202, opened 2024-01-10
refactor: extract `AESCipher` to `crypto.py`

motorina0:
### Summary

`AESCipher` class is used for encrypting/decrypting messages.
Initially it was required for macaroons, but now the auth logic also uses it. Therefore it has been moved to its own `utils.crypto.py` file and the imports updated.

### DIFF
--- a/lnbits/core/views/auth_api.py
+++ b/lnbits/core/views/auth_api.py
@@ -20,9 +20,7 @@ from lnbits.helpers import (
     is_valid_username,
 )
 from lnbits.settings import AuthMethods, settings
-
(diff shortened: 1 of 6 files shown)


prusnak:
Great refactor!

The cashu extension has its own copyof AESCipher here: https://github.com/lnbits/cashu/blob/main/lib/cashu/core/crypto/aes.py

We can refactor it to use `lnbits.utils.crypto.AESCipher` and bump it `min_lnbits_version` to 0.12.0 once released.

talvasconcelos (approved):

PR #2206, opened 2024-01-11
Deactivate all extensions flag

motorina0:
### Summary
Add the `LNBITS_EXTENSIONS_DEACTIVATE_ALL` flag. If set to true then on server startup the extensions will not be loaded. This will exclude running the migration and  the route mapping.

### Test Scenarios

**Scenario 1**
 - start the server with the flag set to `false` (default)
 - test install/uninstall/upgrade
 - new: if an extension is deactivated at the start-up moment then it will not be loaded

**Scenario 1**
 - start the server with the flag set to `true`
 - expect: 
    - no extension should be loaded
    - the extensions are not visible on the left hand side
    - the extensions are visible on the extensions page (http://localhost:5000/extensions)
    - the extensions can be installed/uninstalled/de-activated

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -141,6 +141,9 @@ LNBITS_ADMIN_USERS=""
 # Extensions only admin can access
 LNBITS_ADMIN_EXTENSIONS="ngrok, admin"
 
+# Start LNbits core only. The extensions are not loaded.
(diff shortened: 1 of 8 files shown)


arcbtc:
Love it

arcbtc (approved):

dni (approved):
utACK

PR #2210, opened 2024-01-16
refactor: extract `encrypt_internal_message` and `decrypt_internal_message`

motorina0:
### Summary
Small refactoring that allows extensions also to encrypt messages using the default auth secret.

One use-case is exposing secret endpoints that can only be used between extensions.

### Example

```py
@nostrclient_ext.websocket("/api/v1/{id}")
```

If the decrypted value of `id` is `"relay"` then the connection is allowed (made by another extension). Otherwise the connection is dropped.


prusnak (approved):
utACK

arcbtc (approved):

PR #2218, opened 2024-01-26
refactor: move admin ui tabs around

dni:
move authentication into security in preparation of user manager pr where users tab will be removed. move service fee into server tab, i think its more fitting there.

screenshot-1706257572
screenshot-1706257568


arcbtc (approved):
lovely

motorina0 (approved):

Issue #2211, opened 2024-01-19
[BUG] Alby funding source in LNBits isn't set up properly


theroyalwhale:
**Bug Description**
In the LNBits Admin panel, if you set Alby as your funding source, the value that Alby Wallet API returns will be considered msats while it's actually in sats.

**To Reproduce**
1. Open LNBits instance
2. Go to Admin panel
3. Set Alby as funding source
4. Put your access token
5. Save changes
6. "why do I have 0.012 sats"

**Expected behavior**
It should clearly be treated as sats instead of msats.

**Desktop (please complete the following information):**
- LNbits version: [Latest (I use official LNBits hosting)]
- Database [none]

**Additional context**
Nope.

theroyalwhale:
thanks people in the telegram! :)

dni:
is this issue resolved?

dni:
@bitkarrot  can you take a look at this?

theroyalwhale:
While waiting for this, my Legend LNBits wallet was hacked!
I do not have ANY idea how my user key leaked.
For reference, here is the invoice the hacker used: lnbc510u1pj6an6dsp5l7g8e3fnzv0x66hguvyy9xchwz3fqwmfaqp424dcyvlnlw7ydzmspp54293qafcfuske6qdczccfy9skf2srdddq7x5365e8tlk2qca7t2qdp27z0c02lsn7rmwg8sn7r6huyls7mjpuyls74lp8u8kuxqzjccqpjrzjqvtesvezx7wct8qdgw5serwd8eernw8qvudspfjhecujf4zf3um4gzl6ryqqfdgqqqqqqqlgqqqqqzsqyg9qxpqysgqrlf93e5qu66f3wpuk7qp6y5j439tak09sdjkc3j3wphl5aqna5nr9fathf2ggypmyu4myya8scvd58pa8n99pejaw4f5k8c2gsyq77gpeyy5mk

bitkarrot:
@dni ok i'll take a look at this on version 0.12 

not sure how legend can get hacked unless you accidentally pasted your link or admin keys somewhere visible.

theroyalwhale:
@bitkarrot i was using legend as a TEMPORARY funding source while this alby bug is being patched, i'm CERTAIN i NEVER shared the admin key / user key ANYWHERE AT ALL. perhaps one of the worse parts is that I know who the hacker is but I don't have PROOF it's him... >:(
[...]

PR #2220, opened 2024-01-28
Fix msat balance reporting from alby api

bitkarrot:
Ref https://github.com/lnbits/lnbits/issues/2211#issuecomment-1913219499

- fixed balance with 1000 multiplier for msat conversion
- tested send and receive invoices. 

Note: perhaps we should make a unit test for this, in case API changes in the future. Not sure if there any standard other than the existing tests folder. LMK. 

@dni


lnbits/wallets/alby.py
```diff
@@ -47,7 +47,8 @@ async def status(self) -> StatusResponse:
         data = r.json()["balance"]
         if r.is_error:
             return StatusResponse(data["error"], 0)
```
prusnak, line 49: Unrelated to the PR, but this code looks really weird.

1. First, we parse the JSON, but we check if there is an error only afterwards.
2. What's even weirder is that we access `data["error"]`, but I don't think `data` is a dict at this point.

Maybe this section should be this instead?

I checked the [documentation here](https://guides.getalby.com/developer-guide/v/alby-wallet-api/reference/api-reference/account) but I haven't tested it. 

```suggestion
        if r.is_error:
            error_message = r.json()["message"]
            return StatusResponse(error_message, 0)
        data = r.json()["balance"]
```

prusnak (changes requested):

lnbits/wallets/alby.py
bitkarrot: you are correct, it should message, and consistent with other methods in same class. 

Should we have unit tests to simulate this testing? Most projects at this size start to have them.

bitkarrot:
having some difficulty connecting to Alby API at the moment. but feel free to validate :)

lnbits/wallets/alby.py
bitkarrot: Same pattern appears in opennode.py

https://github.com/lnbits/lnbits/blob/e1bb2113ed92556c6227f6de14126570f808007a/lnbits/wallets/opennode.py#L59-L63

bitkarrot: I reviewed the other wallet files (e.g. lnpay, etc). Every single one has a slightly different style. Not sure what you guys want to do for handling errors and exceptions, so I originally templated the alby wallet integration off of opennode.py

prusnak: > Same pattern appears in opennode.py

Can you please fix opennode in this PR the same way like you fixed alby?

theroyalwhale:
when merge? :)

prusnak:
I cherry picked your changes into https://github.com/lnbits/lnbits/pull/2221 where I also fixed the same issues in opennode

Thanks!

bitkarrot:
@prusnak ok! You beat me to it while I was sleeping! :D

PR #2222, opened 2024-01-29
Add wallet max balance

callebtc:
New `.env` variable `LNBITS_MAX_WALLET_BALANCE_SAT` can limit the maximum balance for each wallet and throw an error if the wallet attempts to create a new invoice.

Example: `LNBITS_MAX_WALLET_BALANCE_SAT=1000000`


prusnak:
Not into main please. We use dev for development

PR #2223, opened 2024-01-29
Wallet limits: max balance, daily max withdraw, transactions per sec

callebtc:
New `.env` variable `LNBITS_MAX_WALLET_BALANCE_SAT` can limit the maximum balance for each wallet and throw an error if the wallet attempts to create a new invoice.

Example: `LNBITS_MAX_WALLET_BALANCE_SAT=1000000`

=============UPDATE==============

Legacy banks, exchanges, anything dealing with money have wallet limits, so adding a few more optional wallet limits seems like a good security tool.

Added daily max limit and transactions per second.

image


dni (changes requested):
it should be editable on the admin ui aswell

arcbtc (approved):
LFG!

package.json
```diff
@@ -20,6 +20,7 @@
     "@chenfengyuan/vue-qrcode": "1.0.2",
     "axios": "^1.6.0",
     "chart.js": "2.9",
+    "fs": "^0.0.1-security",
```
motorina0, line 23: Why is this package required?

https://www.npmjs.com/package/fs:

> This package name is not currently in use, but was formerly occupied by another package. To avoid malicious use, npm is hanging on to the package name, but loosely, and we'll probably give it to you if you want it.

motorina0 (approved):

PR #2225, opened 2024-01-29
Add bip32

callebtc:
This PR adds the `bip32` dependency which I need to generate keys in the latest [cashu extension update](https://github.com/lnbits/cashu/pull/51) that I've been preparing in the last couple of days. `bip32` is a tiny but widely-used library and I believe that LNbits can make use of it also in other extensions, such as the on-chain wallet (to generate keys from seed phrases).

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -106,6 +106,21 @@ files = [
 ]
 
 [[package]]
+name = "bip32"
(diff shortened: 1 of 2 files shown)


motorina0:
Cashu uses `embit` now. Closing issue.

PR #2226, opened 2024-01-30
Improved env layout

arcbtc:
LNBITS_ADMIN_UI has become the most important variable, so should be at the top of the file

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,9 +4,11 @@
 ########### Admin Settings ###########
 ######################################
 
-# Enable Admin GUI, available for the first user in LNBITS_ADMIN_USERS if available
-# Warning: Enabling this will make LNbits ignore this configuration file. Your settings will
-# be stored in your database and you will be able to change them only through the Admin UI.
+# Enable Admin GUI, available for the first user in LNBITS_ADMIN_USERS if available.
+# Warning: Enabling this will make LNbits ignore most configurations in file. Only the
+# configurations defined in `ReadOnlySettings` will still be read from the environment variables.
+# The rest of the settings will be stored in your database and you will be able to change them 
+# only  through the Admin UI.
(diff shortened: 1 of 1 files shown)


prusnak (approved):
I confirmed that the section is just moved and its contents is not modified.

dni:
once again i use this opportunity to questions if admin_ui should be enabled by default, because i think it should be.

.env.example
```diff
@@ -1,5 +1,26 @@
 #For more information on .env files, their content and format: https://pypi.org/project/python-dotenv/
 
+######################################
+########### Admin Settings ###########
+######################################
+
+# Enable Admin GUI, available for the first user in LNBITS_ADMIN_USERS if available
+# Warning: Enabling this will make LNbits ignore this configuration file. Your settings will
+# be stored in your database and you will be able to change them only through the Admin UI.
+# Disable this to make LNbits use this config file again.
```
dni, line 10: i think this is not true, disabling this will do nothing, as soon as its creates the settings table it will use it, so you will need to truncate settings and disabling this.

but honestly i don't have a clue why you should ever switch back or between them.

arcbtc:
> once again i use this opportunity to questions if admin_ui should be enabled by default, because i think it should be.

Its easy enough to enable on first launch with `LNBITS_ADMIN_UI=false poetry run lnbits`

I also feel we willl end up defaulting, but it should be a slow move, to encourage users to use.

.env.example
motorina0, line 10: Some settings like DB connection, `LNBITS_ADMIN_UI` itself (and others) are always read from the environment variables (`.env` file or actual env vars).

These settings are grouped under the [ReadOnlySettings](https://github.com/lnbits/lnbits/blob/dev/lnbits/settings.py#L401) class.

PR #2230, opened 2024-02-01
Fixes annoying qr popup on parsing LNURL withdraw receive in wallet

arcbtc:


dni:
i cannot really tell what the issue is, can you provide a screenshot or something what the problem is?

dni (approved):
still going to approve it, despite missing some context, seems simple enough

lnbits/core/templates/core/wallet.html
```diff
@@ -569,7 +569,10 @@ <h6 class="text-subtitle1 q-mt-none q-mb-sm">
             ></q-spinner>
           </q-form>
         </q-card>
-        <q-card v-else class="q-pa-lg q-pt-xl lnbits__dialog-card">
+        <q-card
+          v-else-if="receive.paymentReq && receive.lnurl == null"
```
motorina0, line 573: the corresponding `v-if` is:
```python
        <q-card
          v-if="!receive.paymentReq"
```
so `receive.paymentReq` is not required in this condition

motorina0, line 573:
```suggestion
          v-else-if="receive.lnurl == null"
```

motorina0: what if this condition is false: `receive.lnurl == null` (`receive.lnurl != null`)?

arcbtc: The invoice QR code should only fire creating an invoice, not when an invoice is parsed out of an lnurl-withdraw

motorina0 (approved):

talvasconcelos (approved):

PR #2231, opened 2024-02-02
Adds optional confetti to incoming payments in wallet

arcbtc:
https://github.com/lnbits/lnbits/assets/33088785/09ccafa8-6eca-4a27-9af0-e8a3faccb332

The confetti websocket can be added to any page with `eventReactionWebocket(event_id, event_function)`

For the wallet example `event_id` is the wallet ID, and `event_function` is the function `eventReactionBothSidesConfetti` (`eventReactionFireworksConfetti` also exists but is not used)

It made sense to leave room for multiple animations the developer could summon using `eventReactionWebocket`, or the user could specify in the profile page.


prusnak:
This is awesome!!!!

Does it look good on mobile too (vertical aspect ratio)?

prusnak:
Needs `make prettier` and `make bundle`.

lnbits/static/js/event-reactions.js
```diff
@@ -0,0 +1,516 @@
+function eventReactionWebocket(event_id, event_function) {
+  localUrl = ''
+  console.log(event_id)
```
dni, line 3: remove console.log

lnbits/static/js/event-reactions.js
```diff
@@ -0,0 +1,516 @@
...
+    })
+    confetti({
+      ...defaults,
+      particleCount,
+      origin: {x: randomInRange(0.7, 0.9), y: Math.random() - 0.2}
+    })
+  }, 250)
+}
+
+!(function (t, e) {
```
dni, line 77: is this a library or what is this? did you write this?

dni:
why are you creating an extra websocket connection? it should be enough to trigger the animation when we also close the popup after a successful payment. we should keep those connections to a minimum.

i don't think that the animation should be disabled by default its such a good feature and every wallet has it, i don't even think its should be configurable at all!

arcbtc:
> This is awesome!!!!
> 
> Does it look good on mobile too (vertical aspect ratio)?

https://github.com/lnbits/lnbits/assets/33088785/bdfcab86-016a-4a39-b7c3-e74408eb911c

arcbtc:
> why are you creating an extra websocket connection? it should be enough to trigger the animation when we also close the popup after a successful payment. we should keep those connections to a minimum.
> 
> i don't think that the animation should be disabled by default its such a good feature and every wallet has it, i don't even think its should be configurable at all!

Its enabled by default. People like customising stuff, the toggle switch will later turn into a select box, where a user could select from various animations. The admin could then set what animations are available, same flow as themes really

lnbits/static/js/event-reactions.js
arcbtc: Well used little  confetti library, thats easy to extend, such as in the above couple of functions.

lnbits/static/js/event-reactions.js
arcbtc: done

talvasconcelos:
How does one create (and add) a new animation?

arcbtc:
For now in the reactions.js file, but later I will add gif support and uploading gifs.

motorina0 (approved):

arcbtc:
> can you add the library via npm and into our build? is it this library? https://www.npmjs.com/package/js-confetti

Nien! Its been changed, its also very small, so I think this approach is unnecessary

PR #2232, opened 2024-02-02
Add support for the new CLNRest as a funding source so that runes can be used to give LNBits limited access to the lightning node

Amperstrand:
Core lightning has a new [CLNRest](https://docs.corelightning.org/docs/rest) interface to talk to a node.

The main advantages of this is that it uses [Runes](https://docs.corelightning.org/reference/lightning-createrune#examples) for authentification. Runes can be made to restrict a clients access to the node. Restrictions can include rate-limiting API requests and how many sats can be paid in a single invoice.

Together a rate-limit and amount-limit can allow you to connect LNBits to core lightning in a way that at most X invoices can be paid over a time period of 24 hours. Even if there is an exploit in LNBits, it won't be able to drain all funds from the lightning node.

Another advantage ov the new CLNRest interface is that unlike the old [c-lightning-REST](https://github.com/Ride-The-Lightning/c-lightning-REST) interface, CLNRest does not have an external dependency on a seperate program.

It is my understanding that the old c-lightning-REST will eventually be deprecated.

This pull request is my attempt at support for CLNRest.

Marked as draft because even though I think it works, it is not well tested. Also missing documentation for now.


Amperstrand:
For now, the code falls back to voidwallet if LNBITS_ADMIN_UI=true:

```
2024-02-02 15:03:21.93 | ERROR | lnbits.app:lnbits_startup:385 | Error initializing CLNRestWallet: cannot initialize CLNRest: missing clnrest_rune
2024-02-02 15:03:21.93 | WARNING | lnbits.app:set_void_wallet_class:176 | Fallback to VoidWallet, because the backend for CLNRestWallet isn't working properly
```

This likely has something to do with an entry being needed in the js file somewhere around [here](https://github.com/lnbits/lnbits/blob/43d92ddb51d7bfad045c5136d9588ced35811282/lnbits/static/js/components/lnbits-funding-sources.js#L43)

I think I also need to edit the file ./lnbits/static/bundle.min.js after making changes to lnbits-funding-sources.js in order to make the UI work again

lnbits/settings.py
```diff
@@ -152,6 +152,12 @@ class CoreLightningRestFundingSource(LNbitsSettings):
     corelightning_rest_cert: Optional[str] = Field(default=None)
 
 
+class CLNRestFundingSource(LNbitsSettings):
+    clnrest_url: Optional[str] = Field(default="https://127.0.0.1:3010")
```
prusnak, line 156: Default in `.env.example` uses `http` - which one is correct `http` or `https`?

dni:
hey! thanks for your contribution! this is amazing. i did not even know about this plugin!

```
CLNRestWallet: cannot initialize CLNRest: missing clnrest_rune
```

something is up with your settings you hitting your own assert, try it on a fresh database or just truncate the `settings` table.

Amperstrand:
To generate a read-only rune that will only allow lnbits to create invoices (but not pay them):

```
lightning-cli createrune restrictions='[["method^list", "method^get", "method=summary", "method=invoice", "method=waitanyinvoice"]]'
```

I'm not sure how to generate a rune that has the above permissions but can also access the pay command in a limited way, but a rune that can only access the pay command and requires you to specify amountsats under 1.001 satoshi can be generated like this:

lightning-cli createrune restrictions='[["method=pay"], ["pnameamountmsat<1001"]]'

A limitation with runes (as far as I can tell) is that they are applied as a filter to what commands can be run before they are run. So for example, I can impose a rune that can invoke the pay method if amountmsat is less than 1001. This rune cannot be used to call the pay command without amountsat being set because it is only valid if amountmsat is set to less than 1001 (and not if it is null).

So this means that if you ask to pay a bolt11 invoice where the amount of 1 satoshi is encoded in the invoice, the rune to pay up to 1001 msat will not work because you can't use the rune unless you specify amountmsat. And you can't specify amountmsat if it is encoded in the invoice. https://github.com/ElementsProject/lightning/issues/7020 tracks this limitiation.

The end goal is that you should be able to give a process (LNBits, btcpayserver, BoltCard, etc) a rune that has full access to the lightning node but has a rate limited spending policy.

Another implication of this is that potentially you can apply some sort of account system on core lightning. This way, someone can hold a virtual balance on core lightning in a way that instead of rate limiting, the rune can be used to check if the balance of the rune is enough (sort of like lndhub).

motorina0:
Is this still a draft?

prusnak:
Needs rebase to remove conflicts.

newtonick:
I'd love to see this work completed as a LNBits and CLN User.

lnbits/wallets/clnrest.py
```diff
@@ -121,7 +121,7 @@ async def status(self) -> StatusResponse:
         if not data.get("channels"):
             return StatusResponse("no data or no channels available", 0)
 
-        total_our_amount_msat = sum(parse_msat(channel["our_amount_msat"]) for channel in data["channels"])
+        total_our_amount_msat = sum(channel["our_amount_msat"] for channel in data["channels"])
```
prusnak, line 182: This breaks if `channel["our_amount_msat"]` does not contain an `int`. Is it guaranteed that it will always contain an `int`?

Schnema1:
If I understand the new CLN REST server with it's runes is still not merged.  There are 3 reviewers, but did someone actually review the code? I would be glad to use the new rest server, as I do not want to setup the old rest server next to the new one. When can we expect the review to be done and the code be merged?

prusnak:
> When can we expect the review to be done and the code be merged?

Why don't you review the code and test that it works?

bitmynt:
> If I understand the new CLN REST server with it's runes is still not merged. There are 3 reviewers, but did someone actually review the code? I would be glad to use the new rest server, as I do not want to setup the old rest server next to the new one. When can we expect the review to be done and the code be merged?

It doesn't quite work and is not ready to be merged.

I have been meaning to get around to cleaning it up and getting it in a more presentable state but haven't had a chance to get around to it. Anyone is of course welcome to pick up where I left off (or start over if that is easier).

I still think this is and important feature because now that runes allow you to [limit max amount per invoice paid](https://github.com/ElementsProject/lightning/issues/7020), it allows you to rate limit how many sats lnbits can withdraw from the backend. This is ideal in for example a merchant setup where you want to be able to use lnbits to accept unlimited payments but only expect to maybe make 5-10 refunds of up to X sats per hour/day.

Amperstrand:
Closing this one in favour of https://github.com/lnbits/lnbits/pull/2778 which works with CLNRest over HTTPS and should just work.

Amperstrand:
> If I understand the new CLN REST server with it's runes is still not merged. There are 3 reviewers, but did someone actually review the code? I would be glad to use the new rest server, as I do not want to setup the old rest server next to the new one. When can we expect the review to be done and the code be merged?

If you want to help test it, feel free to spin up a VM or VPS and run this gist:

https://gist.github.com/Amperstrand/9893e23f501099576b14c8b1f6611898

PR #2235, opened 2024-02-02
 feat: ZBD wallet API Integration

bitkarrot:
EDIT: Ready for review. 

ZBD API Developer Dashboard, the Projects section was used to generate the API Keys
Screenshot 2024-02-05 at 11 41 05 AM

------------------------

This is a work in progress PR to get ZBD wallet integrated into LNBits.

It is not yet complete yet, am still working on it and testing! Will post a comment here once its ready for review. Thanks.


prusnak:
I rebased the branch on top of current dev to remove conflicts.

prusnak (approved):
utACK

Code looks fine

andrerfneves:
huge! thanks @bitkarrot for taking this on. going to find some time to spin up local LNBits and test further 👍

arcbtc:
Amazing work. Async invoice stream needs attention though, for the invoice listeners. Not sure if its possible, it should be.

prusnak:
Rebased on top of dev to get rid of bundle conflict.

bitkarrot:
> Rebased on top of dev to get rid of bundle conflict.

@prusnak  thank you thank you!

motorina0:
@bitkarrot is there a docker container for ZBD that we can use in our regression tests?
If not,  any idea how we can add some tests for this?

bitkarrot:
@motorina0 not sure if there is a docker container for ZBD,  Possibly @andrerfneves would know. Btw, @andrerfneves it would be really helpful if the ZBD API docs included all the possible states for the error and success conditions. Right now I have to guess what they might be and I don't know if there are other conditions that could potentially be missing. 

As for testing: I've been just writing a basic aiohttp/async python script (didn't check it in) to test if my calls are accurate before writing them into the LNBits class.

Ideally the LNBits project would have some standard for pytests or something similar on every wallet integration point. Since there isn't any specific directive, I could, also just make a test script up and put it into the test directory that currently exists for core. What do you think? Do you have a standard you want me to create or follow, @motorina0 Whatever is easiest for you. 


> @bitkarrot is there a docker container for ZBD that we can use in our regression tests? If not, any idea how we can add some tests for this?

andrerfneves:
> Btw, @andrerfneves it would be really helpful if the ZBD API docs included all the possible states for the error and success conditions. Right now I have to guess what they might be and I don't know if there are other conditions that could potentially be missing.

Say no more. I will task the team to revisit this. I'm assuming you're talking about `status` property specifically correct?

> @motorina0 not sure if there is a docker container for ZBD, Possibly @andrerfneves would know.

Unfortunately we do not have this capability.

bitkarrot:
> Say no more. I will task the team to revisit this. I'm assuming you're talking about `status` property specifically correct?
> 

@andrerfneves  Yes please. Here is a good example from another API:

image

@motorina0 I will try to squeeze out a pytest sometime this weekend or by early next. something that can be templated to other API integrations as well :)

rsafier:
ACK, built docker locally, setup superuser, popped in my ZBD API Key and tested receive and send with no issues.

prusnak:
Let's merge so we don't have to keep rebasing it. // cc @dni

bitkarrot:
Will make pytest a separate PR that can be applied to all Wallet integrations,
 if @prusnak has itchy fingers and can't wait to merge this one. ;)

lnbits/wallets/zbd.py
```diff
@@ -0,0 +1,171 @@
...
+    """https://zbd.dev/api-reference/"""
+
+    def __init__(self):
+        if not settings.zbd_api_endpoint:
+            raise ValueError("cannot initialize ZBDWallet: missing zbd_api_endpoint")
+        if not settings.zbd_api_key:
+            raise ValueError("cannot initialize ZBDWallet: missing zbd_api_key")
+
+        self.endpoint = self.normalize_endpoint(settings.zbd_api_endpoint)
+        self.auth = {
```
motorina0, line 29: - any reason to make `auth `a field on this class? I do not see `self.auth` referenced anywhere else.
- nitpick: rename to `headers = { ...`

lnbits/wallets/zbd.py
```diff
@@ -0,0 +1,171 @@
...
+        r = await self.client.post(
+            "payments",
+            json={
+                "invoice": bolt11,
+                "description": "",
+                "amount": "",
+                "internalId": "",
+                "callbackUrl": "",
+            },
+            timeout=None,
```
motorina0, line 103: maybe a timeout should be set though

lnbits/wallets/zbd.py
```diff
@@ -0,0 +1,171 @@
...
+        )
+
+        if r.is_error:
+            error_message = r.json()["message"]
+            return PaymentResponse(False, None, None, None, error_message)
+
+        data = r.json()
+
+        # get the payment hash from the zbd api
+        decoded_request = await self.client.post(
```
motorina0, line 113: Can we avoid this second HTTP call?
The previous HTTP call (https://zbd.dev/api-reference/payments/send) provides us the `preimage` and the `fee`.  The payment hash can be extracted using:

```py
from lnbits import bolt11
invoice = bolt11.decode(bolt11)
# invoice.payment_hash
```

lnbits/wallets/zbd.py
```diff
@@ -0,0 +1,171 @@
...
+            "pending": None,
+            "completed": True,
+            "error": None,
+            "expired": False,
+            "failed": False,
+        }
+
+        return PaymentStatus(statuses[data.get("status")], fee_msat=None, preimage=None)
+
+    async def paid_invoices_stream(self) -> AsyncGenerator[str, None]:
```
motorina0, line 163: If `ZBD webhook listener disabled` then the body of this method is not relevant.
Suggestion: just `log` and `return`

motorina0 (approved):
LGTM
Thank you for adding this implementation.

The only open topic is: https://github.com/lnbits/lnbits/pull/2235/files#r1497611632

lnbits/wallets/zbd.py
bitkarrot: set timeout to 40 like others?

lnbits/wallets/zbd.py
bitkarrot: ok

lnbits/wallets/zbd.py
bitkarrot: OK.

bitkarrot:
OK changes requested by @motorina0 done.

bitkarrot:
> Let's merge so we don't have to keep rebasing it. // cc @dni

Um, yes. I tried to resolve the conflicts in the bundle.min.js but gave up, not sure how to do this best way. 
@prusnak

prusnak:
> Um, yes. I tried to resolve the conflicts in the bundle.min.js but gave up, not sure how to do this best way.

Easiest way is to just drop the `chore: make bundle` during the rebase and recreate it again from scratch by running `make bundle` - that's also the reason why you want to have it as a separate commit.

PS: I just did the process above to resolve the conflicts.

prusnak (approved):

prusnak:
Great work, merged!

bitkarrot:
Thank you! will do that next time :D
> 
> Easiest way is to just drop the `chore: make bundle` during the rebase and recreate it again from scratch by running `make bundle` - that's also the reason why you want to have it as a separate commit.
> 
> PS: I just did the process above to resolve the conflicts.

PR #2247, opened 2024-02-08
chore: update poetry lock file

dni:
currently there is an inconsistency on dev in the `poetry.lock` file


prusnak:
Can you fix the inconsistency with

poetry lock --no-update

instead of

poetry lock

?

motorina0:
How was this issue manifested? I would have expected `poetry install` to fail (that is why the hash is there).

prusnak (approved):
I got the same hash by calling `poetry lock --no-update`

PR #2251, opened 2024-02-09
Improves payment reactions

arcbtc:
* only triggers on incoming
* choice between different reactions


motorina0:
Is this PR a replacement for https://github.com/lnbits/lnbits/pull/2111?

lnbits/static/js/base.js
```diff
@@ -524,12 +523,7 @@ window.windowMixin = {
     } else {
       this.$q.dark.set(true)
     }
-    let reactions = this.$q.localStorage.getItem('lnbits.reactions')
-    if (reactions == null) {
-      this.$q.localStorage.set('lnbits.reactions', true)
-      reactions = true
-    }
-    this.reactionToggle = reactions
+    this.reactionChoice = this.$q.localStorage.getItem('lnbits.reactions')
```
motorina0, line 526:
```suggestion
    this.reactionChoice = this.$q.localStorage.getItem('lnbits.reactions') || 'confettiBothSides'
```

- since the declaration is `reactionChoice: 'confettiBothSides'` I assume this should be the default value.

lnbits/static/js/event-reactions.js
```diff
@@ -1,16 +1,30 @@
-function eventReactionWebocket(event_id, event_function) {
+function eventReactionWebocket(event_id) {
   localUrl = ''
+  reaction = localStorage.getItem('lnbits.reactions')
+  if (reaction == null || reaction == 'None' || reaction == undefined) {
```
motorina0, line 4:
```suggestion
  if (!reaction) {
```

I don't see how `reaction` can be `None`.

arcbtc: Its an option in the account section, if you don't want reactions

motorina0:
- why is `package-lock.json` updated, but `package.json` not
- `package-lock.json` now has `versinon: 1`, previously it was `version: 3`

lnbits/static/js/event-reactions.js
motorina0: nitpick:
```suggestion
  if (!reaction || reaction === 'None') {
```

lnbits/static/js/event-reactions.js
```diff
@@ -1,16 +1,30 @@
...
   connection = new WebSocket(localUrl)
   connection.onmessage = function (e) {
-    event_function()
+    let parsedData = JSON.parse(e.data)
+    if (parsedData.payment.amount < 0) {
+      return
+    } else {
+      try {
+        reaction = localStorage.getItem('lnbits.reactions')
+        eval(reaction.split('|')[1] + '()')
```
motorina0, line 20: `eval` is dangerous, we should not use it:
https://www.digitalocean.com/community/tutorials/js-eval

talvasconcelos (approved):

lnbits/static/js/event-reactions.js
```diff
@@ -1,16 +1,30 @@
...
+    if (parsedData.payment.amount < 0) {
+      return
+    } else {
+      try {
+        reaction = localStorage.getItem('lnbits.reactions')
+        window[reaction.split('|')[1]]()
+      } catch (e) {
+        console.log(e)
+      }
+    }
```
motorina0, line 24:
```suggestion
  connection.onmessage = function (e) {
    try {
      const parsedData = JSON.parse(e.data)
      if (parsedData.payment.amount < 0) {
        return
      }
      reaction = localStorage.getItem('lnbits.reactions')
      if (reaction) {
        window[reaction.split('|')[1]]()
      }
    } catch (e) {
      console.log(e)
    }
  }
```

- `data` might not be valid json
- `reactions` can be `null`

lnbits/static/js/event-reactions.js
```diff
@@ -1,16 +1,30 @@
...
+  reaction = localStorage.getItem('lnbits.reactions')
+  if (!reaction || reaction === 'None') {
+    return
+  }
   if (location.protocol !== 'http:') {
     localUrl = 'wss://' + location.host + '/api/v1/ws/' + event_id
   } else {
     localUrl = 'ws://' + location.host + '/api/v1/ws/' + event_id
   }
   connection = new WebSocket(localUrl)
```
motorina0, line 12: when is this websocket closed?

Suggestion: close if the expected message is received or after a time interval.

lnbits/static/js/base.js
```diff
@@ -444,7 +443,7 @@ window.windowMixin = {
   data: function () {
     return {
       toggleSubs: true,
-      reactionToggle: true,
+      reactionChoice: 'confettiBothSides',
```
motorina0, line 446: shouldn't  this field (and the init below) be in `account.js`?

dni (approved):

Issue #2253, opened 2024-02-09
[BUG] own site description disappered


iWarpBTC:
**Describe the bug**
With 0.12.0 there's a new index page which now shows the generic desc regardless the `SITE_TITLE` setting. 

**Expected behavior**
Custom `LNBITS_SITE_DESCRIPTION` shows when `SITE_TITLE` is changed as it was before.

motorina0:
Fixed in: https://github.com/lnbits/lnbits/pull/2254

PR #2254, opened 2024-02-09
fix: custom site desc

iWarpBTC:
My solution to #2253 bug.

Please feel free to suggest any better solution since I'm really not much experienced with this Quasar positioning and styling.


arcbtc:
Good fix, thanks!

motorina0 (approved):

motorina0:
tACK

Issue #2258, opened 2024-02-09
[BUG] Node Management / Transactions - Expiry time are years


tlindi:
**Describe the bug**
Node Management / Transactions - Expiry time are years

**To Reproduce**

1. Go to 'Enable node management'
2. Click on 'Node Management '
3. Scroll down to 'Transactions '
4. See invoice expiry times are carbage

**Expected behavior**
Show reasonable expiry times of invoices.
I'd estimate LN invoices should expire in less than hundrest or tens of thousands of years. See 

**Screenshots**
![image](https://github.com/lnbits/lnbits/assets/7249767/f8f3ddf6-5cb8-47c4-a4f9-5bff8fb93217)

**Desktop (please complete the following information):**

- LNbits version: 0.11.1
- Database sqlite

**Additional context**
I have not tested on more recent versions, and I have no possibility to make such tests.

tlindi:
0.12.1 - still the same

jackstar12:
@tlindi  What node are you using?

tlindi:
LND on MyNode from separate host machine with LNDRestWallet Funding.

PR #2271, opened 2024-02-13
fix: correct invoice expiry for lndrest

jackstar12:
closes: #2258

The LND api returns the timestamps as strings which wasnt caught because strings can be appended and pydantic just converted it to an integer by itself.


dni (approved):

tlindi:
TY

PR #2280, opened 2024-02-15
feat: deprecate get_current_extension_name

dni:
DEPRECATED: Use the repo name instead, will be removed in the future
    before: `register_invoice_listener(invoice_queue, get_current_extension_name())`
    after: `register_invoice_listener(invoice_queue, "my-extension")`


motorina0 (approved):
docs only, approved

prusnak (approved):
utACK

PR #2281, opened 2024-02-15
fix: extension stop background work

motorina0:
Related to:
 - https://github.com/lnbits/splitpayments/issues/14
 - https://github.com/lnbits/scrub/issues/7
 - https://github.com/lnbits/lnbits/pull/2278

When an extension is uninstalled/upgraded the background tasks have to be stopped.

Issue:
 - the background tasks were being stopped by calling a REST API endpoint on `localhost`
 - `localhost` is not always accessible (docker container for example)

Solution:
 - first try to call the `stop` logic using a normal python function call

### DIFF
--- a/lnbits/core/helpers.py
+++ b/lnbits/core/helpers.py
@@ -64,21 +64,20 @@ async def stop_extension_background_work(
 
 
 async def _stop_extension_background_work(ext_id) -> bool:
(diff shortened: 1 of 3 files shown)


dni (approved):
LGTM

talvasconcelos (approved):

PR #2283, opened 2024-02-19
feat: add node url to api keys & docs

### LINKED ISSUE #2277: [Feature request] Add server url to "API keys and API docs" section
When linking lnbits with external services, (e.g. [zaprite](https://zaprite.com/)) one needs to specify two things: node url and invoice key. 

image

Invoice key is clearly visible in the "API keys and API docs" section, but it's sometimes unclear what my "LNbits Node URL" is. 

image

Display "LNbits Node URL" in "Node URL, API keys and docs"

dni:
closes #2277

@prusnak what would you say would be the easiest way to change that label for all languages.  just remove the key for all languages and run  your script?


lnbits/static/js/wallet.js
```diff
@@ -90,6 +90,7 @@ new Vue({
   mixins: [windowMixin],
   data: function () {
     return {
+      origin: window.location.origin,
```
prusnak, line 93: I don't think this is correct - when LNbits is deployed to https://foobar.example/lnbits then `window.location.origin` returns just `https://foobar.example` and not `https://foobar.example/lnbits` - or is this type of deployment unsupported?

prusnak:
> just remove the key for all languages and run your script?

Yes, I just did that in 72572b5 (which also added some other new strings)

prusnak (approved):

motorina0 (approved):

PR #2288, opened 2024-02-21
refactor: remove unused webhook handler for wallets

dni:
it was only used in lnpay, opennode and alby which all had it disabled anayways with a pretty old TODO, so i guess the feature is disabled for a y ear now and nobody used it, so i just removed it


motorina0 (approved):
LGTM
nice clean-up

PR #2289, opened 2024-02-21
fix: raise failed request to not run `mark_webhook_sent`

dni:
error is not raised, even if failed webhook would be marked as sent every time

### DIFF
--- a/lnbits/core/tasks.py
+++ b/lnbits/core/tasks.py
@@ -51,7 +51,7 @@ async def killswitch_task():
                                 "Switching to VoidWallet. Killswitch triggered."
                             )
                             await switch_to_voidwallet()
-                except (httpx.ConnectError, httpx.RequestError):
+                except (httpx.ConnectError, httpx.RequestError, httpx.HTTPStatusError):
                     logger.error(
                         "Cannot fetch lnbits status manifest."
                         f" {settings.lnbits_status_manifest}"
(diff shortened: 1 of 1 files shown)


motorina0:
In `dispatch_webhook` the status code is set to `-1` in case of failure. We should be consistent in handling the error.

```py
        try:
            r = await client.post(payment.webhook, json=data, timeout=40)
            await mark_webhook_sent(payment, r.status_code)
        except (httpx.ConnectError, httpx.RequestError):
            await mark_webhook_sent(payment, -1)
```

There are 3 places where this pattern ☝️ is repeated. 
A common function should be extracted: `call_webhook(payment, data, timeout)`

lnbits/core/tasks.py
```diff
@@ -51,7 +51,7 @@ async def killswitch_task():
                                 "Switching to VoidWallet. Killswitch triggered."
                             )
                             await switch_to_voidwallet()
-                except (httpx.ConnectError, httpx.RequestError):
+                except (httpx.RequestError, httpx.HTTPStatusError):
```
motorina0, line 54: Is `httpx.ConnectError` not ever thrown?
Can we just catch `Error` and log the actual error?
Is there a case when the error should bubble up?

lnbits/core/tasks.py
```diff
@@ -121,9 +121,18 @@ async def wait_for_paid_invoices(invoice_paid_queue: asyncio.Queue):
             async with httpx.AsyncClient(headers=headers) as client:
                 try:
                     r = await client.post(url, timeout=4)
-                    await mark_webhook_sent(payment, r.status_code)
-                except (httpx.ConnectError, httpx.RequestError):
-                    pass
+                    r.raise_for_status()
+                    await mark_webhook_sent(payment.payment_hash, r.status_code)
+                except httpx.HTTPStatusError as exc:
```
motorina0, line 126: Which is the difference between the two catch clauses (`httpx.HTTPStatusError` vs `httpx.RequestError`), why are they handled differently?

motorina0: why are they handled differently though?

lnbits/core/tasks.py
dni: ConnectError is capture bei Request error look at the exception tree here. it was redundant. https://www.python-httpx.org/exceptions/

dni: statuserror  is triggerbe by `raise_for_status()` and is raised if the response has a bad status code

lnbits/core/tasks.py
dni: because we wannt to save the bad statuscode inside the payment details, it was like that before. so we now we did send it, but the receiver couldnt handle it.

motorina0 (approved):

PR #2290, opened 2024-02-21
refactor: move registering of tasks into `app.py`

dni:
let the app have the responsibility to create the tasks not a helper function inside tasks.py. makes it way easier to follow on how those background tasks are ran.


arcbtc (approved):

motorina0 (approved):
LGTM

PR #2291, opened 2024-02-21
fix: raise proper exception in services pay_invoice

dni:
also values where checked twice

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -189,23 +189,21 @@ async def pay_invoice(
     If the payment is still in flight, we hope that some other process
     will regularly check for the payment.
     """
-    invoice = bolt11_decode(payment_request)
+    try:
+        invoice = bolt11_decode(payment_request)
+    except Exception:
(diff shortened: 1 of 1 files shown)


lnbits/core/services.py
```diff
@@ -189,24 +189,22 @@ async def pay_invoice(
...
-        raise ValueError("Amount in invoice is too high.")
+        raise InvoiceFailure("Amount in invoice is too high.")
 
     await check_wallet_limits(wallet_id, conn, invoice.amount_msat)
 
     async with db.reuse_conn(conn) if conn else db.connect() as conn:
         temp_id = invoice.payment_hash
         internal_id = f"internal_{invoice.payment_hash}"
 
-        if invoice.amount_msat == 0:
```
motorina0, line 205: why is this removed (more than a refactor)?

dni: its actually already check in line 197 `not invoice.amount_msat > 0`

motorina0 (approved):

prusnak (approved):

PR #2293, opened 2024-02-23
chore: remove lnurl wallet and withdraw feature

dni:
i think this feature is undocumented and the code is very outdated. i don't think it is worth to keep.

looking at the `/lnurlwallet` endpoint for example, it creates a new user and wallet without even checking if the creation of users is allowed.

WIP just did this to see how much code is related to this features


arcbtc:
nack
This is a really useful feature used by atms, and faucets

prusnak:
> This is a really useful feature used by atms, and faucets

Shouldn't they use LNURL-w extension for that?

arcbtc:
Actually this might not be what I am thinking of, is it the auto wallet create thingmabob?

arcbtc:
> > This is a really useful feature used by atms, and faucets
> 
> Shouldn't they use LNURL-w extension for that?

I thought this was the auto-wallet builder endpoint, my mistake

arcbtc:
No it is that endpoint!
Thiis is useful, you can add an lnurl prefix so if scanned and opened in browser it will creata a wallet and fund it with the lnurl

dni:
screenshot-1711462299

prusnak:
Jmeter CI job is failing

arcbtc (approved):

talvasconcelos (approved):

PR #2307, opened 2024-03-02
update black to 24.2.0 and ruff to 0.3.0

prusnak:


pyproject.toml
```diff
@@ -152,53 +152,31 @@ extend-exclude = """(
...
+# W - pycodestyle warnings
+# I - isort
+select = ["F", "E", "W", "I"]
 ignore = []
 
 # Allow autofix for all enabled rules (when `--fix`) is provided.
 fixable = ["ALL"]
 unfixable = []
 
-# Exclude a variety of commonly ignored directories.
```
motorina0, line 164: How come these directories do not have to be excluded anymore?

prusnak: because i changed `exclude` to `extend-exclude` which keeps the default excluded dirs only extends it

motorina0: nice!

motorina0 (approved):

arcbtc (approved):

PR #2313, opened 2024-03-08
fix: enforce order of payments

motorina0:
### Summary
The comment (now removed) explains the problem:

```python
    # we have to specify the wallet id here, because postgres and sqlite return
    # internal payments in different order and get_standalone_payment otherwise
    # just fetches the first one, causing unpredictable results
```

The problem is that by specifying the `wallet id` the order is not enforced **if the same wallet is the sender and the receiver**.
An explicit `ORDER BY` has been added to the SQL statement.

### DIFF
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -554,7 +554,8 @@ async def get_wallet_for_key(
     row = await (conn or db).fetchone(
         """
         SELECT *, COALESCE((SELECT balance FROM balances WHERE wallet = wallets.id), 0)
-        AS balance_msat FROM wallets WHERE adminkey = ? OR inkey = ?
+        AS balance_msat FROM wallets
+        WHERE (adminkey = ? OR inkey = ?) AND deleted = false
         """,
         (key, key),
     )
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -602,6 +602,7 @@ async def get_standalone_payment(
         SELECT *
         FROM apipayments
         WHERE {clause}
+        ORDER BY amount
         LIMIT 1
         """,
         tuple(values),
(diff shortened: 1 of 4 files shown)


lnbits/core/crud.py
```diff
@@ -602,6 +602,7 @@ async def get_standalone_payment(
         SELECT *
         FROM apipayments
         WHERE {clause}
+        ORDER BY amount
```
prusnak, line 605: I don't understand the logic why sorting by amount is desired and not by some other field (e.g. time comes to my mind).

dni: i think its for internal payments, the one sending is negative the one receiving positive

dni (approved):

lnbits/core/crud.py
motorina0: The original issue was that the `extra` field was sometimes not correctly fetched for get payment (`/api/v1/payments/${paymenHash}`). The `extra` field contains (among others the webhook status).

The `extra` field is updated like so on `core` and `withdraw` extension:
```py
            await update_payment_extra(
                payment_hash=payment_hash,
                extra={
                    "wh_success": r.is_success,
                    "wh_message": r.reason_phrase,
                    "wh_response": r.text,
                },
                outgoing=True,
            )
```

Having `outgoing=True` makes the update on the outgoing (negative) payment.
telegram-cloud-photo-size-4-6006064172485427192-y

However, now I have just noticed that [the call from lnurlp extension](https://github.com/lnbits/lnurlp/blob/main/tasks.py#L95-L106) does not specify ` outgoing=True`.
This is indeed inconsistent (and fragile).

motorina0: > However, now I have just noticed that [the call from lnurlp extension](https://github.com/lnbits/lnurlp/blob/main/tasks.py#L95-L106) does not specify  outgoing=True.
This is indeed inconsistent (and fragile).

- nvm, `lnurlp` is not an outgoing payment. Its fine as it is.

dni (approved):
lgtm

dni (approved):

talvasconcelos (approved):

PR #2322, opened 2024-03-12
refactor: dev-friendly payment status name

motorina0:
The first parameter of `PaymentStatus` indicates the actual state (`success=True`, `failed=False`, `pending=None`).

This is hard to read and remember.

This PR uses 3 new **explicit** subclasses:

```py

class PaymentSuccessStatus(PaymentStatus):
    paid = True


class PaymentFailedStatus(PaymentStatus):
    paid = False


class PaymentPendingStatus(PaymentStatus):
    paid = None
```

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -31,7 +31,7 @@ from lnbits.settings import (
 )
 from lnbits.utils.exchange_rates import fiat_amount_as_satoshis, satoshis_amount_as_fiat
 from lnbits.wallets import FAKE_WALLET, get_wallet_class, set_wallet_class
-from lnbits.wallets.base import PaymentResponse, PaymentStatus
+from lnbits.wallets.base import PaymentPendingStatus, PaymentResponse, PaymentStatus
 
 from .crud import (
     check_internal,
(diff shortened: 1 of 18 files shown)


lnbits/core/services.py
```diff
@@ -577,7 +577,7 @@ async def check_transaction_status(
         wallet_id, payment_hash, conn=conn
     )
     if not payment:
-        return PaymentStatus(None)
+        return PaymentPendingStatus()
     if not payment.pending:
         # note: before, we still checked the status of the payment again
         return PaymentStatus(True, fee_msat=payment.fee)
```
prusnak, line 583: Shouldn't we also change this?

```suggestion
        return PaymentSuccessStatus(fee_msat=payment.fee)
```

So we will remove all usages of `PaymentStatus` from the codebase and just use the specific classes and never the generic one?

motorina0: Done in commit: https://github.com/lnbits/lnbits/pull/2322/commits/575bc5154bc979a75cb88cc17aa76c3832338a85

prusnak:
Fair enough 👍

One nitpick though:

I find PaymentStatusPending, PaymentStatusFailed and PaymentStatusSuccess better than PaymentPendingStatus, PaymentFailedStatus and PaymentSuccessStatus. Any thoughts?

motorina0:
> I find PaymentStatusPending, PaymentStatusFailed and PaymentStatusSuccess better than PaymentPendingStatus, PaymentFailedStatus and PaymentSuccessStatus. Any thoughts?

I agree, will be added to the https://github.com/lnbits/lnbits/pull/2323 together with `PaymentResponse` changes.

prusnak (approved):
utACK

dni (approved):

PR #2324, opened 2024-03-14
fix: pending property for PaymentStatus

motorina0:
A payment status is pending when the `paid` value is none.

In the old implementation if a payment was failed (`paid == False`) then the `pending` property would be `True` (which is incorrect)


motorina0:
fixing failed jobs...

arcbtc (approved):

PR #2327, opened 2024-03-17
bug: recover user for export-to-phone

arbadacarbaYK:
Is it that easy ? 😅🤠


dni (approved):
yes :D, we wanted to remove the userid from the freontend though, but in that case without it that qr feature would never work

talvasconcelos (approved):

PR #2341, opened 2024-03-21
feat: complete use of `LNBITS_EXTENSIONS_DEACTIVATE_ALL`

dni:
- add to admin ui
- move to `ExtensionsSettings`
- hide left navbar when disabled
- do not register extension routes on startup
- do not check for installed extensions on startup
screenshot-1711036830
screenshot-1711036664


talvasconcelos (approved):

arbadacarbaYK:
Does this affect paid extensions ? Is the payment to be redone if reactivated ?
Are API calls for extensions also prevented ?

dni:
disabled all extension and it lets you run lnbits without them, maybe if you don't need any.
its most useful when developing on core.

lnbits/app.py
```diff
@@ -396,11 +385,16 @@ async def lnbits_startup():
...
             await check_funding_source()
 
-            # check extensions after restart
-            await check_installed_extensions(app)
+            init_core_routers(app)
 
-            # register core and extension routes
-            register_routes(app)
+            # check extensions after restart
+            if not settings.lnbits_extensions_deactivate_all:
```
motorina0, line 391: was this not checked before?

dni: nope. :) i did not see that

lnbits/app.py
```diff
@@ -285,17 +285,6 @@ async def restore_installed_extension(app: FastAPI, ext: InstallableExtension):
         ext.nofiy_upgrade()
 
 
-def register_routes(app: FastAPI) -> None:
```
motorina0, line 288: - why remove this method?
- it is better to have many small methods with well defined names rather than bigger methods which are hard to follow

dni: in this case i absolutely disagree, register_routes mixing up registering extension routes with initializing the core routers.

helper function for helper functions are hard to follow because you jump around everywhere in the code and makes it so you cant follow it!

it is a balance and it this case it went 1 abstraction to far

motorina0 (approved):

PR #2346, opened 2024-03-22
hotfix: initial currency conversion on balance

dni:
fiat amount not shown in balance on initial pageload, lets reintroduce the first fetch_balance to fix it.

maybe in the future balance call and wallet itself should give the current fiat value if the wallet uses fiat tracking so you dont need the extra conversion call

### DIFF
--- a/lnbits/static/js/wallet.js
+++ b/lnbits/static/js/wallet.js
@@ -818,10 +818,8 @@ new Vue({
     }
   },
   watch: {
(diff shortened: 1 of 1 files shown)


motorina0:
Please also remove initial balance assignment from `created`

dni:
done!

motorina0 (approved):

talvasconcelos (approved):

PR #2347, opened 2024-03-22
feat: remove bundle

dni:
breaking change in how to install lnbits
you will need to run `make build` instead of `poetry install` and it required to install nodejs for building
after an update you now should also run `make build`


arcbtc:
So make will bundle?

dni:
yep that was the idea

arcbtc:
Will this impact nix @prusnak ?

dni:
decided its not worth it to force the user to have npm installed

Issue #2194, opened 2023-12-25
[BUG] Wrong "Expiry" in finalized service fee wallet


DoktorShift:
**Describe the bug**
I have setup a service fee. When i look into wallet where the service fees arrive and check the internal payment details of the finalized transaction i can see a expiry date 01/01/1970.

**Screenshots**

![service fee expiry](https://github.com/lnbits/lnbits/assets/106493492/ffe9b182-3b7d-4c76-a5e7-6e840c1060fa)

- LNbits version: 0.11.3
- Database sqlite

**Additional context**
Service fee is working for me like intended. No problems spotted so far.

PR #2349, opened 2024-03-22
bug:  When "expiry" is null, do not display "expiry" in payment details

dethos:
Fixes: #2194 

`null` is a valid value for the `expiry` field of payment records. Which means that when this value is not populated, `payment.expirydate` and `payment.expirydateFrom` will be generated from an invalid `Date` (`new Date(null*1000)`, which renders to 1970-01-01 or 1969-12-31 depending on the timezone). 

This PR hides the `expiry` line from the payment details, when there is no information about it, just like it is done with `preimage`.

_Note: for the change to be effective, we need to execute `make bundle`. However, it is not clear whether a new generated bundle should be included with every PR or not._


dni:
cACK

but somethign still wrong with the bundle, as of right now we still include it in prs

dethos:
> thanks for the contribution! cACK
> 
> but something still wrong with the bundle, as of right now we still include it in prs

Fixed

dni (approved):

motorina0 (approved):
LGTM

PR #2354, opened 2024-03-24
feat: add commands `delete-walle`, `delete-wallet-payment` and `verbose` option, 

motorina0:
add commands:
  - `delete-wallet`
  - `delete-wallet-payment`

add `-verbose` option for check invalid payments


motorina0:
Merging without approval.

PR #2356, opened 2024-03-24
simplify description in i18n

prusnak:
so we[1] do not have to always edit description when a new funding source is added

[1] and by "we" I obviously mean bitkarrot :D


dni (approved):

bitkarrot:
ACK!

motorina0 (approved):

PR #2361, opened 2024-03-25
perf: cln faster init

motorina0:
The call`data = self.ln.listinvoices()` lists all invoices on the CLN node. This takes ~10min on a big node like legend.

The solution is to fetch the last created invoice from the DB and call `listinvoices` for that respective hash (which should be much faster)

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -26,7 +26,7 @@ from starlette.responses import JSONResponse
 from lnbits.core.crud import (
     get_dbversions,
     get_installed_extensions,
-    get_last_incomming_payment,
+    get_last_incoming_payment,
 )
 from lnbits.core.helpers import migrate_extension_database
 from lnbits.core.services import websocketUpdater
@@ -461,7 +461,7 @@ def log_server_info():
 
 async def init_latest_payment_hash():
     try:
-        last_payment = await get_last_incomming_payment()
+        last_payment = await get_last_incoming_payment()
         if last_payment:
             settings.last_payment_hash_on_start = last_payment.checking_id
     except Exception as e:
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -23,7 +23,11 @@ from slowapi.util import get_remote_address
 from starlette.middleware.sessions import SessionMiddleware
 from starlette.responses import JSONResponse
 
-from lnbits.core.crud import get_dbversions, get_installed_extensions
+from lnbits.core.crud import (
+    get_dbversions,
+    get_installed_extensions,
+    get_last_incomming_payment,
+)
 from lnbits.core.helpers import migrate_extension_database
 from lnbits.core.services import websocketUpdater
 from lnbits.core.tasks import (  # watchdog_task
@@ -383,6 +387,7 @@ def register_startup(app: FastAPI):
             await check_webpush_settings()
 
             log_server_info()
+            await init_latest_payment_hash()
 
             # initialize WALLET
             try:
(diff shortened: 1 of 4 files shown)


lnbits/wallets/corelightning.py
```diff
@@ -46,12 +46,17 @@ def __init__(self):
...
-        # invoices: dict = self.ln.listinvoices()  # type: ignore
-        # for inv in invoices["invoices"][::-1]:
-        #     if "pay_index" in inv:
-        #         self.last_pay_index = inv["pay_index"]
-        #         break
+        payment_hash = settings.last_payment_hash_on_start
+        data: dict = self.ln.listinvoices(payment_hash=payment_hash)  # type: ignore
+        if len(data["invoices"]) == 0:
+            logger.warning("Last payment hash was not fund on the funding source.")
+            logger.warning("Listing all invoices, this might take serveral minutes.")
```
dni, line 53:
```suggestion
            logger.warning("Listing all invoices, this can take some minutes...")
```

lnbits/core/crud.py
```diff
@@ -642,6 +642,24 @@ async def get_wallet_payment(
     return Payment.from_row(row) if row else None
 
 
+async def get_last_incomming_payment(
```
dni, line 645:
```suggestion
async def get_last_incoming_payment(
```

lnbits/app.py
```diff
@@ -454,6 +459,16 @@ def log_server_info():
     logger.info(f"Service fee wallet: {settings.lnbits_service_fee_wallet}")
 
 
+async def init_latest_payment_hash():
+    try:
+        last_payment = await get_last_incomming_payment()
```
dni, line 464:
```suggestion
        last_payment = await get_last_incoming_payment()
```

lnbits/app.py
```diff
@@ -454,6 +459,16 @@ def log_server_info():
...
+        if last_payment:
+            settings.last_payment_hash_on_start = last_payment.checking_id
+    except Exception as e:
+        logger.error(str(e))
+        raise ImportError("Failed to fetch latest incomming payment.")
```
dni, line 469:
```suggestion
        raise ImportError("Failed to fetch latest incoming payment.")
```

lnbits/wallets/corelightning.py
```diff
@@ -45,8 +45,17 @@ def __init__(self):
...
+        payment_hash = settings.last_payment_hash_on_start
+        data: dict = self.ln.listinvoices(payment_hash=payment_hash)  # type: ignore
+        if len(data["invoices"]) == 0:
+            logger.warning(
+                "Last payment hash was not found on the Core Lightning Node."
+            )
+            logger.warning("Listing all invoices, this can take some minutes...")
+
+            data = self.ln.listinvoices()  # type: ignore
+        for inv in data["invoices"][::-1]:
```
dni, line 58: what does `data["invoices"][::-1]:` mean? i know it was there before but its hard to read

lnbits/app.py
```diff
@@ -454,6 +459,16 @@ def log_server_info():
     logger.info(f"Service fee wallet: {settings.lnbits_service_fee_wallet}")
 
 
+async def init_latest_payment_hash():
```
dni, line 462: why not running the crud function directly where it is needed, i dont see the point of prefetching it into settings?
you could utilize the cache utility for this instead of settings

motorina0: - I didn't want to introduce a dependency from `wallet` to `crud`
- it is a minor/fast operation

lnbits/wallets/corelightning.py
motorina0: it reverts the list, then loops over it

```py
>>> a = ["x", "y", "z"]
>>> print(a[::-1])
['z', 'y', 'x']
```

dni: lets use `reversed()` makes it easier to understand

PR #2371, opened 2024-03-27
[test] add jmeter tests to install and enable all vetted extensions

motorina0:
This PR introduces a new JMeter test. It does not change the code.

This test will check that:
 - there are no backwards incompatible changes in `core` that affect the extensions (eg: some functions moved in `core` from one file to another and the extension fails to install)
 - the vetted extensions do not use deprecated APIs


dni (approved):
LGTM! nice

talvasconcelos (approved):

PR #2381, opened 2024-04-01
Clarified top-up success msg 

arbadacarbaYK:
Common misconceptioin is that the top up is related to the funds on the funding source. 

Success msg  extended with info that correlated funds on funding source are needed and the amount is virtual until fitting.


dni:
i think the explanation should be in the frontend not the API. i could also be translated there.

arbadacarbaYK:
> i think the explanation should be in the frontend not the API. i could also be translated there.

Isnt this part the success-message that is shown in the frontend ? Thats what i wanted to edit 😃

lnbits/core/views/admin_api.py
```diff
@@ -125,7 +125,10 @@ async def api_topup_balance(data: CreateTopup) -> dict[str, str]:
 
     await update_wallet_balance(wallet_id=data.id, amount=int(data.amount))
 
-    return {"status": "Success"}
+    return {
+        "status": "Success creating virtual funds."
+        "Payments depend on actual funds on funding source."
```
dni, line 130:
```suggestion
        "Payments are dependent on the actual funds available in the funding source."
```

motorina0:
Fixed:
image

Note: i18n does not work on `admin.js` (yet).

lnbits/static/js/base.js
```diff
@@ -144,12 +144,15 @@ window.LNbits = {
           id: wallet_id
         })
         .then(_ => {
+          credit = parseInt(credit)
           Quasar.Notify.create({
```
dni, line 148: this notify should not be in base.js, this function just should return the promise

lnbits/static/js/admin.js
```diff
@@ -199,8 +199,9 @@ new Vue({
         .then(response => {
```
dni, line 199: just like here the notification should put off here and also translated here

lnbits/core/views/admin_api.py
arbadacarbaYK: > 

was extra making it short cause long msg are hardly readable but thats also good.

dni: ok

dni (approved):
thanks! that small wording change actually improved the code in the end :)

lnbits/static/js/components.js
```diff
@@ -691,9 +691,25 @@ Vue.component('lnbits-update-balance', {
...
-        this.callback({value: res, wallet_id: this.wallet_id})
-      })
+      LNbits.api
+        .updateBalance(credit, this.wallet_id)
+        .then(res => {
+          this.callback({value: res, wallet_id: this.wallet_id})
+        })
+        .then(_ => {
+          credit = parseInt(credit)
+          Quasar.Notify.create({
```
dni, line 701: this is bit notpicky, but couldnt it be `this.$q.notify` also like we have in other places in the code?

lnbits/static/i18n/en.js
```diff
@@ -33,6 +33,8 @@ window.localisation.en = {
   reset_defaults_tooltip: 'Delete all settings and reset to defaults.',
   download_backup: 'Download database backup',
   name_your_wallet: 'Name your %{name} wallet',
+  wallet_topup_ok:
+    'Success creating virtual funds (%{amount} sats). Payments depend on actual funds on funding source.',
```
arbadacarbaYK, line 37: shouldnt this say "(%{credit} sats)" in each translation ?

dni (approved):

lnbits/static/i18n/en.js
motorina0: no, there is an object which is passed as param:

```js
            message: this.$t('wallet_topup_ok', {
              amount: credit
            }),
```

dni: no, amount is fine, if you look into the `this. $q.notify()`

motorina0 (approved):

Issue #2308, opened 2024-03-02
enable more checks for ruff


prusnak:
Currently we enable the F, E, W and I checks for ruff.

There are more we could enable (and fix the reports), such as:
- C - mccabe
- N - naming
- UP - pyupgrade
- B - flake8-bugbear
- A - flake8-builtins
- RUF - ruff specific

More info: https://docs.astral.sh/ruff/rules/

dni:
lets go for it, are you going to try them or should i :)?

prusnak:
> lets go for it, are you going to try them or should i :)?

I already tried some of them.

* `C` is interesting, some things are trivial (like unnecessary `map` usage), some are not that easy (function `pay_invoice` is too complex) which require lot of changes
* `N` is just pain in the ass :D
* `UP` does not respect minimum Python version, so it wants to replace type hints `Union[X, Y]` with `X | Y` which is cool but only supported on Python 3.10 and we still support 3.9
* `B` is mostly about exception handling, might be nice to address the issues
* `A` same as `N`, just pain in the ass
* `RUF` seems nice, good improvements

To recap: I think it is worth fixing issues for `C`, `B` and `RUF`, others not so much.

PR #2382, opened 2024-04-01
feat: adhere to ruff's `RUF` rules

dni:
https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf

#2308

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -314,7 +314,7 @@ async def check_installed_extension_files(ext: InstallableExtension) -> bool:
 
     zip_files = glob.glob(os.path.join(settings.lnbits_data_folder, "zips", "*.zip"))
 
-    if f"./{str(ext.zip_path)}" not in zip_files:
+    if f"./{ext.zip_path!s}" not in zip_files:
         await ext.download_archive()
     ext.extract_archive()
 
@@ -417,7 +417,7 @@ def register_all_ext_routes(app: FastAPI):
         try:
             register_ext_routes(app, ext)
         except Exception as e:
-            logger.error(f"Could not load extension `{ext.code}`: {str(e)}")
+            logger.error(f"Could not load extension `{ext.code}`: {e!s}")
 
 
 def initialize_server_logger():
@@ -488,7 +488,7 @@ def register_exception_handlers(app: FastAPI):
     async def exception_handler(request: Request, exc: Exception):
         etype, _, tb = sys.exc_info()
         traceback.print_exception(etype, exc, tb)
-        logger.error(f"Exception: {str(exc)}")
+        logger.error(f"Exception: {exc!s}")
         # Only the browser sends "text/html" request
         # not fail proof, but everything else get's a JSON response
         if (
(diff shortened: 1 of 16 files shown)


lnbits/app.py
```diff
@@ -262,7 +262,7 @@ async def check_installed_extension_files(ext: InstallableExtension) -> bool:
 
     zip_files = glob.glob(os.path.join(settings.lnbits_data_folder, "zips", "*.zip"))
 
-    if f"./{str(ext.zip_path)}" not in zip_files:
+    if f"./{ext.zip_path!s}" not in zip_files:
```
motorina0, line 265: was ist das?

lnbits/core/models.py
```diff
@@ -177,7 +177,7 @@ class Payment(FromRowModel):
     preimage: str
     payment_hash: str
     expiry: Optional[float]
-    extra: Dict = {}
+    extra: Dict
```
motorina0, line 180: why not assign it a default value?

lnbits/core/models.py
```diff
@@ -318,7 +318,7 @@ class PaymentFilters(FilterModel):
     preimage: str
     payment_hash: str
     expiry: Optional[datetime.datetime]
-    extra: Dict = {}
+    extra: ClassVar[Dict] = {}
```
motorina0, line 321: what does `ClassVar` mean in this context?

lnbits/core/sso/keycloak.py
```diff
@@ -11,7 +11,7 @@ class KeycloakSSO(SSOBase):
     """Class providing login via Keycloak OAuth"""
 
     provider = "keycloak"
-    scope = ["openid", "email", "profile"]
+    scope: list[str] = ["openid", "email", "profile"]  # noqa: RUF012
```
motorina0, line 14: why no use `ClassaVar` here?

lnbits/db.py
```diff
@@ -370,7 +370,7 @@ def from_row(cls, row: Row):
 
 
 class FilterModel(BaseModel):
-    __search_fields__: List[str] = []
+    __search_fields__: ClassVar[List[str]] = []
```
motorina0, line 373: Aren't `__sort_fields__` below in the same situation?

lnbits/wallets/fake.py
```diff
@@ -31,8 +31,8 @@
 
 class FakeWallet(Wallet):
     queue: asyncio.Queue = asyncio.Queue(0)
-    payment_secrets: Dict[str, str] = dict()
-    paid_invoices: Set[str] = set()
+    payment_secrets: ClassVar[Dict[str, str]] = {}
+    paid_invoices: ClassVar[Set[str]] = set()
```
motorina0, line 35: how was decided that only these two vars are `ClassVar` while the other not?

lnbits/app.py
dni: !s 	Interpolates the string representation from the .__str__() method
:D

lnbits/wallets/fake.py
dni: ClassVar marks which are mutable object
https://docs.astral.sh/ruff/rules/mutable-class-default/

PR #2389, opened 2024-04-03
chore: run `poetry lock`

dni:
update lock file


prusnak:
Let's update deps in one PR next time

Updating them one by one and then resolving conflicts in poetry.lock is not fun

PR #2403, opened 2024-04-11
fix: properly fix check_funding_source

dni:
saas did not like that retrying loop at all :(

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -177,43 +177,26 @@ def create_app() -> FastAPI:
 
 
 async def check_funding_source() -> None:
-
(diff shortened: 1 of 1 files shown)


talvasconcelos (approved):

motorina0:
NACK

motorina0 (changes requested):
`max_retries` should be an env setting

motorina0:
I would rather prefer this more flexible approach: https://github.com/lnbits/lnbits/pull/2404

prusnak:
I like https://github.com/lnbits/lnbits/pull/2404 more than this approach

PR #2417, opened 2024-04-16
feat: `catch_everything_and_restart` print name of the task

dni:
and remove type from `Coroutine` from the create_tasks


motorina0:
What is wrong with the type?

motorina0 (approved):

talvasconcelos (approved):

PR #2441, opened 2024-04-17
AI prompt manners

arcbtc:
Just because they are robots that will prob one day kill us, we should still remember our manners (then maybe they will spare us).


dni (approved):
^^

dni (changes requested):
conflicting, pavol changed to prompt recently

Issue #2399, opened 2024-04-11
[feat] add `app_running: bool = True` transient setting


motorina0:
There should be no `while True` loops in the app.
Instead `while settings.app_running:` should be used.

On the shutdown signal do `settings.app_running = False`.

PR #2450, opened 2024-04-19
feat: add `settings.lnbits_running `

motorina0:
https://github.com/lnbits/lnbits/issues/2399

There should be no `while True` loops in the app.
Instead `while settings.lnbits_runing:` should be used.

On the shutdown signal do `settings.lnbits_runing = False`.

This change also helps while testing. Previous infinite loops can no be stoped.

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -104,7 +104,7 @@ async def startup(app: FastAPI):
 
 
 async def shutdown():
-    settings.lnbits_runing = False
+    settings.lnbits_running = False
     # shutdown event
     cancel_all_tasks()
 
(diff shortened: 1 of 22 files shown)


lnbits/app.py
```diff
@@ -181,7 +182,7 @@ async def check_funding_source() -> None:
     max_retries = settings.funding_source_max_retries
     retry_counter = 0
 
-    while True:
+    while settings.lnbits_runing:
```
dni, line 185:
```suggestion
    while settings.lnbits_running:
```

typo

motorina0: fixed, thanks!

motorina0:
@dni does it make sense to set this flag also when `server_restart.set()` is called (`/api/v1/restart` endpoint)?

talvasconcelos (approved):

dni:
@motorina0 i would set it in `server.py`

dni (approved):

PR #2456, opened 2024-04-22
Feat: more extension details

arcbtc:
For new users its disorientating to install extensions and not know what they do. This pr makes use of an extended `config.json` in extensions https://github.com/lnbits/lnurldevice/blob/main/config.json
The pr will build a `More` dialogue that can display images, videos, extended description. When we have reviews this will also be how they are accessed.

You can test with https://raw.githubusercontent.com/lnbits/tpos/main/manifest.json

This pr will also enable a seperate splash page to be built like https://en-gb.wordpress.org/plugins/


lnbits/core/views/extension_api.py
```diff
@@ -189,6 +190,24 @@ async def get_extension_releases(ext_id: str):
         ) from exc
 
 
+@extension_router.get("/{ext_id}/details")
```
dni, line 193: this can be done on the frontend only

motorina0:
Is there a manifest.json that I can use for testing?

lnbits/core/views/generic.py
```diff
@@ -136,16 +135,22 @@ async def extensions_install(
...
         extensions = [
             {
                 "id": ext.id,
                 "name": ext.name,
                 "icon": ext.icon,
                 "shortDescription": ext.short_description,
-                "extendedDescription": ext.description_md or None,
-                "immages": ext.images or None,
-                "contributors": ext.contributors or None,
+                "immages": ext.images if ext.images else None,
```
dni:
```suggestion
                "images": ext.images if ext.images else None,
```

dni: isnt ext.images `None` this if makes no sense, imho?

lnbits/core/templates/core/extensions.html
```diff
@@ -1089,6 +1085,7 @@ <h6 class="q-my-sm" v-text="extensionConfig.short_description"></h6>
         ...e,
         inProgress: false
       }))
+      console.log(this.extensions)
```
dni: log

PR #2467, opened 2024-04-25
Fix mishandling of send_payment response

dangeross:
The application should receive a `SendPaymentResponse` from the SDK when calling `send_payment()`. Within the response it contains the payment.


dni:
thanks!

PR #2468, opened 2024-04-25
refactor: fees should not be negative

dni:
WIP

this is really misleading and also the reason we have `abs()` all over the place, fees should be a positive int on default


lnbits/core/services.py
```diff
@@ -344,13 +344,11 @@ class PaymentKwargs(TypedDict):
             # payment.ok can be True (paid) or None (pending)!
             logger.debug(f"updating payment {temp_id}")
             async with db.connect() as conn:
+                fee = payment.fee_msat or 0 + service_fee_msat
```
prusnak, line 347: Let's make the logic more obvious
```suggestion
                fee = (payment.fee_msat or 0) + service_fee_msat
```

prusnak (changes requested):

PR #2469, opened 2024-04-26
fix: do not delete payments

dni:
as a safety measure we do not delete failed payments

### DIFF
--- a/lnbits/core/models.py
+++ b/lnbits/core/models.py
@@ -272,38 +272,13 @@ class Payment(FromRowModel):
             status = await funding_source.get_invoice_status(self.checking_id)
 
         logger.debug(f"Status: {status}")
-
-        if self.is_in and status.pending and self.is_expired and self.expiry:
-            expiration_date = datetime.datetime.fromtimestamp(self.expiry)
-            logger.debug(
-                f"Deleting expired incoming pending payment {self.checking_id}: "
-                f"expired {expiration_date}"
-            )
-            await self.delete(conn)
-        # wait at least 15 minutes before deleting failed outgoing payments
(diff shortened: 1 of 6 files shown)


motorina0:
Great PR. This was a performance improvement with serious negative implications.
If an outgoing payment is deleted by mistake then the balance of the LNbits server would be higher than the one on the Funding Source.

The deletion of old not successful payments should happen outside the main business logic of the application.
There is a command for deleted old (more than 6 months) empty wallets. 
There should be a command to delete old non-successful payments. @dni can you create one (I believe you also handled the wallets command).

motorina0 (approved):

talvasconcelos (approved):

arcbtc (approved):

arcbtc (changes requested):
This is a great idea. I was worried about reintroducing the infinite pending transaction issue, but thats something to tackle after we merge this important update. 
Sorry for not appreciating the importance of this pr originally 🙏

lnbits/core/models.py
```diff
@@ -253,58 +253,25 @@ async def set_pending(self, pending: bool) -> None:
...
+
         funding_source = get_funding_source()
         if self.is_out:
             status = await funding_source.get_payment_status(self.checking_id)
         else:
             status = await funding_source.get_invoice_status(self.checking_id)
 
         logger.debug(f"Status: {status}")
-
-        if self.is_in and status.pending and self.is_expired and self.expiry:
```
motorina0, line 276: good clean-up?

dni: yes is see no reason to keep this in the model class

PR #2481, opened 2024-05-08
fix: pay invoice status

motorina0:
### Summary

This PR addresses an issue for **failed** `outgoing payments`.

If an outgoing payment (invoice paid by the user) is marked as failed, then it will be deleted after a certain time interval.
The (current) reason for deleting the failed outgoing payment is to unlock the user funds. Because while the payment is still present the amount is considered locked.

The issue is that sometimes the `outgoing payment` can be marked as failed for technical reason, BUT the invoice is not actually canceled.

The solution in this PR is to:
 - never mark a payment as `failed` unless the funding source explicitly returns a status that says the payment is failed
 - any technical errors will keep the invoice as `pending`
 - make sure `get_payment_status` for all funding sources returns `failed` under _some_ condition
   - `alby` and `poenixd` have an issue here

### DIFF
--- a/.github/workflows/regtest.yml
+++ b/.github/workflows/regtest.yml
@@ -80,3 +80,8 @@ jobs:
           file: ./coverage.xml
           token: ${{ secrets.CODECOV_TOKEN }}
           verbose: true
+
+      - name: docker lnbits logs
+        if: ${{ inputs.backend-wallet-class == 'LNbitsWallet' }}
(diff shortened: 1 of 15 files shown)


motorina0:
Note: regtest fails for reasons outside this PR. Should be fixed soon.

dni (approved):

dni (approved):

PR #2490, opened 2024-05-11
fix: show proper total balances fix cleanups

dni:
payments are not deleted if we delete a wallet, so to get a accurate total representation of the lnbits balance we need to create the balances view based on the wallets table, not payments, else deleted balances will still show up.

2nd, `delete_unused_wallets` and `delete_accounts was` never working because they never got an `updated_at` time, so i just check if its null else i check to timedelta on created_at


motorina0:
> 2nd, `delete_unused_wallets` and `delete_accounts was` never working
can we please add some unit tests to this

talvasconcelos (approved):

arcbtc (approved):
Works great for me

PR #2504, opened 2024-05-14
fix: topup wallet was showing NaN

dni:


lnbits/static/js/users.js
```diff
@@ -351,12 +351,14 @@ new Vue({
...
     },
     topupCallback(res) {
-      this.wallets.forEach(wallet => {
-        if (res.wallet_id === wallet.id) {
-          wallet.balance_msat += res.value * 1000
-        }
-      })
-      this.fetchUsers()
+      if (res.success) {
+        this.wallets.forEach(wallet => {
```
motorina0, line 355: why is there a foreach here?

lnbits/static/js/users.js
```diff
@@ -351,12 +351,14 @@ new Vue({
       console.log('export users')
     },
     topupCallback(res) {
-      this.wallets.forEach(wallet => {
-        if (res.wallet_id === wallet.id) {
-          wallet.balance_msat += res.value * 1000
-        }
-      })
-      this.fetchUsers()
+      if (res.success) {
```
motorina0, line 354: nitpick:
```suggestion
 if (!res.success) {
     return
 }
```

lnbits/static/js/users.js
dni: i want to update the balance inside the list and this.wallets is the reactive property so i iterate through them to change the balance there

arcbtc (approved):

Issue #2054, opened 2023-10-21
[BUG] Invoice created show wrong expiry time


d6n13l0l1v3r:
**Describe the bug**
When I create a invoice the expiry Time I sent to  a month but on the setting on the invoice management is set to 3600 seconds

**To Reproduce**

1. Go to 'wallet'
2. Click on ' create invoice'
3. click to check the information of the invoice 
4. See expiry 

**Expected behavior**
the expiry time must be according with the setting define on the server in the .env file or in admin page

**Screenshots**

![image](https://github.com/lnbits/lnbits/assets/25048143/4b4108e3-177a-443c-befc-b25627f6856a)

![image](https://github.com/lnbits/lnbits/assets/25048143/2f06af15-e340-4090-96f5-93fd9598bdfd)

**Desktop (please complete the following information):**

- LNbits version: 0.11.1
- Database sqlite 
- Implementation CoreLighthing v23.08.1

d6n13l0l1v3r:
adding more information, today I got in start9 the possibility to upgrade my lnbits to version 0.11.1 when I did it, the bug happen, for that I did a downgrade to version 0.10.10 and now the expiry date for an invoice is calculated property 

it's very clear that have a bug on version 0.11.1 

after upgrade 

![image](https://github.com/lnbits/lnbits/assets/25048143/821b5462-7ea2-4a88-a952-617f79a8316f)


after downgrade 

![image](https://github.com/lnbits/lnbits/assets/25048143/3bee55b0-85b2-4f2c-85fc-8f8a91d0d219)

d6n13l0l1v3r:
I thing maybe the  issue could be related to 

services.py 

`
def get_bolt11_expiry(invoice: Bolt11) -> datetime.datetime:
    if invoice.expiry:
        return datetime.datetime.fromtimestamp(invoice.date + invoice.expiry)
    else:
        # assume maximum bolt11 expiry of 31 days to be on the safe side
        return datetime.datetime.now() + datetime.timedelta(days=31)
`

that is include in version 0.11.1 and not is on version 0.10.10

dethos:
I suspect this issue might be related to https://github.com/lnbits/bolt11/issues/40 .

Since on my tests, `invoice.expiry` is always `None` here:

```python
def get_bolt11_expiry(invoice: Bolt11) -> datetime.datetime:
    if invoice.expiry:  # <------ HERE --------
        return datetime.datetime.fromtimestamp(invoice.date + invoice.expiry)
[...]

PR #2506, opened 2024-05-15
Fix: Wrong expiration date in invoice details

dethos:
Addresses: https://github.com/lnbits/lnbits/issues/2054

The rationale for this change is explained in the issue's discussion: https://github.com/lnbits/lnbits/issues/2054#issuecomment-2112247822

**Summary:** The Bolt 11 spec says that an invoice, without the `x` tag field specified, should expire by default in 1 hour (3600 seconds). The `bolt11` package used by lnbits does not implement this, setting the `expiry` to `None` when the tag field is absent. The current implemented fallback sets the expiration date to 1 month after the creation date, which is incompatible with the spec and creates the reported differences between lnbits and other tools that decode the same invoice.

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -8,7 +8,6 @@ from typing import Dict, List, Optional, Tuple, TypedDict
 from urllib.parse import parse_qs, urlparse
 
 import httpx
-from bolt11 import Bolt11
 from bolt11 import decode as bolt11_decode
 from cryptography.hazmat.primitives import serialization
 from fastapi import Depends, WebSocket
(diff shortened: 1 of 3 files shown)


lnbits/core/services.py
```diff
@@ -801,8 +801,6 @@ async def get_balance_delta() -> BalanceDelta:
 
 
 def get_bolt11_expiry(invoice: Bolt11) -> datetime.datetime:
```
dni, line 803: can you remove this function altogether and use this directly instead? https://github.com/lnbits/bolt11/blob/main/bolt11/types.py#L112
```invoice.expire_date`

you should also update bolt11 library to 2.0.6 i just fixed an issue there

dethos: Done

dni (approved):

dni:
closes #2054

arcbtc (approved):

PR #2513, opened 2024-05-17
Update phoenixd.py

Hodladi:
Fix for the description_hash with phoneixd server

### DIFF
--- a/lnbits/wallets/phoenixd.py
+++ b/lnbits/wallets/phoenixd.py
@@ -68,7 +68,6 @@ class PhoenixdWallet(Wallet):
                 return StatusResponse(f"Server error: '{error_message}'", 0)
 
             if len(data["channels"]) == 0:
-                # todo: add custom unit-test for this
                 return StatusResponse(None, 0)
 
             balance_msat = int(data["channels"][0]["balanceSat"]) * 1000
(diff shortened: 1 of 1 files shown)


prusnak:
Please keep the todo comments - you have not addressed them, so you should not remove them

Bashy:
Was it tested? I might have time to test this weekend, and maybe provide a PR if this ones is not ok, but I think this is **not** going to work since
```py
            data: Dict = {
                "amountSat": f"{msats_amount}",
                "description": memo,
                "externalId": "",
            }
```
`description` here - as I understand - should not has `memo`, but rather the expected `description_hash`, so we should have something like

```py
            data: Dict = {
                "amountSat": f"{msats_amount}",
                "description": description_hash,
                "externalId": "",
            }
```

Here https://github.com/ACINQ/phoenixd/pull/50/files#diff-154655ec623b797aed456fbb139aba5ea257320178b9574f12f676560f3c3b4c, you can see line -154/+162 that `description` expects the `description_hash`

`memo` is not relevant for phoenixd

Bashy:
To clarify my previous comment, it might work but not the expected way where phoenixd actually processes the `description_hash`

Hodladi:
It works, I'm running my LNBits instance with that change and it works perfectly.

I'm no python dev so this is probably possible to do better but as I said, it do work.

Funman2:
this branch is able to even receive nostr zaps, whereas in  #2514 they don't go through.

prusnak:
Closed in favor of https://github.com/lnbits/lnbits/pull/2514

PR #2514, opened 2024-05-17
fix: phoenixd wallet description field supports lnurlp

vicariousdrama:
Similar resolution as #2513, but leaves in comments and applies description.

Tested with lnurlp extension locally with phoenixd 0.1.5-6845a31 in testnet, with assignment of unhashed_description or memo.

### DIFF
--- a/lnbits/wallets/phoenixd.py
+++ b/lnbits/wallets/phoenixd.py
@@ -103,10 +103,11 @@ class PhoenixdWallet(Wallet):
                 desc = memo
                 if desc is None and unhashed_description:
                     desc = unhashed_description.decode()
-                desc = "" if desc is None else desc
+                desc = desc or ""
                 if len(desc) > 128:
-                    desc = hashlib.sha256(desc.encode("utf-8")).hexdigest()
-                data["description"] = desc
+                    data["descriptionHash"] = hashlib.sha256(desc.encode("utf-8")).hexdigest()
+                else:
+                    data["description"] = desc
 
             r = await self.client.post(
                 "/createinvoice",
--- a/lnbits/wallets/phoenixd.py
+++ b/lnbits/wallets/phoenixd.py
@@ -1,5 +1,6 @@
 import asyncio
 import base64
+import hashlib
 import json
 import urllib.parse
 from typing import AsyncGenerator, Dict, Optional
@@ -17,7 +18,6 @@ from .base import (
     PaymentStatus,
     PaymentSuccessStatus,
     StatusResponse,
-    UnsupportedError,
     Wallet,
 )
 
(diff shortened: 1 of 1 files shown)


Funman2:
I did a quick test, with this branch nostr zaps dont work, but with #2513 they work.

vicariousdrama:
Yeah, its unclear to me why the pattern for the lnurlp extension is sending the metadata as description if that needs to be dropped on the floor and use the memo instead.  There's likely breakage for lnurlp and nostr zaps across other funding sources for the same reason (opennode,zbd,...)

vicariousdrama:
Thanks @Funman2 for the check on nostr. I hadn't tried that but just put in the tweak needed to make it work by encoding the description field when it needs hashed.

Funman2:
It works now with nostr.

vicariousdrama:
Rebased to dev, and squashed commits

vicariousdrama:
Resolved what I could of linting issues, but there's still some disparity between len and Sized

prusnak:
> disparity between len and Sized

Where exactly? I don't see any CI failures now

prusnak (approved):

Hodladi:
> Thanks @Funman2 for the check on nostr. I hadn't tried that but just put in the tweak needed to make it work by encoding the description field when it needs hashed.

To me the nostr native zaps still not working... 
I still get the long descripition with relays and alot of stuff and it fails the payment.

lnbits/wallets/phoenixd.py
```diff
@@ -87,17 +87,27 @@ async def create_invoice(
...
 
+            # Either 'description' (string) or 'descriptionHash' must be supplied
+            # PhoenixD description limited to 128 characters
+            if description_hash:
+                data["descriptionHash"] = description_hash.hex()
+            else:
+                desc = memo
+                if desc is None and unhashed_description:
+                    desc = unhashed_description.decode()
+                desc = "" if desc is None else desc
```
dni, line 106:
```suggestion
                desc = desc or ""

File: lnbits/wallets/phoenixd.py [hunk 2]
```diff
@@ -87,17 +87,27 @@ async def create_invoice(
...
+                if len(desc) > 128:
+                    desc = hashlib.sha256(desc.encode("utf-8")).hexdigest()
+                data["description"] = desc
```
dni, line 109: shouldn't it just set the descriptionHash here, not description? if its hasing?

vicariousdrama: Modified to set descriptionHash if its getting hashed here

lnbits/wallets/phoenixd.py
vicariousdrama: Included

prusnak:
@dni this one is ready, plz review and merge

motorina0:
tACK

motorina0 (approved):

PR #2522, opened 2024-05-21
[feat] Check payment tag

motorina0:
### Summary
Do not allow payments from extensions that are disabled by the user.

**Steps**:
 - user enables `withdraw` extension
 - user create  `withdraw link` and shares it
 - user disables  `withdraw` extension

**Actual Result**:
 - `withdraw link` still working

**Expected Result**:
 - `withdraw link` does not work until the `withdraw` extension is enabled again

**Note**:
 - this functionality relies on the fact that extensions use the `"tag"` field in the "'extra'" options
   -  this is not enforced in the code in any way at the moment

 - the function `_check_wallet_balance` is a simple refactor (no business logic changes) it was required by the code check tools (parent function was getting too large)

image

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -17,7 +17,11 @@ from py_vapid.utils import b64urlencode
 
 from lnbits.core.db import db
 from lnbits.db import Connection
-from lnbits.decorators import WalletTypeInfo, require_admin_key
+from lnbits.decorators import (
+    WalletTypeInfo,
+    check_user_extension_access,
+    require_admin_key,
+)
 from lnbits.helpers import url_for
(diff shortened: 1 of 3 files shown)


lnbits/core/services.py
```diff
@@ -402,6 +401,22 @@ class PaymentKwargs(TypedDict):
     return invoice.payment_hash
 
 
+def _check_wallet_balance(
```
motorina0, line 404: this function was extracted because the parent function was getting too large and flake8 complained

dni (approved):

talvasconcelos (approved):

Issue #2525, opened 2024-05-22
[BUG] Migration to 0.12.6 crashes


psic4t:
I use the official docker image. After updating from 0.12.5 to 0.12.6 lnbits crashes on startup:

```
lnbits    | 2024-05-22 07:23:24.57 | INFO | Started server process [12]
lnbits    | 2024-05-22 07:23:24.57 | INFO | Waiting for application startup.
lnbits    | 2024-05-22 07:23:24.58 | DEBUG | lnbits.core.helpers:run_migration:40 | running migration core.18
lnbits    | running migration core.18
lnbits    | 2024-05-22 07:23:24.58 | DEBUG | lnbits.core.helpers:run_migration:40 | running migration core.19
lnbits    | running migration core.19
lnbits    | 2024-05-22 07:23:24.59 | INFO | lnbits.commands:migrate_databases:164 | ✔️ All migrations done.
lnbits    | 2024-05-22 07:23:24.59 | ERROR | Traceback (most recent call last):
lnbits    |   File "/root/.cache/pypoetry/virtualenvs/lnbits-9TtSrW0h-py3.10/lib/python3.10/site-packages/starlette/routing.py", line 734, in lifespan
lnbits    |     async with self.lifespan_context(app) as maybe_state:
lnbits    |   File "/usr/local/lib/python3.10/contextlib.py", line 199, in __aenter__
lnbits    |     return await anext(self.gen)
lnbits    |   File "/app/lnbits/app.py", line 117, in lifespan
lnbits    |     await startup(app)
lnbits    |   File "/app/lnbits/app.py", line 75, in startup
lnbits    |     await check_admin_settings()
lnbits    |   File "/app/lnbits/core/services.py", line 671, in check_admin_settings
lnbits    |     settings_db = await get_super_settings()
lnbits    |   File "/app/lnbits/core/crud.py", line 1121, in get_super_settings
lnbits    |     return SuperSettings(**{"super_user": row["super_user"], **editable_settings})
lnbits    |   File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
lnbits    | pydantic.error_wrappers.ValidationError: 1 validation error for SuperSettings
lnbits    | lnbits_site_description
lnbits    |   none is not an allowed value (type=type_error.none.not_allowed)
lnbits    |
lnbits    | 2024-05-22 07:23:24.59 | ERROR | Application startup failed. Exiting.
```

It seems that everything in the settings table got wiped.

[...]

PR #2526, opened 2024-05-22
hotfix: make `lnbits_site_description` optional again

motorina0:
Reported here: https://github.com/lnbits/lnbits/issues/2525

Steps:
 - make sure `lnbits_site_description` is `null` in the DB
 - you can do this by adding a new temp line to `update_admin_settings`:

```py
async def update_admin_settings(data: EditableSettings) -> None:
    row = await db.fetchone("SELECT editable_settings FROM settings")
    editable_settings = json.loads(row["editable_settings"]) if row else {}
    editable_settings.update(data.dict(exclude_unset=True))
    editable_settings["lnbits_site_description"] = None # <-- this is new
    await db.execute(
        "UPDATE settings SET editable_settings = ?", (json.dumps(editable_settings),)
    )
``` 
 - save the settings as admin (this will simulate the `lnbits_site_description = null`
 - restart LNbits

**Actual Result**:
 - server will fail to start with
```log
line 75, in startup
    await check_admin_settings()
  File "/Users/moto/Documents/GitHub/motorina0/lnbits/lnbits/core/services.py", line 686, in check_admin_settings
    settings_db = await get_super_settings()
  File "/Users/moto/Documents/GitHub/motorina0/lnbits/lnbits/core/crud.py", line 1128, in get_super_settings
    return SuperSettings(**{"super_user": row["super_user"], **editable_settings})
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for SuperSettings
lnbits_site_description
  none is not an allowed value (type=type_error.none.not_allowed)
```

**Expected Result**:
 - server starts OK even if `lnbits_site_description ` is `null`

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -261,10 +261,10 @@ async def build_all_installed_extensions_list(
     MUST be installed by default (see LNBITS_EXTENSIONS_DEFAULT_INSTALL).
     """
     installed_extensions = await get_installed_extensions()
+    settings.lnbits_all_extensions_ids = {e.id for e in installed_extensions}
 
-    installed_extensions_ids = [e.id for e in installed_extensions]
     for ext_id in settings.lnbits_extensions_default_install:
-        if ext_id in installed_extensions_ids:
+        if ext_id in settings.lnbits_all_extensions_ids:
             continue
 
         ext_releases = await InstallableExtension.get_extension_releases(ext_id)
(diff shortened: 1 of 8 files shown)


dni:
cherry picked onto main

PR #2537, opened 2024-05-25
refactor: add status column to apipayments

### LINKED ISSUE #2480: [Feature request] add `paid` column to apipayments table
very useful to keep failed payments and delete them later.

pending = False and paid = False means its failed

dni:
keep track of the payment status with an enum and persist it as string to db. `pending`, `success`, `failed`.

- database migration
- remove deleting of payments, failed payments stay

closes #2469
closes #2480

LNDhub requires an update: https://github.com/lnbits/lndhub/pull/32


lnbits/core/crud.py
```diff
@@ -722,7 +723,7 @@ async def get_latest_payments_by_extension(ext_name: str, ext_id: str, limit: in
     rows = await db.fetchall(
         f"""
         SELECT * FROM apipayments
-        WHERE pending = false
+        WHERE status = 'success'
```
motorina0, line 726: Why not use:
```suggestion
        WHERE status = '{PaymentState.SUCCESS}'
```

lnbits/core/models.py
```diff
@@ -213,6 +223,18 @@ class Payment(FromRowModel):
...
+        return self.status == PaymentState.SUCCESS.value
+
+    @property
+    def failed(self) -> bool:
+        return self.status == PaymentState.FAILED.value
+
+    @property
+    def pending(self) -> bool:
+        return self.status == PaymentState.PENDING.value
+
```
motorina0, line 237:
```suggestion

    def __str__(self):
        return str(self.value)
```

dni: https://stackoverflow.com/questions/58608361/string-based-enum-in-python#58608362
screenshot-1716795178

i think this is not needed ffor string based enums

dni: i changed it anyways :)

lnbits/core/views/public_api.py
```diff
@@ -20,7 +20,7 @@ async def api_public_payment_longpolling(payment_hash):
         raise HTTPException(
             status_code=HTTPStatus.NOT_FOUND, detail="Payment does not exist."
         )
-    elif not payment.pending:
+    if payment.success:
         return {"status": "paid"}
```
motorina0, line 24: "paid" is not a valid value for status with the changes in this PR.
```suggestion
      return {"status": payment.status}
```

The LNbits funding source should also be correlated.

lnbits/core/crud.py
```diff
@@ -766,9 +767,11 @@ async def get_payments_paginated(
     if complete and pending:
         pass
     elif complete:
-        clause.append("((amount > 0 AND pending = false) OR amount < 0)")
+        clause.append(
+            f"((amount > 0 AND status NOT '{PaymentState.PENDING}') OR amount < 0)"
```
motorina0, line 771: Should be a positive inclusion condition rather than an exclusion one.
What happens with the `FAILED`, should they be included here?
What if we introduce a different `PENDING`-like status (eg `IN_FLIGHT`)?

```suggestion
            f"((amount > 0 AND status IN (<all complete statues here>)) OR amount < 0)"
```

lnbits/core/migrations.py
```diff
@@ -513,3 +513,30 @@ async def m019_balances_view_based_on_wallets(db):
...
     )
+
+
+async def m020_add_success_failed_to_apipayments(db):
+    """
+    Adds success and failed columns to apipayments.
+    """
+    await db.execute("ALTER TABLE apipayments ADD COLUMN status TEXT DEFAULT 'pending'")
+    #  set all not pending to success true, failed payments were deleted until now
+    await db.execute("UPDATE apipayments SET status = 'success' WHERE NOT pending")
```
motorina0, line 524: `state` vs `status`
the class is `PaymentState` (due to name conflict), but the column and variables are `status`

lnbits/core/migrations.py
```diff
@@ -513,3 +513,30 @@ async def m019_balances_view_based_on_wallets(db):
...
     """
     )
+
+
+async def m020_add_success_failed_to_apipayments(db):
+    """
+    Adds success and failed columns to apipayments.
+    """
+    await db.execute("ALTER TABLE apipayments ADD COLUMN status TEXT DEFAULT 'pending'")
+    #  set all not pending to success true, failed payments were deleted until now
```
motorina0, line 523: how about the 15 minutes delay?

lnbits/core/migrations.py
```diff
@@ -520,3 +520,30 @@ async def m020_add_column_column_to_user_extensions(db):
...
+    await db.execute(
+        """
+        CREATE VIEW balances AS
+        SELECT apipayments.wallet,
+               SUM(apipayments.amount - ABS(apipayments.fee)) AS balance
+        FROM wallets
+        LEFT JOIN apipayments ON apipayments.wallet = wallets.id
+        WHERE (wallets.deleted = false OR wallets.deleted is NULL)
+        AND (
+            (apipayments.status = 'success' AND apipayments.amount > 0)
```
motorina0, line 543: why not use the `PaymentState` enum?

lnbits/core/models.py
```diff
@@ -197,9 +201,18 @@ class LoginUsernamePassword(BaseModel):
     password: str
 
 
+class PaymentState(str, Enum):
```
motorina0, line 204: just an idea
```suggestion
class ApiPaymentStatus(str, Enum):
```

lnbits/static/js/components/payment-list.js
```diff
@@ -375,9 +382,9 @@ Vue.component('payment-list', {
                           :payment="props.row"
                         ></lnbits-payment-details>
                       </div>
-                      <div v-else-if="props.row.isOut && props.row.pending">
-                        <q-icon name="settings_ethernet" color="grey"></q-icon>
-                        <span v-text="$t('outgoing_payment_pending')"></span>
+                      <div v-else-if="props.row.isFailed">
```
motorina0, line 385: how can I simulate this so I can test the UI?

lnbits/core/models.py
```diff
@@ -262,80 +287,20 @@ def is_expired(self) -> bool:
     def is_uncheckable(self) -> bool:
         return self.checking_id.startswith("internal_")
 
-    async def update_status(
```
motorina0, line 265: ahh, what a useless public method

lnbits/core/models.py
```diff
@@ -262,80 +287,20 @@ def is_expired(self) -> bool:
...
-
-        self.pending = pending
-
-        await update_payment_status(self.checking_id, pending)
-
-    async def check_status(
-        self,
-        conn: Optional[Connection] = None,
-    ) -> PaymentStatus:
+    async def check_status(self) -> PaymentStatus:
```
motorina0, line 290: wow, this method is so simple now!

lnbits/core/services.py
```diff
@@ -371,10 +375,12 @@ class PaymentKwargs(TypedDict):
                 logger.debug(f"payment successful {payment.checking_id}")
         elif payment.checking_id is None and payment.ok is False:
             # payment failed
-            logger.warning("backend sent payment failure")
             async with db.connect() as conn:
-                logger.debug(f"deleting temporary payment {temp_id}")
-                await delete_wallet_payment(temp_id, wallet_id, conn=conn)
+                await update_payment_status(
```
motorina0, line 379: a `TRACE` level logger might come in handy

lnbits/core/services.py
```diff
@@ -612,12 +618,11 @@ async def check_transaction_status(
     )
     if not payment:
         return PaymentPendingStatus()
-    if not payment.pending:
-        # note: before, we still checked the status of the payment again
+
+    if payment.status is PaymentState.SUCCESS.value:
```
motorina0, line 622: why use `is` for `str`?

lnbits/tasks.py
```diff
@@ -131,45 +132,38 @@ async def check_pending_payments():
...
-            f"Task: pending check finished for {len(pending_payments)} payments"
-            f" (took {time.time() - start_time:0.3f} s)"
-        )
-        # we delete expired invoices once upon the first pending check
-        if incoming:
-            logger.debug("Task: deleting all expired invoices")
-            start_time = time.time()
-            await delete_expired_invoices()
+        count = len(pending_payments)
+        if count > 0:
```
motorina0, line 145: nitpick:
```suggestion
        if count == 0:
             continue

File: lnbits/tasks.py [hunk 15]
```diff
@@ -131,45 +132,38 @@ async def check_pending_payments():
     the backend and also to delete expired invoices. Incoming payments will be
     checked only once, outgoing pending payments will be checked regularly.
     """
-    outgoing = True
-    incoming = True
 
     while settings.lnbits_running:
```
motorina0, line 136: This loop is terminated if there is an SQL error (or any kind of exception raised).
```suggestion
    while settings.lnbits_running:
       try:
            _check_pending_payments()
       except Exception as exc:
           logger.warning(exc)
           
       await asyncio.sleep(60 * 30)  # every 30 minutes
```

lnbits/static/js/components/payment-list.js
motorina0, line 385: The color of the icons is different for failed payments.

image

Maybe a [different icon](https://fonts.google.com/icons?icon.set=Material+Icons&selected=Material+Icons+Outlined:priority_high:&icon.query=error&icon.size=24&icon.color=%23e8eaed) should be used for failed payments.

motorina0: `Failed` payments should not be retuned unless explicitly required so.
For example `BlueWallet` will list them as successful.


image

motorina0:
JMeter tests fixed here: https://github.com/lnbits/lnbits-extensions/pull/364/files
The issue was that the payments at `/api/v1/payments` no longer have the `pending` bool field.

lnbits/core/models.py
```diff
@@ -221,7 +246,7 @@ def from_row(cls, row: Row):
             bolt11=row["bolt11"] or "",
             preimage=row["preimage"] or "0" * 64,
             extra=json.loads(row["extra"] or "{}"),
-            pending=row["pending"],
+            status=row["status"],
```
motorina0, line 249: for backwards compatibility:

```suggestion
pending=row["status"] == PaymentState.PENDING.value,
```

see: https://github.com/lnbits/lnbits/pull/2537#issuecomment-2136880675

lnbits/core/migrations.py
dni: imports in migrations are weird, had circular import issue i think, lets keep it simple with str.

motorina0: ok

lnbits/core/models.py
dni: i like it, but in followup pr

lnbits/core/models.py
dni: yes... :D :dagger:

motorina0 (approved):
tACK

motorina0:
1. what happens with unpaid invoices after they expire
 2. what happens if I try to pay an invoice and it fails, then I try to pay a second time (will there be any DB conflicts for the same bolt11)?

lnbits/core/services.py
dni: added debug

lnbits/core/services.py
```diff
@@ -396,11 +391,51 @@ class PaymentKwargs(TypedDict):
...
+    conn: Optional[Connection],
+    **payment_kwargs,
+) -> Payment:
+    fee_reserve_total_msat = fee_reserve_total(amount_msat, internal=False)
+
+    old_payment = await get_standalone_payment(temp_id, conn=conn)
+    if old_payment:
+        if old_payment.pending:
+            raise PaymentError("Payment is still pending.", status="pending")
+        elif old_payment.success:
```
motorina0, line 411: nitpick: if `raise` or `return` used on the if branch the else is not required

```suggestion
        if old_payment.success:
```

lnbits/core/services.py
```diff
@@ -396,11 +391,51 @@ class PaymentKwargs(TypedDict):
...
+                raise PaymentError(
+                    "Failed payment was already paid on the fundingsource.",
+                    status="success",
+                )
+
+    logger.debug(f"creating temporary payment with id {temp_id}")
+    # create a temporary payment here so we can check if
+    # the balance is enough in the next step
+    try:
+        new_payment = old_payment or await create_payment(
```
motorina0, line 425: - this is the same as returning `old_payment` at the end of the `if old_payment:` branch
- in other words, `create_payment(...)` will never be called if `old_payment` exists

lnbits/core/services.py
```diff
@@ -396,11 +391,51 @@ class PaymentKwargs(TypedDict):
...
+    fee_reserve_total_msat = fee_reserve_total(amount_msat, internal=False)
+
+    old_payment = await get_standalone_payment(temp_id, conn=conn)
+    if old_payment:
+        if old_payment.pending:
+            raise PaymentError("Payment is still pending.", status="pending")
+        elif old_payment.success:
+            raise PaymentError("Payment already paid.", status="success")
+        elif old_payment.failed:
+            status = await old_payment.check_status()
```
motorina0, line 414: `check_status()` has this condition:

```py
        if self.failed:
            return PaymentFailedStatus()
```

will it ever make a call to the funding source in this scenario?

dni: where is that line of code? in my manual test it did work.

dni: yep, you were right, i removed this conditions, check should always go to fundingsource

lnbits/core/migrations.py
```diff
@@ -520,3 +520,30 @@ async def m020_add_column_column_to_user_extensions(db):
...
+            OR (apipayments.status IN ('success', 'pending') AND apipayments.amount < 0)
+        )
+        GROUP BY apipayments.wallet
+    """
+    )
+    await db.execute("ALTER TABLE apipayments DROP COLUMN pending")
```
motorina0, line 549:
```suggestion
    # todo: run with the next release
    # await db.execute("ALTER TABLE apipayments DROP COLUMN pending")
```

talvasconcelos (approved):
Seems to be working as expected

PR #2548, opened 2024-06-11
"Failed to connect to https://ws:" crash

evd0kim:
fixing bug in normalize_enpoints


lnbits/wallets/base.py
```diff
@@ -136,7 +136,7 @@ def paid_invoices_stream(self) -> AsyncGenerator[str, None]:
 
     def normalize_endpoint(self, endpoint: str, add_proto=True) -> str:
         endpoint = endpoint[:-1] if endpoint.endswith("/") else endpoint
-        if add_proto:
+        if add_proto and not endpoint.startswith("ws"):
```
prusnak, line 139: This is better. The original code does not behave correctly if `endpoint` like `wsexample.com` is provided.

```suggestion
        if add_proto:
            if endpoint.startswith("ws://") or endpoint.startswith("wss://"):
                return endpoint
```

prusnak (changes requested):

lnbits/wallets/base.py
evd0kim: Sure. Thanks.

dni (approved):

motorina0 (approved):

prusnak (approved):

PR #2558, opened 2024-06-21
feat: improve on extension page layout

dni:
simplified and improved responsiveness

screenshot-1718971749
screenshot-1718971302
screenshot-1718971295
screenshot-1718971288


motorina0 (approved):
LGTM

arcbtc:
I think its nice to keep the short description once activated, and the "new version" badge seems a bit floaty

blackcoffeexbt:
@dni I'd agree with @arcbtc on this. Keeping the short desc visible is important and the "new version" label should be positioned somewhere that it feels more "docked" to the card. Closer to a corner with a fixed distance would be good.

arcbtc:
I like @dni chnages, looks waaaayyy better on mobile, but I think we should keep the short description

arcbtc (approved):

PR #2561, opened 2024-06-25
feat: frontend gradient option

arcbtc:
Adds a gradient option to themes
Bit bloaty js, but performance not impacted


motorina0:
Love it!
Why still in draft?

lnbits/static/js/base.js
```diff
@@ -422,6 +422,18 @@ window.LNbits = {
       converter.setFlavor('github')
       converter.setOption('simpleLineBreaks', true)
       return converter.makeHtml(text)
+    },
+    hexToRgb: function (hex) {
```
dni, line 426: i dont think wrapping this into `LNbits.utils` makes sense, just use it directly

arcbtc:
> Love it! Why still in draft?

Because its pure spaghetti!

lnbits/static/js/base.js
arcbtc: Thought this way it would be useful for others

motorina0:
- I haven't managed to get the gradient effect:
image

 - Mobile looks strange when the left-side menu is visible:
image

- it is not clear in the UI when the gradient toggle is on or off (button state does not change)

dni (approved):

prusnak:
This PR should not modify package-lock.json because it does not touch package.json

Please revert that change

dni:
> This PR should not modify package-lock.json because it does not touch package.json
> 
> Please revert that change

done

PR #2566, opened 2024-06-28
[feat] add `authenticated_user_id` decorator

motorina0:
Light decorator used to check if there is an authenticated user or not.
Does not enforce authentication.
Required when querying for user specific resources.

### DIFF
--- a/lnbits/decorators.py
+++ b/lnbits/decorators.py
@@ -168,7 +168,6 @@ async def authenticated_user_id(
         account = await _get_account_from_token(access_token)
         return account.id if account else None
 
-
     return None
 
 
--- a/lnbits/decorators.py
+++ b/lnbits/decorators.py
@@ -158,6 +158,19 @@ async def check_user_exists(
     return user
 
 
+async def fetch_user_id(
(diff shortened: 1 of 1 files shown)


arcbtc (approved):

lnbits/decorators.py
```diff
@@ -158,6 +158,19 @@ async def check_user_exists(
...
 
 
+async def authenticated_user_id(
+    access_token: Annotated[Optional[str], Depends(check_access_token)],
+    usr: Optional[UUID4] = None,
+) -> Optional[str]:
+    if usr and settings.is_auth_method_allowed(AuthMethods.user_id_only):
+        return usr.hex
+    if access_token:
+        account = await _get_account_from_token(access_token)
```
dni, line 168: hm i think this should always return an account or raise an exception

motorina0: why?

talvasconcelos (approved):

lnbits/decorators.py
motorina0: renamed to `optional_user_id` to avoid confusion

PR #2571, opened 2024-07-09
[feat] add `default_user_extensions` setting

motorina0:
### Summary 
Have a list of extensions which are pre-enabled for the regular users.

**Steps:**
 - login as admin
 - go to: Server > Server (tab)
 - add the extensions that you want to pre-enable
 - save

image

 - create a new user
 - the selected extensions should be enabled

image

Note: existing users will not get these extensions enabled by default.

### DIFF
--- a/lnbits/core/crud.py
+++ b/lnbits/core/crud.py
@@ -2,7 +2,7 @@ import datetime
 import json
 from time import time
 from typing import Any, Dict, List, Literal, Optional
-from uuid import UUID, uuid4
+from uuid import uuid4
 
 import shortuuid
 from passlib.context import CryptContext
@@ -26,7 +26,6 @@ from lnbits.settings import (
 from .models import (
     Account,
     AccountFilters,
-    CreateUser,
     Payment,
     PaymentFilters,
     PaymentHistoryPoint,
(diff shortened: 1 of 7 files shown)


lnbits/core/crud.py
```diff
@@ -42,63 +41,23 @@
 # --------
 
 
-async def create_user(
```
motorina0, line 45: Note: `create_user` and `create_account` were overlapping. There is now a function in `services` called `create_user_account` which encapsulates the logic from here

lnbits/core/crud.py
```diff
@@ -42,63 +41,23 @@
...
-        f"""
-            INSERT INTO accounts
-            (id, email, username, pass, extra, created_at, updated_at)
-            VALUES (?, ?, ?, ?, ?, {tsph}, {tsph})
-        """,
-        (
-            user_id,
-            data.email,
-            data.username,
-            pwd_context.hash(data.password),
```
dni, line 71: we lost the password hashing

dni: ok i see its in `services.py` now

dni (approved):

arcbtc (approved):

PR #2572, opened 2024-07-09
fix: lib secp256k fails building

dni:
issues in release action

https://github.com/lnbits/lnbits/actions/runs/9855301099

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,7 +43,7 @@ jobs:
   pypi:
     runs-on: ubuntu-latest
     steps:
-      - name: fix secp256k1
+      - name: Install dependencies for building secp256k1
         run: |
           sudo apt-get update
           sudo apt-get install -y build-essential automake libtool libffi-dev libgmp-dev
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,6 +43,10 @@ jobs:
   pypi:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


arcbtc (approved):

motorina0 (approved):

.github/workflows/release.yml
```diff
@@ -43,6 +43,10 @@ jobs:
   pypi:
     runs-on: ubuntu-latest
     steps:
+      - name: fix secp256k1
```
michael1011, line 46:
```suggestion
      - name: Install dependencies for building secp256k1
```

Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM python:3.10-slim-bookworm
 
 RUN apt-get clean
 RUN apt-get update
-RUN apt-get install -y curl pkg-config build-essential libnss-myhostname
+RUN apt-get install -y curl pkg-config build-essential libnss-myhostname automake libtool libffi-dev libgmp-dev
```
michael1011, line 5: Out of scope for this PR I guess, but this is not the most effective way to use apt in a Docker build. Using it like this saves some space in the image: https://github.com/BoltzExchange/boltz-backend/blob/master/docker/boltz/Dockerfile#L30

Also why are dependencies needed to *build* secp256k1 included in the image?

dni: https://github.com/lnbits/lnbits/actions/runs/9855301099

actions fails building it :(

PR #2578, opened 2024-07-10
fix: ci build fails on poetry install

dni:
wip


prusnak:
use `poetry lock --no-update` to not update all dependencies while locking

PR #2580, opened 2024-07-12
fix: always create default wallet for user

motorina0:
The default wallet was created only if the user was accessing the UI right after register.

This fix creates the default wallet also when the user is created via an API call.

### DIFF
--- a/lnbits/core/services.py
+++ b/lnbits/core/services.py
@@ -785,6 +785,7 @@ async def create_user_account(
     email: Optional[str] = None,
     username: Optional[str] = None,
     password: Optional[str] = None,
+    wallet_name: Optional[str] = None,
     user_config: Optional[UserConfig] = None,
 ) -> User:
     if not settings.new_accounts_allowed:
@@ -805,6 +806,7 @@ async def create_user_account(
     password = pwd_context.hash(password) if password else None
 
     account = await create_account(user_id, username, email, password, user_config)
+    await create_wallet(user_id=account.id, wallet_name=wallet_name)
 
     for ext_id in settings.lnbits_user_default_extensions:
         await update_user_extension(user_id=account.id, extension=ext_id, active=True)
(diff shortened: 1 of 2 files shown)


talvasconcelos (approved):

lnbits/core/views/api.py
```diff
@@ -69,8 +69,11 @@ async def api_create_account(data: CreateWallet) -> Wallet:
             status_code=HTTPStatus.FORBIDDEN,
             detail="Account creation is disabled.",
         )
-    account = await create_user_account()
-    return await create_wallet(user_id=account.id, wallet_name=data.name)
+    account = await create_user_account(wallet_name=data.name)
+    wallets = await get_wallets(account.id)
+    assert len(wallets) != 0, "Wallet not created for user."
```
dni, line 74:
```suggestion
    assert len(wallets) == 0, "Wallet not created for user."
```

dni: hm, i dont really like the assert here, i think it should be assert inside crud.

lnbits/core/services.py
```diff
@@ -805,6 +806,7 @@ async def create_user_account(
     password = pwd_context.hash(password) if password else None
 
     account = await create_account(user_id, username, email, password, user_config)
+    await create_wallet(user_id=account.id, wallet_name=wallet_name)
```
dni, line 809:
```suggestion
    wallet = await create_wallet(user_id=account.id, wallet_name=wallet_name)
    assert wallet, "could not create wallet"
```

motorina0: this `assert` is redundant, there is one already in `crud.py/create_wallet`

lnbits/core/views/api.py
motorina0: it is

dni (approved):

PR #2610, opened 2024-07-29
fix: skip pending payment check on void

dni:
skips payment check if voidwallet is active

### DIFF
--- a/lnbits/tasks.py
+++ b/lnbits/tasks.py
@@ -133,8 +133,14 @@ async def check_pending_payments():
     the backend and also to delete expired invoices. Incoming payments will be
     checked only once, outgoing pending payments will be checked regularly.
     """
+    sleep_time = 60 * 30  # 30 minutes
 
     while settings.lnbits_running:
+        funding_source = get_funding_source()
+        if funding_source.__class__.__name__ == "VoidWallet":
+            logger.warning("Task: skipping pending check for VoidWallet")
+            await asyncio.sleep(sleep_time)
+            continue
         start_time = time.time()
         pending_payments = await get_payments(
             since=(int(time.time()) - 60 * 60 * 24 * 15),  # 15 days ago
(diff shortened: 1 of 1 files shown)


motorina0:
why?

dni:
if voidwallet is active or enabled, payments should not be checked

lnbits/tasks.py
```diff
@@ -133,8 +133,14 @@ async def check_pending_payments():
...
     checked only once, outgoing pending payments will be checked regularly.
     """
+    sleep_time = 60 * 30  # 30 minutes
 
     while settings.lnbits_running:
+        funding_source = get_funding_source()
+        if funding_source.__class__.__name__ == "VoidWallet":
+            logger.warning("Task: skipping pending check for VoidWallet")
+            await asyncio.sleep(sleep_time)
+            continue
```
motorina0, line 143: Why not exit the loop and let the task complete. There is no way to switch from `VoidWallet` to a real wallet.
```suggestion
            return
```

lnbits/tasks.py
```diff
@@ -133,8 +133,14 @@ async def check_pending_payments():
     the backend and also to delete expired invoices. Incoming payments will be
     checked only once, outgoing pending payments will be checked regularly.
     """
+    sleep_time = 60 * 30  # 30 minutes
 
     while settings.lnbits_running:
+        funding_source = get_funding_source()
+        if funding_source.__class__.__name__ == "VoidWallet":
```
motorina0, line 140: As a side note:
 - maybe funding sources should have an `info()` method (defined at the `base.Wallet` level as abstract)
 - this method would return info about the funding source supported features. Eg:
```json
{
    "invoices_stream": true, // some funding sources do not implement `paid_invoices_stream()` 
    "real_backend": true // in this case only VoidWallet would have false
}
```

talvasconcelos (approved):

motorina0 (approved):

PR #2611, opened 2024-07-29
bug: update poetry lock

dni:
update packages


pyproject.toml
```diff
@@ -23,18 +23,16 @@ lnurl = "0.4.2"
...
 sse-starlette = "1.6.5"
-typing-extensions = "4.8.0"
 uvicorn = "0.23.2"
 uvloop = "0.19.0"
 websockets = "11.0.3"
 loguru = "0.7.2"
 grpcio = "1.62.2"
 protobuf = "4.24.3"
-pyln-client = "23.8"
+pyln-client = "24.5"
```
prusnak, line 35: do not upgrade this without adjusting the code - i tried to change it earlier and it failed because they changed the API

dni: i found out myself aswell :),  :dagger:

lnbits/db.py
```diff
@@ -409,28 +407,24 @@ def parse_query(cls, key: str, raw_values: list[Any], model: type[TFilterModel])
...
 
     @property
     def statement(self):
-        if self.model and self.model.__fields__[self.field].type_ == datetime.datetime:
-            placeholder = compat_timestamp_placeholder()
-        else:
-            placeholder = "?"
         if self.op in (Operator.INCLUDE, Operator.EXCLUDE):
-            placeholders = ", ".join([placeholder] * len(self.values))
+            placeholders = ", ".join([])
```
dni, line 424: @jackstar12  can you maybe help me here? i cant figure out the statement in the filters with sqlalchemy 1.4 placeholders

lnbits/db.py
```diff
@@ -388,9 +387,8 @@ class Page(BaseModel, Generic[T]):
 class Filter(BaseModel, Generic[TFilterModel]):
     field: str
     op: Operator = Operator.EQ
-    values: list[Any]
-
     model: Optional[type[TFilterModel]]
+    values: Optional[dict] = None
```
jackstar12, line 391: This doesnt make sense as a dictionary since all the keys would be the same (`field`)

dni: it can be key value now, because placeholders can be `:key` not only `?`

dni:
@motorina0 can you tackle jmeter?

.github/actions/prepare/action.yml
```diff
@@ -46,7 +46,10 @@ runs:
 
     - name: Install the project dependencies
       shell: bash
-      run: poetry install
+      run: |
+        poetry install
+        # needed for conv tests
+        poetry add psycopg2-binary
```
prusnak, line 52: Why add it here? It is already in pyproject.toml. Or am I missing something?

prusnak:
needs 1) rebase and 2) fixing jmeter failing test

motorina0 (approved):

PR #2612, opened 2024-07-29
feat: filter response fields for `/api/v1/payments/decode`

motorina0:
Some devices cannot handle a large response payload.
This PR adds the `filter_fields` optional field that only includes the filtered fields. If `filter_fields` is missing or an empty list then all fields are returned.
It only filters top-level fields.

**Request**:
```curl
curl --location 'http://localhost:5000/api/v1/payments/decode' \
--header 'Content-Type: application/json' \
--data '{
    "data": "lnbc230n1pn20zexsp5yh56523hvkysgak3jyv3c4tj8jaljjwchpdr4sxfj0lf87lnshyqpp5u5enyfn0wyxmrcpdhrmc2sa3r385qdq9a2f9ntscnhvqezg334kqdq8xye8xcgcqpjrzjqvnwdyky2hw42nrsnwa5wzcv5lstkpq49amh69z9l59lxuy69qe6xrrpfyqqtlcqqgqqqqqqqqqqqzgq0q9qxpqysgqfqvvxnnd4j2s3yqn6j0xcdjf80ar76dvlrutwsalc6aevhaf6avhtlumuu3pctk3mstr0pk6e72jz5w7th5vtx5gn536z9feqsz5yuspqnahvd",
    "filter_fields": ["amount_msat", "payment_hash"]
}'
```

**Response**:
```json
{
    "amount_msat": 23000,
    "payment_hash": "e53332266f710db1e02db8f78543b11c4f403405ea9259ae189dd80c89118d6c"
}
```

image

### DIFF
--- a/lnbits/helpers.py
+++ b/lnbits/helpers.py
@@ -226,7 +226,7 @@ def decrypt_internal_message(m: Optional[str] = None) -> Optional[str]:
 
 
 def filter_dict_keys(data: dict, filter_fields: Optional[list[str]]) -> dict:
-    if not filter_fields or len(filter_fields) == 0:
+    if not filter_fields:
         # return shallow clone of the dict even if there are no filters
         return {**data}
     _data = {}
--- a/lnbits/helpers.py
+++ b/lnbits/helpers.py
@@ -229,10 +229,4 @@ def filter_dict_keys(data: dict, filter_fields: Optional[list[str]]) -> dict:
     if not filter_fields:
         # return shallow clone of the dict even if there are no filters
         return {**data}
-    _data = {}
-    for key in filter_fields:
-        if key not in data:
-            continue
-        _data[key] = data[key]
-
-    return _data
+    return {key: data[key] for key in filter_keys if key in data}
--- a/lnbits/helpers.py
+++ b/lnbits/helpers.py
@@ -225,8 +225,8 @@ def decrypt_internal_message(m: Optional[str] = None) -> Optional[str]:
     return AESCipher(key=settings.auth_secret_key).decrypt(m)
 
 
-def filter_dict_keys(data: dict, filter_fields: Optional[list[str]]) -> dict:
-    if not filter_fields:
+def filter_dict_keys(data: dict, filter_keys: Optional[list[str]]) -> dict:
+    if not filter_keys:
         # return shallow clone of the dict even if there are no filters
         return {**data}
     return {key: data[key] for key in filter_keys if key in data}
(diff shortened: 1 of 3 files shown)


lnbits/helpers.py
```diff
@@ -223,3 +223,15 @@ def decrypt_internal_message(m: Optional[str] = None) -> Optional[str]:
     if not m:
         return None
     return AESCipher(key=settings.auth_secret_key).decrypt(m)
+
+
+def filter_dict_keys(data: dict, filter_fields: Optional[list[str]]) -> dict:
+    if not filter_fields or len(filter_fields) == 0:
+        return {**data}
```
dni, line 230:
```suggestion
        return data
```
you can just return the dict

motorina0: Not the same thing. `{**data}` returns a shallow clone of the original dict.
If there are filters then a shallow clone with filtered fields will be returned.
Therefore `{**data}` handles the `output` consistent no matter the `inputs`.

prusnak: maybe add a comment so someone does not "fix" this in the future?

talvasconcelos (approved):

lnbits/helpers.py
motorina0: done!

dni (approved):

prusnak:
The filter function can be reduced to:

```
filtered_data = {key: data[key] for key in keys_to_keep if key in data}
```

lnbits/helpers.py
```diff
@@ -223,3 +223,16 @@ def decrypt_internal_message(m: Optional[str] = None) -> Optional[str]:
     if not m:
         return None
     return AESCipher(key=settings.auth_secret_key).decrypt(m)
+
+
+def filter_dict_keys(data: dict, filter_fields: Optional[list[str]]) -> dict:
+    if not filter_fields or len(filter_fields) == 0:
```
prusnak, line 229: Can be shortened:

```suggestion
    if not filter_fields:
```

lnbits/helpers.py
```diff
@@ -223,3 +223,16 @@ def decrypt_internal_message(m: Optional[str] = None) -> Optional[str]:
...
+        # return shallow clone of the dict even if there are no filters
+        return {**data}
+    _data = {}
+    for key in filter_fields:
+        if key not in data:
+            continue
+        _data[key] = data[key]
+
+    return _data
```
prusnak, line 238:
```suggestion
    return {key: data[key] for key in data if key in filter_keys}
```

prusnak (changes requested):

lnbits/helpers.py
motorina0: short and sweet!

PR #2622, opened 2024-08-01
Set a maximium sleep time when retrying to connect to the funding source

blackcoffeexbt:
Set a max wait time of 60 seconds. After 7 retries, all future retries will have a 60 second max wait.

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -206,7 +206,7 @@ async def check_funding_source() -> None:
             break
 
         retry_counter += 1
-        sleep_time = 0.25 * (2**retry_counter)
+        sleep_time = min(0.25 * (2 ** retry_counter), 60)
         logger.warning(
             f"Retrying connection to backend in {sleep_time} seconds... "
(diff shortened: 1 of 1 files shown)


prusnak:
> Set a max wait time of 60 seconds. After 7 retries, all future retries will have a 60 second max wait.

Motivation for this change?

blackcoffeexbt:
> > Set a max wait time of 60 seconds. After 7 retries, all future retries will have a 60 second max wait.
> 
> Motivation for this change?

image

After some retries, the retry time becomes very large (because exponentials) so it makes sense, imo, to have an upper bound. Maybe it should be something other than 60 seconds though.

I've come across this with the NWC funding source, sometimes a relay might be down for a few minutes, or maybe your home internet becomes disconnected for a while, in these cases, it could be that LNbits doesn't try and reconnect for a few days in some circumstances.

Is there a reason we shouldn't have an upper bound?

prusnak (approved):

motorina0 (approved):

PR #2623, opened 2024-08-01
Invoice creation UI: Replace input mask with pattern and inputmode

blackcoffeexbt:
Replacing the input masks with pattern and inputmode has a number of benefits:

+ On mobile, the keyboard shows only numbers (See iamge below)
+ Allows additional specification for currencies that need more 2 decimal places (e.g. Chile, Libya) - http://www.londonfx.co.uk/ccylist.html
+ Improves the UX of entering an invoice amount on mobile. The previous reverse input mask was a bit of a mind bender.

image


motorina0:
Why was `package-lock.json` updated?

motorina0 (approved):

blackcoffeexbt:
> Why was `package-lock.json` updated?

idk, will check and revert. I think it's a node version issue on my side.

prusnak:
also change in `lnbits/static/bundle.min.js` should be reverted

prusnak (changes requested):
also the change in `lnbits/static/bundle.min.js` should be reverted

please run `make check` before commiting changes next time

prusnak (approved):

dni (approved):

PR #2627, opened 2024-08-06
fix: db helpers to be used with timestamps

dni:
those helpers are used in boltz extension and they did not take dates into consideration yet

### DIFF
--- a/lnbits/db.py
+++ b/lnbits/db.py
@@ -65,6 +65,13 @@ def compat_timestamp_placeholder():
         return "?"
 
 
+def get_placeholder(type_: type) -> str:
+    if type_ == datetime.datetime:
+        return compat_timestamp_placeholder()
+    else:
+        return "?"
+
+
 class Compat:
     type: Optional[str] = "<inherited>"
     schema: Optional[str] = "<inherited>"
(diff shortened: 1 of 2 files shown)


lnbits/db.py
```diff
@@ -65,6 +65,13 @@ def compat_timestamp_placeholder():
         return "?"
 
 
+def get_placeholder(type_: type) -> str:
+    if type_ == datetime.datetime:
+        return compat_timestamp_placeholder()
+    else:
```
motorina0, line 71: no need for `else`

motorina0 (approved):

motorina0:
LGTM, but unit tests fail

lnbits/db.py
motorina0: maybe this?

```py
def get_placeholder(model: Any, field: str) -> str:
   type_ = model.__fields__[field].type_
   ...
```

then call like this:
```py
placeholder = get_placeholder(self.model, self.field)
```

looks cleaner and encapsulates the object reflection logic in one place

prusnak (approved):

prusnak:
added one commit with `get_placeholder` refactor as suggested by @motorina0

PR #2638, opened 2024-08-14
feat: add generic lnurl error response handler

dni:
this is used multiple times in extensions to safeguard `views_lnurl.py` endpoint to not return a wrong lnurl error response.

you use it by just setting following on your lnurl router/view

```
withdraw_ext_lnurl = APIRouter(prefix="/api/v1/lnurl")
withdraw_ext_lnurl.route_class = LNURLErrorResponseHandler
```

### DIFF
--- a/lnbits/lnurl.py
+++ b/lnbits/lnurl.py
@@ -1,1 +1,42 @@
-from lnurl import LnurlErrorResponse, decode, encode, handle  # noqa: F401
+from typing import Callable
+
+from fastapi import HTTPException, Request, Response
+from fastapi.responses import JSONResponse
+from fastapi.routing import APIRoute
+from lnurl import LnurlErrorResponse, decode, encode, handle
+from loguru import logger
+
+
+class LnurlErrorResponseHandler(APIRoute):
+    """
+    Custom APIRoute class to handle LNURL errors.
+    LNURL errors always return with status 200 and
+    a JSON response with `status="ERROR"` and a `reason` key.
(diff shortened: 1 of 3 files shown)


prusnak (approved):

lnbits/lnurl.py
```diff
@@ -1 +1,46 @@
...
+        original_route_handler = super().get_route_handler()
+
+        async def custom_route_handler(request: Request) -> Response:
+            try:
+                response = await original_route_handler(request)
+                return response
+            except HTTPException as exc:
+                logger.debug(f"HTTPException: {exc}")
+                response = JSONResponse(
+                    status_code=200,
```
michael1011, line 32: Is that spec? HTTP status 200 quite literally means `OK` and is *bad* practice for error responses

dni: i was surprised aswell, it has been like that for a log time, zeus and WoS will wallets error out with unexpected error, as per spec. https://github.com/lnurl/luds/blob/luds/01.md#http-status-codes-and-content-type

```
Neither status codes or any HTTP Header has any meaning. Servers may use whatever they want. Clients should ignore them (and be careful when using libraries that treat responses differently based on headers and status codes) and just parse the response body as JSON, then interpret it accordingly.
```
https://github.com/lnurl/luds/blob/luds/02.md
```{"status": "ERROR", "reason": "error details..."}```

you always return 200 on matter what

prusnak:
the last commit introduces a circular import reported by the CI

arcbtc (approved):

PR #2644, opened 2024-08-20
feat: add baseurl to admin ui settings

dni:
server tab with a hint that it is currently not used.

i ran into an issue developing an extension where i needed to know the url inside a task, where i cannot pass the `Request` object. so i depend on `settings.lnbits_baseurl` there

### DIFF
--- a/lnbits/core/templates/admin/_tab_server.html
+++ b/lnbits/core/templates/admin/_tab_server.html
@@ -9,8 +9,7 @@
           <q-input
             filled
             v-model.number="formData.lnbits_baseurl"
-            label="Base URL"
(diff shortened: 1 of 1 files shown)


arcbtc (approved):

remcoros:
Have you considered that relying on a statically configured base url could create issues when exposing lnbits on multiple domains. E.g. an http .onion service and an https clearnet domain.

arcbtc:
Its more a fix for broken proxies. Its an optional feature. Maybe the tooltip should explain more clearly @dni ?

motorina0 (approved):

Issue #2649, opened 2024-08-26
[BUG] Encrypted Macaroon and Admin Macaroon problem for LndWallet 


vicariousdrama:
**Describe the bug**

LndWallet (GRPC) can't update Admin macaroon after initialization

In lnbits/lnbits/wallets/lndgrpc.py, the macaroon is assigned the value of settings.lnd_grpc_macaroon which is set during initialization.

The Admin Macaroon field is available in the UI for setting lnd_grpc_admin_macaroon, but will have no impact since lnd_grpc_macaroon is already set.

So the only way to update the macaroon appears to be via the Encrypted Macaroon option.

The guidance for this according to the docs (https://docs.lnbits.org/guide/wallets.html) is to run 
`poetry run python lnbits/wallets/macaroon/macaroon.py`.  On a stock clone setup, this unfortunately results in an error as loguru is not found by default. While external, even after fulfilling that dependency, the dependencies for lnbits itself wont be met without building the whole project. To rectify, I can spin up a docker instance of lnbits and run this

```sh
cd /app
poetry run python lnbits/wallets/macaroon/macaroon.py
```
And then paste in the hexadecimal for the macaroon. Next, it wants a password because AESCipher wasn't initialized with a key.  
Providing no password produces an encrypted value, but this isn't going to get very far as the decrypt operation also wants to prompt for a password.  Indeed, on startup, providing a value for Encrypted Macaroon will result in this error

```sh
Enter macaroon decryption password:/usr/local/lib/python3.10/getpass.py:91: GetPassWarning: Can not control echo on the terminal.
  passwd = fallback_getpass(prompt, stream)
Warning: Password input may be echoed.
```

**To Reproduce**

1. Initialize an LNbits instance with LndWallet settings
2. Attempt to change the macaroon
3. New macaroon cannot take effect, either by using Admin Macaroon field (because its subordinate to the encrypted macaroon field), or the Encrypted Macaroon field (because of the bug with password handling)

**Expected behavior**

If a HEX value for macaroon or file path is provided in Admin Macaroon field, it should update the lnd_grpc_macaroon value in addition ot the lnd_grpc_admin_macaroon.

It should be possible to provide an Encrypted macaroon with no password, or else a default internal password should be used or perhaps based on the AUTH_SECRET_KEY. In any event, encrypting documentation will need to be updated to reflect what value the user should specify for password.

**Screenshots**

[...]

PR #2650, opened 2024-08-26
[WIP] Init key for LND encrypt/decrypt

vicariousdrama:
UT.
Potential resolution for #2649, using the auth_secret_key


vicariousdrama:
This is only a WIP because I have not yet been able to get the build environment for lnbits setup on my local system to test this end to end.  We need to do more to improve the documentation in docs/devs/development.md as there are dependencies that need fulfilled in order to build that are not addressed in the Makefile or pyproject.toml

prusnak:
> as there are dependencies that need fulfilled in order to build that are not addressed in the Makefile or pyproject.toml

Can you open another issue stating what your problem is?

vicariousdrama:
I've created another issue with notes regarding whats needed to setup to build.
I'm able to successfully build, and relatedly, was able to test this change.
It's now working as I'd expect to at least leverage encrypted macaroon as a workaround to change the macaroon in use.

vicariousdrama:
Summarized from #2649 notes

Here's what I see from a stock setup, but with LNBITS_ADMIN_UI=true, starting as VoidWallet and other defaults. Attempt to change with Admin Macaroon field on LndWallet. Logs show this after restart

image

In the UI, there is no longer a reference to this value 
'/home/bob/.lnd/data/chain/bitcoin/mainnet/admin.macaroon'

Set to a read only / invoice only macaroon using Encrypted Macaroon field
image

Set to macaroon with pay permissions (not full admin.macaroon, but custom to least privileges), paying an invoice generated on same node (but external to this lnbits instance)
image

Then attempt to pay to an external node
image

lnbits/wallets/lndgrpc.py
```diff
@@ -80,9 +80,9 @@ def __init__(self):
         )
         encrypted_macaroon = settings.lnd_grpc_macaroon_encrypted
         if encrypted_macaroon:
-            macaroon = AESCipher(description="macaroon decryption").decrypt(
-                encrypted_macaroon
-            )
+            macaroon = AESCipher(
+                key=settings.auth_secret_key, description="macaroon decryption"
```
motorina0, line 84: Why is `settings.auth_secret_key` used here, this secret is related to the auth logic (for signing JWTs)?

Should a new `encryption_key` field be added for `LndRestFundingSource` & `LndGrpcFundingSource`

lnbits/wallets/macaroon/macaroon.py
```diff
@@ -2,6 +2,7 @@
 
 from loguru import logger
 
+from lnbits.settings import settings
```
motorina0, line 5: - settings should not be imported into this utility file
- better to read the secret from `input` (just as it is done with the `macaron` on line `42`)

lnbits/wallets/lndgrpc.py
vicariousdrama: If a key is not provided in the initialization of AESCipher here, then the call to decrypt will fail when the server runs as there is no user input in headless mode.  This is because the decrypt method in lnbits/lnbits/utils/crypto.py calls for passphrase, which checks if key is initialized or else prompts for it.

Any of the funding sources that use encryption would need to be updated. Should these use lnbits/helpers.py instead?

lnbits/wallets/macaroon/macaroon.py
vicariousdrama: It's not viable to read the settings from input here as that causes a server error on startup.

PR #2656, opened 2024-08-28
Fix lnurl fallback

### LINKED ISSUE #1599: [BUG] Does not support LNURL LUD-01 Fallback Schemes
## Description

[This](https://github.com/lnurl/luds/blob/luds/01.md#fallback-scheme) seems to not be supported as yet when sending from wallet. Can test with the QR code [here](https://pay.blink.sv/vindard/print) for reference which resolves to `https://pay.blink.sv/VINDARD?LIGHTNING=LNURL1DP68GURN8GHJ7URP0YHXYMRFDE4JUUMK9UH8WETVDSKKKMN0WAHZ7MRWW4EXCUP0WE5KUERPWFJQSU3QFK`

talvasconcelos:
Wrong string case search

Closes #1599

### DIFF
--- a/lnbits/static/js/wallet.js
+++ b/lnbits/static/js/wallet.js
@@ -225,6 +225,7 @@ new Vue({
         this.parse.data.request = this.parse.data.request.slice(6)
       } else if (req.indexOf('lightning=lnurl1') !== -1) {
         this.parse.data.request = this.parse.data.request
+          .toLowerCase()
           .split('lightning=')[1]
           .split('&')[0]
       }
--- a/lnbits/static/js/wallet.js
+++ b/lnbits/static/js/wallet.js
@@ -221,9 +221,9 @@ new Vue({
       let req = this.parse.data.request.toLowerCase()
 
       if (req.startsWith('lightning:')) {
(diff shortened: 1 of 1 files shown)


dni (approved):

lnbits/static/js/wallet.js
```diff
@@ -225,6 +225,7 @@ new Vue({
         this.parse.data.request = this.parse.data.request.slice(6)
       } else if (req.indexOf('lightning=lnurl1') !== -1) {
```
prusnak, line 226: should we also use req.toLowerCase here?

talvasconcelos: the variable was set, probably to avoid writing `this.parse.data.request` a few times... it wasn't used, BTW... 

could this all block code could be optimized? yes...

talvasconcelos: can you check with latest commit, if it looks better? at least i think it reduces a couple lines

prusnak: Yes. Looks much more clean now.

prusnak (approved):

PR #2619, opened 2024-07-31
chore: update pyproject dependencies

prusnak:
i did not touch sqlalchemy since this is being worked on here: https://github.com/lnbits/lnbits/pull/2611

PR #2664, opened 2024-08-30
simple bash install

arcbtc:
Just creates a bash script that can be used to run and install. Has fail-safe in case someone doesnt read install guide and uses from root directory.

```
wget https://raw.githubusercontent.com/lnbits/lnbits/snapcraft/lnbits.sh &&
chmod +x lnbits.sh &&
./lnbits.sh
```


lnbits.sh
```diff
@@ -0,0 +1,50 @@
...
+  if [ -d lnbits ]; then # Someone didnt follow the install guide
+    # Check out the main branch
+    git checkout main
+  else
+    # Clone the LNbits repository
+    git clone https://github.com/lnbits/lnbits.git
+    cd lnbits
+
+    # Check out the main branch
+    git checkout main
```
motorina0, line 27: `git checkout main` runs on both branches of the `if`. Is the `if` required?

motorina0 (approved):
LGTM
see comment

prusnak:
I am wondering if this method should not be the number two method - while keeping poetry as recommended option one.

Also I think we should not install soon-to-be unsupported Python 3.9. We should just make sure that LNbits run with all supported versions (up to 3.12 - PR https://github.com/lnbits/lnbits/pull/2619 takes care of this).

arcbtc:
> I am wondering if this method should not be the number two method - while keeping poetry as recommended option one.
> 
> Also I think we should not install soon-to-be unsupported Python 3.9. We should just make sure that LNbits run with all supported versions (up to 3.12 - PR #2619 takes care of this).

Sure, lets move to option 2, sqlalchemy seems like a big job tho and will require a good few weeks worths of work.

lnbits.sh
arcbtc: fixed

PR #2667, opened 2024-09-03
fix: internal extra was missing

dni:
- refactoring create_payment a bit to use a model instead of 10 kwargs


motorina0:
This will break a lot of extension. How do we handle that?

prusnak:
> This will break a lot of extension. How do we handle that?

We are breaking all extensions with 1.0.0 so I guess now it's a good time?

motorina0:
> We are breaking all extensions with 1.0.0 so I guess now it's a good time?

Yeah, but its going to require a bit of testing too :)

dni:
no it wont,i checked it, `create_payment` should not be used by extensions and isn't used! exts should use `create_invoice` and `pay_invoice`. `create_payment` is more lowlevel to core

motorina0 (approved):

PR #2669, opened 2024-09-03
feat: explicitly export for extensions from `__init__.py`

dni:
makes it clear what extensions are expected to use and also makes future changes to the structure of core safer because extension can just depend on ```from lnbits import require_admin_key``` for example.

TODO: still couple of exports from `crud.py` are missing, still thinking about making proper services.py fn's instead of extension import from core crud


motorina0:
I can't figure out why the tests fail for this PR...

dni:
> I can't figure out why the tests fail for this PR...

it is weird, isnt it? i have no clue aswell, ci fails seems unrelated...

Issue #2540, opened 2024-05-29
[BUG] Lightning Addresses no longer work


tiker:
**Describe the bug**
LN Addresses do not appear to be functional now.  I noticed a problem with 0.12.6 and just updated to 0.12.8 but still doesn't seem to work.  I have had the LNURLP extension loaded and configured for quite a while, receiving payments to my account but it stopped working with one of the recent updates.

**To Reproduce**
I created a test account on https://demo.lnbits.com to verify that it is not isolated to my node.

1. I have created a test account - https://demo.lnbits.com/wallet?usr=3c8518c4915b4699a0bb0ee677c3876a
2. Enabled the LNURLP (Pay Links) extension
3. Created a new payment link with the address "test123@demo.lnbits.com"
4. Viewing https://demo.lnbits.com/.well-known/lnurlp/test123 returns a JSON document with "address not found".
5. LN decoder pages such as https://lightningdecoder.com/test123@demo.lnbits.com also report the same error.

**Expected behavior**
The expected behavior would be a valid response with payment details?  (Not sure what the correct wording is.)  The functionality that used to work.

**Additional context**
The account linked above is a test account for this issue.  Feel free to play around with it, use it for troubleshooting, delete it if you like, whatever.

bitkarrot:
Payments to Lightning addresses appear to be working on https://demo.lnbits.com which is currently running

 LNBits **0.12.9** and Pay links extension version 0.5.0  as of Sun Jul 28 05:51:52 UTC 2024

Tested: 
1.  https://demo.lnbits.com/lnurlp/link/bRRLRy
2.  bitkarrot@demo.lnbits.com 
<img width="495" alt="Screenshot 2024-07-27 at 10 58 15 PM" src="https://github.com/user-attachments/assets/50795f8b-6325-46a4-b17f-8435d9f7cd49">



Note: lnbits v0.12.6 does not work with Pay links 0.5.0 as noted by OP.

prusnak:
Thanks for testing

Closing

prusnak:
> I'm not sure what test was performed but this issue is not fixed.

Are you on the latest LNbits version?

tiker:
This is sending to the official demo instance (demo.lnbits.com) which @bitkarrot claims works.

prusnak:
@bitkarrot did you remove the alias `bitkarrot@demo.lnbits.com` after testing?

tiker:
My tests address was not removed and still does not work.
[...]

PR #2671, opened 2024-09-04
Fix overlapping redirect paths

motorina0:
### Summary
Derived from: https://github.com/lnbits/lnbits/issues/2540
If two extensions (`ext_a` and `ext_b`) have overlapping redirect paths then the second one (`ext_b`) will be installed but deactivated.
If `ext_a` is deactivated, then `ext_b` can be activated.
See test scenario here:
image

The PR with JMeter tests: https://github.com/lnbits/lnbits-extensions/pull/415

This PR extended a bit outside the scope defined in the title. Some of the concepts in extension manager got clearer and I've decide to also refactor the extension manager and add tests to guard for future bugs.

### Changes
Short explanation of the relevant changes per file:
- `app.py` (refactor)
  - extract redirect logic from this file to `settings.activate_extension_paths`

- `commands.py` (refactor)
  - update function calls and imports
- `crud.py` (code quality only)
- `extension_manager.py`
    - moved to the `core/extensions` directory
    - all the classes have been moved to `extensions/models.py`
    - utilities have been moved to `extensions/helpers.py`
    - it contains only the logic for install/uninstall, activate/deactivate and stop_background_work
      - `stop_extension_background_work` changes:
        - do not call via API anymore (all extensions have `_start` & `_stop` now)
        - fix `stop_fn()` in case it is not an `async` function
 - `extensions/helpers.py`
   -  helpers extracted from the `extension_manager.py` (no new logic)
 - `extensions/models.py`
   - classes extracted from  `extension_manager.py` (no new logic, some standalone functions are now `@classmethods`)
- `extensions_api.py` (extract some methods and update imports) 
- `middleware.py` (simplify, extract path logic to its own `RedirectPath` class)
- `settings.py`
  - encapsulate the path redirect logic here instead of having it in multiple places in the app
  - fix bug for update extension
- `test_settings.py` (add tests for the `RedirectPath` class)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,8 +49,8 @@ fly.toml
 lnbits-backup.zip
 
 # Ignore extensions (post installable extension PR)
-extensions
-upgrades/
+/extensions
+/upgrades/
 
 # builded python package
 dist
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -383,13 +383,7 @@ def register_ext_routes(app: FastAPI, ext: Extension) -> None:
 
     if hasattr(ext_module, f"{ext.code}_redirect_paths"):
         ext_redirects = getattr(ext_module, f"{ext.code}_redirect_paths")
-        print("### ext_redirects", ext.code, ext_redirects)
-        print("### lnbits_extensions_redirects 0", settings.lnbits_extensions_redirects)
-
-        print("### lnbits_extensions_redirects 1", settings.lnbits_extensions_redirects)
         activate_extension_settings(ext.code, ext_redirects)
-
-        print("### lnbits_extensions_redirects 2", settings.lnbits_extensions_redirects)
 
     logger.trace(f"adding route for extension {ext_module}")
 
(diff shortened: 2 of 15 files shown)


prusnak:
Can you describe in words how do you fix the issue?

What happens if two extensions register the same path?

motorina0:
> Can you describe in words how do you fix the issue?
> 
> What happens if two extensions register the same path?

OMG @prusnak you're like the `eye of sauron`, but with good intentions :)

The plan is simple: 
  - do not allow the second extension to be installed if there is a conflict
  - inform the admin user about the nature of the conflict and the name of the other extension that it conflicts with
  - similar with two apps listening on the same port

motorina0:
> The plan is simple:
> 
>     * do not allow the second extension to be installed if there is a conflict
> 
>     * inform the admin user about the nature of the conflict and the name of the other extension that it conflicts with
> 
>     * similar with two apps listening on the same port

More details and final decision once the PR is out of draft.

prusnak:
> OMG @prusnak you're like the `eye of sauron`, but with good intentions :)

sorry, I don't want to be a pain in the ass, I was just very curious how to solve this particular issue since I first stumbled upon it for the first time a few weeks ago :

arcbtc (approved):

PR #2673, opened 2024-09-05
feat: introduce `NotFoundError`

dni:
instead of asserting and returning Optional on get operation in crud, we through and NotFoundError. in the context of an API request the FastApi errorhandler will catch it and throw an 404 like we have for Invoice- and PaymentError.

In non-api context we can simly try catch the notfound like so:
```python
try:
  url = await get_tinyurl("test")
except NotFoundError:
  # url does not exist do whatever
  pass
```


dni:
its a very much used and well known pattern for fastapi and not only there. it simlifies the code a lot and is used all around.
https://blog.stackademic.com/error-and-exception-handling-in-fastapi-c0949bb42e1b

PR #2676, opened 2024-09-05
refactor: remove `get_key_type` decorator

dni:
its always better to explicitly choose `require_invoice_key` or `require_admin_key` makes the code more readable and you always know upfront which key you need.

breaking change for 1.0.0


motorina0:
Just curious:
 1. does a `require_invoice_key` work if an admin key is provided?
 1. what extension are using this decorator?

dni:
1. yes
2.  all of them, i think it was there from the start of lnbits

motorina0:
I guess the JMeter tests will fail until we update all the extensions.
Approved but, we should only merge this PR after we take care of the extensions.

motorina0 (approved):

PR #2678, opened 2024-09-06
Check for theme params on the URL

talvasconcelos:
- Fixed an inconsistency on the apply gradient, where the background color wouldn't take effect until refresh

- added to base.html a snippet to look for URL params related to theme
- allow for enforcing the desired styling for display pages (tpos, events, tickets, etc...)


lnbits/templates/base.html
```diff
@@ -258,6 +258,62 @@
         { value: 'fi', label: 'Suomi', display: '🇫🇮 FI' }
       ]
     </script>
+    <script>
+      const currentUrl = window.location.href
+      let url = new URL(currentUrl)
```
motorina0, line 263: why not `const`?

lnbits/templates/base.html
```diff
@@ -258,6 +258,62 @@
...
+      let params = new URLSearchParams(url.search)
+      let q = Quasar.plugins.LocalStorage
+
+      const setColors = () => {
```
motorina0, line 267: why define a separate internal function if only called once?

arcbtc: isnt it neater?

dni, line 267: i think this should be on `created()` in base.js not in the html, i would only put variables there

talvasconcelos: it was already set this way... i can change it, but looks neater as @arcbtc said!

does it have memory/performance impact?

motorina0 (approved):

PR #2358, opened 2024-03-25
feat: add boltz client fundingsource

dni:
https://docs.boltz.exchange/v/boltz-client

this fundingsource utilizing the boltz client standalone function: https://github.com/BoltzExchange/boltz-client/pull/123 this makes him act like a lightning node while submarine swapping everything on liquid network. like aqua does in its wallet. its basically a liquid wallet.

live mainnet instance: https://boltz.dni.guru/

PR #2681, opened 2024-09-11
feat: add Breez *Liquid* SDK  funding source

riccardobl:
A funding source that uses the liquid variant of breez sdk for lightning over liquid.
Running lnbits with this is the same as https://github.com/lnbits/lnbits/pull/2358 but without the need of any additional daemon, in fact,  it only needs a seed phrase, making it a very low barrier "non custodial" option (with the known liquid compromises).
Since there is a receiving fee when using boltz swaps, i had to add a field to InvoiceResponse to account for that.

I originally built this for my own use, but I thought it might be a valuable addition for others as well, so I submitted a PR, feel free to share your thoughts.

Also, i don't know how to write tests for this, but we can probably figure something out.


pyproject.toml
```diff
@@ -59,6 +59,7 @@ python-crontab = "3.2.0"
 wallycore = {version = "1.3.0", optional = true}
 # needed for breez funding source
 breez-sdk = {version = "0.5.2", optional = true}
+breez-sdk-liquid = "^0.3.0"
```
dni, line 62: should be optional aswell

dni:
very cool!

lnbits/core/services.py
```diff
@@ -167,18 +164,18 @@ async def create_invoice(
...
         payment_request=payment_request,
         payment_hash=invoice.payment_hash,
-        amount=amount_msat,
+        amount=1000 * amount_sat,
         expiry=invoice.expiry_date,
         memo=memo,
         extra=extra,
         webhook=webhook,
         conn=conn,
+        fee=abs(fees_msats or 0),
```
prusnak, line 178: using `abs()` feels dirty

what do you think @dni? are there funding sources that return negative fee and some funding sources that return positive fee? if that is the case, we should fix the funding sources to always return positive fee.

pyproject.toml
```diff
@@ -59,9 +59,11 @@ python-crontab = "3.2.0"
 wallycore = {version = "1.3.0", optional = true}
 # needed for breez funding source
 breez-sdk = {version = "0.5.2", optional = true}
+breez-sdk-liquid = {version = "0.3.0", optional = true}
 
 [tool.poetry.extras]
 breez = ["breez-sdk"]
+breez_liquid = ["breez-sdk-liquid"]
```
prusnak, line 66: let's use dash instead of underscore
```suggestion
breez-liquid = ["breez-sdk-liquid"]
```

lnbits/core/services.py
```diff
@@ -148,17 +148,14 @@ async def create_invoice(
...
-        checking_id,
-        payment_request,
-        error_message,
-    ) = await funding_source.create_invoice(
-        amount=amount_sat,
-        memo=invoice_memo,
-        description_hash=description_hash,
-        unhashed_description=unhashed_description,
-        expiry=expiry or settings.lightning_invoice_expiry,
+    (ok, checking_id, payment_request, error_message, fees_msats) = (
```
prusnak, line 151:
```suggestion
    ok, checking_id, payment_request, error_message, fees_msats = (
```

prusnak (changes requested):

.env.example
```diff
@@ -134,6 +134,9 @@ BREEZ_GREENLIGHT_INVITE_CODE=CODE
 BREEZ_GREENLIGHT_DEVICE_KEY="/path/to/breezsdk/device.pem"  # or BASE64/HEXSTRING
 BREEZ_GREENLIGHT_DEVICE_CERT="/path/to/breezsdk/device.crt"  # or BASE64/HEXSTRING
 
+# BreezLiquidSdkWallet
+BREEZ_MNEMONIC="MNEMONIC SEED PHRASE"
```
prusnak, line 138: 1. let's use BREEZ_LIQUID prefix to differentiate this from BreezSdkWallet
2. let's use seed instead of mnemonic, since BreezSdkWallet above uses this term
3. this has to be changed in `docs/guide/wallets.md` too and of course in the code

```suggestion
BREEZ_LIQUID_SEED=SEED
```

prusnak (changes requested):

lnbits/core/services.py
dni: i think its ok for now, tried to tackle it here, but somehow i broke regtest and couldnt relally figure it out. https://github.com/lnbits/lnbits/pull/2468 going to give it another try sometime

dni (approved):

lnbits/settings.py
```diff
@@ -305,7 +305,8 @@ def fee_reserve(self, amount_msat: int, internal: bool = False) -> int:
             return 0
         reserve_min = self.lnbits_reserve_fee_min
         reserve_percent = self.lnbits_reserve_fee_percent
-        return max(int(reserve_min), int(amount_msat * reserve_percent / 100.0))
+        _max = max(int(reserve_min), int(abs(amount_msat) * reserve_percent / 100.0))
+        return _max
```
motorina0: why temp var?

lnbits/core/services/payments.py
```diff
@@ -288,6 +288,7 @@ async def create_invoice(
         memo=memo,
         extra=extra,
         webhook=webhook,
+        fee=abs(payment_response.fee_msat or 0),
```
motorina0: fee should be negative (handled here on on crud level)

lnbits/wallets/breez_liquid.py
```diff
@@ -0,0 +1,289 @@
...
+                ) from exc
+
+        async def cleanup(self):
+            self.sdk_services.disconnect()
+
+        async def status(self) -> StatusResponse:
+            try:
+                info: breez_sdk.GetInfoResponse = self.sdk_services.get_info()
+            except Exception as exc:
+                return StatusResponse(f"Failed to connect to breez, got: '{exc}...'", 0)
```
motorina0, line 113:
```suggestion
                logger.warning(exc)
                return StatusResponse(f"Failed to connect to breez, got: '{exc}...'", 0)
```

lnbits/wallets/breez_liquid.py
```diff
@@ -0,0 +1,289 @@
...
+        async def create_invoice(
+            self,
+            amount: int,
+            memo: Optional[str] = None,
+            description_hash: Optional[bytes] = None,
+            unhashed_description: Optional[bytes] = None,
+            **_,
+        ) -> InvoiceResponse:
+            try:
+                # TODO: issue with breez sdk, receive_amount is of type BITCOIN
```
motorina0: is there anything left `TODO` here?

lnbits/wallets/breez_liquid.py
```diff
@@ -0,0 +1,289 @@
...
+        async def pay_invoice(
+            self, bolt11: str, fee_limit_msat: int
+        ) -> PaymentResponse:
+            invoice_data = bolt11_decode(bolt11)
+
+            try:
+                prepare_req = breez_sdk.PrepareSendRequest(destination=bolt11)
+                req = self.sdk_services.prepare_send_payment(prepare_req)
+
+                fee_limit_sat = 50 + int(fee_limit_msat / 1000)
```
motorina0: what is the `50` magic number?
is this something that can be part of `settings`?

lnbits/wallets/breez_liquid.py
```diff
@@ -0,0 +1,289 @@
...
+                        ),
+                    )
+
+                send_response = self.sdk_services.send_payment(
+                    breez_sdk.SendPaymentRequest(prepare_response=req)
+                )
+
+            except Exception as exc:
+                logger.warning(exc)
+                # assume that payment failed?
```
motorina0: No, do not assume that. If the payment is actually successful then the amount will not be subtracted from the LNbits wallet.

It must return `pending` and let the user or the background task to re-check the status later.

lnbits/wallets/breez_liquid.py
```diff
@@ -0,0 +1,289 @@
...
+            return PaymentResponse(
+                ok=True,
+                checking_id=checking_id,
+                fee_msat=payment.fees_sat * 1000,
+                preimage=payment.details.preimage,
+            )
+
+        def _find_payment(self, payment_hash: str) -> Optional[breez_sdk.Payment]:
+            offset = 0
+            while True:
```
motorina0:
```suggestion
            while settings.lnbits_running:
```

lnbits/wallets/breez_liquid.py
```diff
@@ -0,0 +1,289 @@
...
+                list_payments_request = breez_sdk.ListPaymentsRequest(
```
motorina0: This is not good, it will iterate over all payments each time there is a status check.

There has to be a way to check by `payment_hash` of some other params.

lnbits/wallets/breez_liquid.py
```diff
@@ -0,0 +1,289 @@
...
+                    offset=offset, limit=100
+                )
+                history: list[breez_sdk.Payment] = self.sdk_services.list_payments(
+                    req=list_payments_request
+                )
```
motorina0: there is this:
```py
self.sdk_services.get_payment(GetPaymentRequest(???))
```

lnbits/wallets/breez_liquid.py
dni: 50 is some kind of baseline fee for the onchain swap

dni: yes i think it makes sense to make that configurable

lnbits/settings.py
```diff
@@ -552,6 +552,7 @@ class BreezSdkFundingSource(LNbitsSettings):
 class BreezLiquidSdkFundingSource(LNbitsSettings):
     breez_liquid_api_key: str | None = Field(default=None)
     breez_liquid_seed: str | None = Field(default=None)
+    breez_liquid_fee_limit_offset: int = Field(default=50)
```
motorina0: should be added to `.env.example` too

motorina0: add a comment, not clear what this value if for, not clear if tis `sat` or `msat`

motorina0 (approved):

PR #2703, opened 2024-09-23
[feat] Nostr Login

motorina0:
### Summary
- allow users to login with their nostr pubkey
- different nostr browser extensions can be used
- if a user does not exist it will be created
- an existing user can configure its public key in the `account` page then it will be able to login with nostr
- as an extra security feature the user only has 2 minutes after login to change its credentials (password or pubkey). Then he has to re-login.

### Screen Recording
Configure server and login with Alby

https://github.com/user-attachments/assets/202e0d23-021a-4a39-a829-945400b84364

Login with nos2x:

https://github.com/user-attachments/assets/761032ee-2c64-4369-8ba8-d2b156226064

Set pubkey to existing user:

https://github.com/user-attachments/assets/f0edf526-2ee6-45f8-966d-53862346cd89

Try to update credentials after treshold:

https://github.com/user-attachments/assets/20ee9fd4-0a66-442b-8c63-af5c8059365a


bitkarrot:
👀 Hell yes, with Nos2x extension?! Thank you @motorina0

bitkarrot:
I will be more than happy to be the first to test this.

motorina0:
@bitkarrot thank you for your interest, the PR is now ready to review

The only thing open is adding more tests and improving some labels.

bitkarrot:
Ah I thought this would include something like a NIP05 login, but this is more general just npub. Looks good so far.

motorina0:
> Ah I thought this would include something like a NIP05 login, but this is more general just npub. Looks good so far.

Is there a nip for nip5 login?

lnbits/core/models.py
```diff
@@ -184,8 +185,12 @@ class UpdateUserPassword(BaseModel):
     user_id: str
     password: str = Query(default=..., min_length=8, max_length=50)
     password_repeat: str = Query(default=..., min_length=8, max_length=50)
-    password_old: Optional[str] = Query(default=None, min_length=8, max_length=50)
```
dni, line 187: was this not needed?

lnbits/settings.py
```diff
@@ -443,11 +444,20 @@ class AuthSettings(LNbitsSettings):
             AuthMethods.username_and_password.value,
         ]
     )
+    # How many seconds after login the user is allowed to update its credentials.
+    # A fresh login is required afterwards.
+    auth_credetials_update_threshold: int = Field(default=120)
```
dni, line 449: i like it good security feature

lnbits/static/js/account.js
```diff
@@ -13,11 +13,13 @@ window.app = Vue.createApp({
         'confettiStars'
       ],
       tab: 'user',
-      passwordData: {
+      credentialsData: {
         show: false,
         oldPassword: null,
```
dni, line 18: there still is oldPassword

dni (commented):
nice!

dni:
ci fails ```FAILED tests/api/test_auth.py::test_register_ok - AssertionError: User created.
assert 401 == 200
 +  where 401 = <Response [401 Unauthorized]>.status_code```

bitkarrot:
> > Ah I thought this would include something like a NIP05 login, but this is more general just npub. Looks good so far.
> 
> Is there a nip for nip5 login?

I don't think there is a nip other than https://github.com/nostr-protocol/nips/blob/master/05.md, but there is this interesting thing they are doing with nsec bunker apps like this on nostr-login where you can login with a username@domain to a  keystore  like this: 


Screenshot 2024-09-26 at 1 36 57 AM

bitkarrot:
Ok there is also NIP46 https://github.com/nostr-protocol/nips/blob/4438b892d8abd189afc75f9f54532ec95f108d9f/46.md

lnbits/static/js/account.js
motorina0: yes, it is still possible to change passwords

dni (approved):

PR #2718, opened 2024-09-26
feat: Create Appimage/dmg/flatpack workflows

arcbtc:
Being able to easily install Lnbits is important. On releases we should be creating assets.
GH actions/workflows to create:
* Appimage ☐
* dmg ☐
* Flatpack  ☐


bitkarrot:
https://www.cloudron.io/

PR #2731, opened 2024-10-07
fix(boltz): sanitize invoice description

jackstar12:
Boltz rejects nbsp char (produced by JS Intl.NumberFormat api), so simply replace it with normal space

### DIFF
--- a/lnbits/wallets/boltz.py
+++ b/lnbits/wallets/boltz.py
@@ -95,8 +95,10 @@ class BoltzWallet(Wallet):
             wallet_id=self.wallet_id,
             accept_zero_conf=True,
             external_pay=True,
-            description=memo,
         )
+        if memo is not None:
+            # boltz rejects nbsp char (produced by JS Intl.NumberFormat api)
+            request.description = memo.replace("\xa0", " ")
         response: boltzrpc_pb2.CreateReverseSwapResponse
         try:
             response = await self.rpc.CreateReverseSwap(request, metadata=self.metadata)


lnbits/wallets/boltz.py
```diff
@@ -95,7 +95,8 @@ async def create_invoice(
             wallet_id=self.wallet_id,
             accept_zero_conf=True,
             external_pay=True,
-            description=memo,
+            # boltz rejects nbsp char (produced by JS Intl.NumberFormat api)
+            description=memo.replace(" ", " "),  # noqa: RUF001
```
motorina0, line 99: it looks like space char is replaced by space char?
am I not seeing something right?

motorina0:
lint failing

lnbits/wallets/boltz.py
jackstar12: Browser, renders it the same way, but its a different char under the hood.
For example, pycharm:
image

lnbits/wallets/boltz.py
```diff
@@ -95,8 +95,10 @@ async def create_invoice(
             wallet_id=self.wallet_id,
             accept_zero_conf=True,
             external_pay=True,
-            description=memo,
         )
+        if memo is not None:
+            # boltz rejects nbsp char (produced by JS Intl.NumberFormat api)
+            request.description = memo.replace(" ", " ")  # noqa: RUF001
```
prusnak, line 101: IMO this is better: 

```suggestion
            request.description = memo.replace("\xa0", " ")
```

prusnak (changes requested):

lnbits/wallets/boltz.py
motorina0: I agree, its much clearer.
@jackstar12 can you please test and commit if all good

jackstar12: ill test real quick

jackstar12: works

prusnak (approved):

motorina0 (approved):

PR #2741, opened 2024-10-18
Fix db cache

motorina0:


lnbits/core/helpers.py
```diff
@@ -41,7 +41,8 @@ async def run_migration(
     current_version: Optional[DbVersion] = None,
 ):
     matcher = re.compile(r"^m(\d\d\d)_")
-    for key, migrate in migrations_module.__dict__.items():
+
+    for key, migrate in list(migrations_module.__dict__.items()):
```
motorina0, line 45: otherwise it raises `dictionary changed size during iteration`

lnbits/core/migrations.py
```diff
@@ -597,8 +598,17 @@ async def m026_update_payment_table(db):
...
 
-async def m027_update_apipayments_data(db):
+async def m027_update_apipayments_data(db: Connection):
+    result = None
+    try:
+        result = await db.execute("SELECT * FROM apipayments")
+    except Exception as exc:
+        logger.warning("Could not select, trying again after cache cleared.")
+        logger.debug(exc)
+        await db.execute("COMMIT")
```
motorina0, line 608: `db.conn.commit()` does not work

dni (approved):

PR #2742, opened 2024-10-18
[refactor] split services.py

motorina0:
Split the huge `services.py` into smaller files
Keep the code backwards compatible by using `__all__`


dni (approved):
nice, clean

lnbits/core/services/users.py
```diff
@@ -0,0 +1,128 @@
...
+    assert user, "Cannot find user for account."
+
+    return user
+
+
+async def check_admin_settings():
+    if settings.super_user:
+        settings.super_user = to_valid_user_id(settings.super_user).hex
+
+    if settings.lnbits_admin_ui:
```
dni, line 78: could be if not, do nothing

motorina0: I've just moved code, no other type of refactoring
can be addressed in different PR

PR #2754, opened 2024-10-29
fix: setting the default for dates in model

dni:
how i did it does not work, it actually initialized the default date once on the server start and keeps that. i was quite sure it will recalculate now() on each model initialization, but that was a false assumption.

the fix is to make it a required value, and initalizing it when creating a model

### DIFF
--- a/lnbits/core/crud/payments.py
+++ b/lnbits/core/crud/payments.py
@@ -231,6 +231,8 @@ async def create_payment(
     previous_payment = await get_standalone_payment(checking_id, conn=conn)
     assert previous_payment is None, "Payment already exists"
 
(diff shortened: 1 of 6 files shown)


prusnak:
I think you might want to assign now only once and then to use the variable.

e.g.

```
now = datetime.now(timezone.utc)

payment = Payment(
    ...
    time=now,
    created_at=now,
    updated_at=now
)
```

otherwise you end up with different values for time, created_at and updated_at

dni:
it still a bit annoying, i would love to have them set properly on initializing the model

prusnak:
closing in favor of https://github.com/lnbits/lnbits/pull/2755

PR #2756, opened 2024-10-30
fix: handle list models

motorina0:
Corner case for dict to model.

### DIFF
--- a/lnbits/db.py
+++ b/lnbits/db.py
@@ -626,9 +626,7 @@ def dict_to_submodel(model: type[TModel], value: Union[dict, str]) -> Optional[T
         _subdict = json.loads(value)
     elif isinstance(value, dict):
         _subdict = value
-    else:
-        logger.warning(f"Expected str or dict, got {type(value)}")
-        return None
+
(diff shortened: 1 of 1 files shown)


lnbits/db.py
```diff
@@ -626,9 +626,7 @@ def dict_to_submodel(model: type[TModel], value: Union[dict, str]) -> Optional[T
         _subdict = json.loads(value)
     elif isinstance(value, dict):
         _subdict = value
-    else:
```
motorina0, line 629: `value` can only be `dict` or `str`

lnbits/db.py
```diff
@@ -656,7 +660,7 @@ def dict_to_model(_row: dict, model: type[TModel]) -> TModel:
             else:
                 _dict[key] = value
             continue
-        if issubclass(type_, BaseModel) and value:
+        if issubclass(type_, BaseModel):
```
motorina0, line 663: `value` already checked for null at the beginning of the `for`

dni (approved):

talvasconcelos (approved):

PR #2760, opened 2024-11-03
feat: lnd_rest allow self payments

jjmmbb:
- added the ability to receive self payments on the node

[issue] You have to enable circular payment on lnd.conf. Note that lnd.conf doesn't allow allow_self_payment parameter, that is why is necessary to define that on transaction.

[checkout for more](https://github.com/search?q=repo%3Alightningnetwork%2Flnd%20allow_self_payment&type=code)


prusnak:
What happens if you provide this parameter but circular payments are not enabled in the lnd.conf?

prusnak:
This does not do what you think it does. It only enables circular payments.

motorina0 (approved):

motorina0 (changes requested):
Please run `make format`

jjmmbb:
What do you mean about make format?

jjmmbb:
> Please run `make format`

What do you mean about `make format`?

prusnak:
> > Please run `make format`
> 
> What do you mean about `make format`?

cd lnbits
poetry run make format

jjmmbb:
I think it done.

remcoros:
According to the FAQ here: https://github.com/lnbits/lnbits/blob/dev/docs/guide/faq.md?plain=1#L18

We should set "allow-self-payment=1" in lnd.conf, but I've been scanning the LND source code and cannot find anything about a "allow-self-payment" setting in lnd.conf

According to https://github.com/lightningnetwork/lnd/issues/7419#issuecomment-2039073501 the "--allow_self_payment" is a CLI arg and an RPC parameter.

Currently i'm getting a "no_route" error for every payment I try between wallets on the same instance.

Is the FAQ wrong?
And will this PR fix sending payments between LNBits wallets on the same instance?

jjmmbb:
> According to the FAQ here: https://github.com/lnbits/lnbits/blob/dev/docs/guide/faq.md?plain=1#L18
> 
> We should set "allow-self-payment=1" in lnd.conf, but I've been scanning the LND source code and cannot find anything about a "allow-self-payment" setting in lnd.conf
> 
> According to [lightningnetwork/lnd#7419 (comment)](https://github.com/lightningnetwork/lnd/issues/7419#issuecomment-2039073501) the "--allow_self_payment" is a CLI arg and an RPC parameter.
> 
> Currently i'm getting a "no_route" error for every payment I try between wallets on the same instance.
> 
> Is the FAQ wrong? And will this PR fix sending payments between LNBits wallets on the same instance?

Exactly. The FAQ is wrong. And yes... If you enable circular route and add this fix, that will work.

prusnak:
Does this only affect LND REST?

I use LND gRPC and don't see any issues with the self-send.

motorina0:
The initial code had invalid python syntax (`true` instead of `True`)
```py
json={"payment_request": bolt11, "fee_limit": lnrpc_fee_limit, "allow_self_payment": true},
```

I have add the option for `allow_self_payment` on the funding source configuration, so it can be set to both true or false (default false).

image

motorina0 (approved):

dni (approved):

prusnak:
> I have add the option for `allow_self_payment` on the funding source configuration, so it can be set to both true or false (default false).

Why would anyone want this to be set to false?

arcbtc:
> > I have add the option for `allow_self_payment` on the funding source configuration, so it can be set to both true or false (default false).
> 
> Why would anyone want this to be set to false?

So not to bother lnd with internal payments I guess

PR #2777, opened 2024-11-13
feat: add `pubkey` to the searchable fields

motorina0:


lnbits/core/crud/users.py
```diff
@@ -71,16 +73,23 @@ async def get_accounts(
         model=AccountOverview,
         group_by=["accounts.id"],
     )
+    for account in accounts.data:
+        account.refresh_privileges()
```
motorina0, line 77: # TODO: replace with pydating 2 'model_post_init'

talvasconcelos (approved):

arcbtc:
Would also be nice to nave a link to a user wallet so, if the link login is enabled in the install, the super user could click and paste into a different browser page to easily access the user.

arcbtc (approved):

PR #2796, opened 2024-12-09
fix: last_payment on userlist was broken

dni:


lnbits/static/js/users.js
```diff
@@ -173,7 +173,7 @@ window.app = Vue.createApp({
 
   methods: {
     formatDate: function (value) {
-      return LNbits.utils.formatDate(value)
```
motorina0, line 176: `LNbits.utils.formatDate` expects a date in seconds:

```js
    formatDate: function (timestamp) {
      return Quasar.date.formatDate(
        new Date(timestamp * 1000),
        'YYYY-MM-DD HH:mm'
      )
    },
```

We should create a `LNbits.utils.formatDateValue`:
```js
    formatDate: function (timestamp) {
      return Quasar.date.formatDate(
        new Date(timestamp),
        'YYYY-MM-DD HH:mm'
      )
    },
```

Having `'YYYY-MM-DD HH:mm'` all over the app is NOK.

motorina0 (approved):

talvasconcelos (approved):

PR #2819, opened 2024-12-12
fix: make startup extension check sync

dni:
wip


lnbits/app.py
```diff
@@ -395,16 +398,21 @@ def register_new_ratelimiter_fn():
...
-    """Register FastAPI routes for extension."""
+def register_ext_tasks(ext: Extension) -> None:
+    """Register extension async tasks."""
     ext_module = importlib.import_module(ext.module_name)
 
-    ext_route = getattr(ext_module, f"{ext.code}_ext")
-
     if hasattr(ext_module, f"{ext.code}_start"):
         ext_start_func = getattr(ext_module, f"{ext.code}_start")
         ext_start_func()
```
motorina0, line 407: we should also support `async xxx_start()`, just like for stop

motorina0 (approved):

PR #2823, opened 2024-12-12
feat: return the price with the fiat rate

motorina0:
Previously it only returned `rate`, now it also returns price.

```
http://localhost:5000/api/v1/rate/usd

{"rate":987.5047628982029,"price":101265.33436305918}

```


dni (approved):

dni:
could be a pydantic model for better api docs

PR #2827, opened 2024-12-13
fix: don't show when custom badge is not set

talvasconcelos:
Proper remove CUSTOM BADGE if it's no text is set

### DIFF
--- a/lnbits/templates/base.html
+++ b/lnbits/templates/base.html
@@ -67,10 +67,11 @@
             </q-badge>
             {%endif%} {% if LNBITS_CUSTOM_BADGE is defined %}
             <q-badge
+              v-if="`{{LNBITS_CUSTOM_BADGE}}`.length > 0"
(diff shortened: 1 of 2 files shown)


lnbits/templates/base.html
```diff
@@ -67,10 +67,11 @@
             </q-badge>
             {%endif%} {% if LNBITS_CUSTOM_BADGE is defined %}
             <q-badge
+              v-if="`{{LNBITS_CUSTOM_BADGE}}`.length > 0"
```
motorina0, line 70: isn't this what caused the original issue?

talvasconcelos: no. it was a jinja conditional, i think

i tested with a couple of special characters, `'`, `\`, `<` , `;`

all worked

motorina0 (approved):

motorina0:
image

PR #2841, opened 2024-12-17
fix: get_payments where clause did query properly on certing arguments

dni:
i discovered that when testing lndhub, failed payments where still shown is pending. since we introduced the failed state we need now check for success and pending `if complete and pending` and for the else case, `not complete and not pending` we should query for failed


motorina0:
This PR requires a unit test.

dni:
failing tests here https://github.com/lnbits/lnbits/pull/2844

talvasconcelos (approved):
Seems to work as expected!

Still waiting for one invoice to expire from testing!

motorina0 (approved):

PR #2817, opened 2024-12-11
feat: Wallet UI tweaks

arcbtc:
### This pr has been broken up into smaller prs
Balance card and fiat update: https://github.com/lnbits/lnbits/pull/2851
Background images and neon borders: https://github.com/lnbits/lnbits/pull/2850
Consistent page layouts: https://github.com/lnbits/lnbits/pull/2848
Cleaned transaction table https://github.com/lnbits/lnbits/pull/2859
Extra calls to action, for easier navigation https://github.com/lnbits/lnbits/pull/2860

---------------------

All thanks to some excellent design mockups and ideas from Ben Weeks! 🙌 🙌 🙌
Will have:
* Other wallet cards in the Wallet UI
* Configurable bg image option (from admin as global and user if not set in admin)
* Better exchange rate workflow
* Button cleanup

So far....
image
image

PR #2850, opened 2024-12-18
Add option to add a background image and neon borders

arcbtc:
Part of the break up of https://github.com/lnbits/lnbits/pull/2817

Background images are blurred so smaller images can be used and also as most images scaled suck.
People love customising, and I'm sure users will have a lot of fun with this OPTIONAL feature.
(if you nack, you nack fun, are anti-choice and hate liberty and expression)

Next: after merge I will add to settings so an admin can set as global, so an admin can force their users to use their garish theme.

image
image
image


BenGWeeks:
Ha ha, nice work. Look forward to seeing more.

BenGWeeks:
Would love to see some pre-built extensive themes, such as "Cypherpunk".

arcbtc:
> Would love to see some pre-built extensive themes, such as "Cypherpunk".


Nice idea, but it would get to messy. I cant wait to see what the people will make on there own

talvasconcelos (approved):

motorina0 (approved):
Hot!

Styling logic should be extracted in `ui-customisation.js`

PR #2851, opened 2024-12-18
Fixed fiat tracking

arcbtc:
Part of the break up of https://github.com/lnbits/lnbits/pull/2817

* Adds a fiat tracking as default for newly created wallets (because most people want that)
* Removes goofy page reloads
* Stores last exchange rate in local to avoid clunky loads
* Allows for fiat on top (because there are a bunch of people who want this feature (weird)
* Improves fiat tracking form
Extra:
* fixes the superuser credit/debit button

https://github.com/user-attachments/assets/0dfe4988-8b8d-45f0-a5a1-27db3aeb9352


lnbits/core/migrations.py
```diff
@@ -684,3 +684,22 @@ async def m029_create_audit_table(db: Connection):
...
     )
+
+
+async def m030_add_currency_to_wallet(db: Connection):
+    """
+    Setting currency to default is heavy, so easier just to add a trigger.
+    """
+    await db.execute(
+        """
+        CREATE TRIGGER IF NOT EXISTS set_default_currency
```
dni, line 695: i am not sure about a db trigger for default values, we do not handle defaults like that anywhere, why not setting the default on for example `create_wallet` function?

arcbtc: yeah thats a better idea

lnbits/static/js/wallet.js
```diff
@@ -665,27 +696,52 @@ window.app = Vue.createApp({
...
+    },
+    
   },
   mounted() {
     // show disclaimer
     if (!this.$q.localStorage.getItem('lnbits.disclaimerShown')) {
       this.disclaimerDialog.show = true
       this.$q.localStorage.set('lnbits.disclaimerShown', true)
     }
+    // check blanace priority
```
dni, line 736: typo

lnbits/templates/components.vue
```diff
@@ -506,7 +506,38 @@
...
+    <q-tooltip v-text="$t('credit_hint')"></q-tooltip>
+  </q-btn>
+
+  <q-btn
+    v-if="admin && !small_btn"
+    color="primary"
+    :label="$t('credit_debit')"
+    class="float-right q-mt-sm"
+    size="sm"
+  >
```
dni, line 540: can we look at this change of the update-balance component in another pr?
i think instead of adding a small_btn  property yyou could just add the `size=` on the html of the compoenent itself like `<update-balance size="3">...`

lnbits/static/js/wallet.js
```diff
@@ -665,27 +696,52 @@ window.app = Vue.createApp({
...
+      if (value == true) {
+        this.mobileSimple = false
+      }
+    },
     updatePayments() {
-      this.updateFiatBalance()
+      this.fetchBalance()
+    },
+    'update.currency'(newValue) {
+      if (this.ignoreWatcher || this.update.currency == '') return
```
dni, line 719: what is the watcher?

lnbits/static/js/components.js
```diff
@@ -484,7 +484,7 @@ window.app.component('lnbits-dynamic-chips', {
 window.app.component('lnbits-update-balance', {
   template: '#lnbits-update-balance',
   mixins: [window.windowMixin],
-  props: ['wallet_id', 'credit-value'],
+  props: ['wallet_id', 'credit-value', 'small_btn'],
```
dni, line 487: my comment below, i dont think it should be a property

lnbits/core/templates/core/wallet.html
```diff
@@ -133,35 +215,32 @@ <h3 class="q-my-none text-no-wrap">
...
-              >
+              <q-btn unelevated color="secondary" icon="qr_code_scanner">
                 <q-tooltip
                   ><span v-text="$t('camera_tooltip')"></span
                 ></q-tooltip>
               </q-btn>
+              <lnbits-update-balance
+                v-if="$q.screen.gt.md"
+                :wallet_id="this.g.wallet.id"
+                :callback="updateBalanceCallback"
```
dni, line 237: there is no `callback` anymore

PR #2857, opened 2024-12-19
ci: update extensions lock on release and pr to ext repo

dni:


motorina0:
This logic is done on the `lnbits-extensions` repo

PR #2864, opened 2024-12-20
[feat] access control lists (with access tokens)

motorina0:


lnbits/core/crud/users.py
```diff
@@ -185,3 +186,20 @@ async def get_user_from_account(
...
+async def get_user_access_control_lists(
+    user_id: str, conn: Optional[Connection] = None
+) -> UserACLs:
+    user_acls = await (conn or db).fetchone(
+        "SELECT id, access_control_list FROM accounts WHERE id = :id",
+        {"id": user_id},
+        UserACLs,
+    )
+
+    return user_acls if user_acls else UserACLs(id=user_id)
```
dni, line 205:
```suggestion
    return user_acls or UserACLs(id=user_id)
```

talvasconcelos (approved):
Tested.

Could use some hints on how to use but, works as expected

tests/copilot_prompt.md
```diff
@@ -0,0 +1,48 @@
...
+
+- use register endpoint to obtain the access token (see example in test_register_ok)
+- the test should only check that the ACLs are sorted alphabeticaly by name
+
+_Sample 1_
+@workspace /tests Develop a test for the selected code (only the function check_user_exists in decorators.py file).
+Requirements:
+
+- use register endpoint to obtain the access token (see example in the file test_auth.py the function test_register_ok())
+- the test should register a new user, optain the access token then delete the user. Then check that check_user_exists() fails as expected
```
dni, line 43:
```suggestion
- the test should register a new user, obtain the access token then delete the user. Then check that check_user_exists() fails as expected
```

lnbits/decorators.py
```diff
@@ -288,11 +287,48 @@ async def _get_account_from_token(access_token) -> Optional[Account]:
...
+    account = None
+    if payload.sub is not None:
+        account = await get_account_by_username(payload.sub)
+    if payload.usr is not None:
+        account = await get_account(payload.usr)
+    if payload.email is not None:
+        account = await get_account_by_email(payload.email)
+
+    if not account:
+        return None
```
dni, line 302: could be an else if, with else return None

lnbits/core/models/__init__.py
```diff
@@ -73,6 +74,7 @@
     "Account",
     "AccountFilters",
     "AccountOverview",
+    "UserACLs",
```
dni, line 77: maybe, `UserAcls` for keeping consistent with upper camelcase?

lnbits/core/models/misc.py
```diff
@@ -38,6 +38,11 @@ class SimpleStatus(BaseModel):
     message: str
 
 
+class SimpleItem(BaseModel):
```
dni, line 41: its very generic name and hard to know where it belongs, can we maybe do something like `AclTokenListItem`

dni (approved):

arcbtc (approved):

lnbits/core/models/__init__.py
motorina0: done

lnbits/core/models/misc.py
motorina0: Its generic indeed and located in `models/misc.py`.
It can be used in different context.

PR #2867, opened 2024-12-23
Removes reloads

arcbtc:
Because reloads are bad practice/ux for a webapp

### DIFF
--- a/lnbits/static/js/base.js
+++ b/lnbits/static/js/base.js
@@ -8,7 +8,7 @@ window.i18n = new VueI18n.createI18n({
 
 const websocketPrefix =
   window.location.protocol === 'http:' ? 'ws://' : 'wss://'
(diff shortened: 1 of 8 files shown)


lnbits/static/js/base.js
```diff
@@ -8,7 +8,7 @@ window.i18n = new VueI18n.createI18n({
 
 const websocketPrefix =
   window.location.protocol === 'http:' ? 'ws://' : 'wss://'
-const websocketUrl = `${websocketPrefix}${window.location.host}/api/v1/ws`
+const websocketUrl = `${'http:' ? 'ws://' : 'wss://'}${window.location.host}/api/v1/ws`
```
remcoros, line 11: I think this isn't correct? this will always resolve to 'ws://' even for https.

PR #2872, opened 2025-01-06
v1 feat: Vue Routes

arcbtc:
Vue routes working alongside FastAPI. No more horrid page reloads in core.

https://github.com/user-attachments/assets/20764848-5009-4f92-8c88-f5ff20484266

Currently adding a page is as easy as this `/admin` example (and will get easier):

```
 <q-btn to="/admin">Admin</q-btn>

#### static/js/admin.js #####

 window.AdminPageLogic = {
  mixins: [windowMixin],
  data() {
    return {
      settings: {},
      logs: [],
      etc....

#### admin/index.html ( remove js, as route loads) #####
 
 {% if not ajax %} {% extends "base.html" %} {% endif %} {% from "macros.jinja"
import window_vars with context %} 
{% block scripts %} {{ window_vars(user) }}{% endblock %} 

#### generic.py (to just fetch without extending base.html) #####

is_ajax = request.headers.get("X-Requested-With") == "XMLHttpRequest"
return template_renderer().TemplateResponse(
     request,
     "admin/index.html",
     {
         "user": user.json(),
         "settings": settings.dict(),
         "balance": balance,
         "currencies": list(currencies.keys()),
         "ajax": is_ajax,
     },
)
    
##### static/js/init-app.js #####

 {
    path: '/admin',
    name: 'Admin',
    component: DynamicComponent,
    props: {
      fetchUrl: '/admin',
      scripts: ['/static/js/admin.js']
    }
 },
  
  ```

### CODE REVIEW

File: lnbits/templates/base.html [hunk 1]
```diff
@@ -244,33 +255,43 @@
...
+            window.LOCALE = 'en'
+      window.dateFormat = 'YYYY-MM-DD HH:mm'
+      window.i18n = new VueI18n.createI18n({
+        locale: window.LOCALE,
+        fallbackLocale: window.LOCALE,
+        messages: window.localisation
+      })
+      const websocketPrefix =
+        window.location.protocol === 'http:' ? 'ws://' : 'wss://'
+      const websocketUrl = `${'http:' ? 'ws://' : 'wss://'}${window.location.host}/api/v1/ws`
```
remcoros, line 294: this if check is always true and will always use 'ws://' even for https

motorina0 (approved):

remcoros:
@motorina0 Did you see my comment? The ws:/wss: check in base.html is not ok and will not work when using https

PR #2873, opened 2025-01-07
feat: add python-multipart for uploads

dni:
that is currently only used by an extension but we could easily add fileuploads to core if needed


motorina0:
has conflict

motorina0:
> that is currently only used by an extension but we could easily add fileuploads to core if needed

what extension?

PR #2875, opened 2025-01-08
fix: pay_invoice timeout to prevent blocking

dni:
### Handle long running payments
When making a payment the funding source can take a long time to respond. Either because it is performing path finding or its a hodl invoice, or any other reason.

The current PR runs the `pay_invoice()` function in a background task, thus avoiding the server from slowing down.
The timeout can be configured in the admin page.
image

The pending invoices are checked at regular intervals or can be checked manually:
image


motorina0 (approved):

talvasconcelos (approved):

lnbits/core/services/payments.py
```diff
@@ -491,13 +493,18 @@ async def _pay_external_invoice(
...
-    payment_response: PaymentResponse = await funding_source.pay_invoice(
-        create_payment_model.bolt11, fee_reserve_msat
+    task = create_task(
+        _fundingsource_pay_invoice(checking_id, payment.bolt11, fee_reserve_msat)
     )
-    logger.debug(f"backend: pay_invoice finished {checking_id}, {payment_response}")
+
+    # make sure a hold invoice or deferred payment is not blocking the server
+    try:
+        payment_response = await asyncio.wait_for(task, 5)
```
motorina0, line 502: Magic number.

motorina0: hint: hodl invoice, bolz, etc

arcbtc (approved):

PR #2879, opened 2025-01-09
chore: generate new lockfile

dni:


prusnak:
Are you sure you want to bump all dependencies to latest?

Or did you just want to do `lock --no-update` and reflect the Python min version change?

dni:
if its possible i would like to have all bumped for a new year cleanup :)

prusnak:
needs rebase

Issue #2883, opened 2025-01-10
[BUG] NWC boot loop stuck if connection to backend fail


ElucGeek:
When NWC is set as a source, it might successfully connect to the getalby relay, but then could fail to connect the node where Alby Hub is installed behind. In this case it's stuck in a loop waiting for connection but never fail back to VoidWallet as it already consider that the funding is successful.

`2025-01-10 22:15:50.56 | INFO | Funding source: NWCWallet
2025-01-10 22:15:50.56 | INFO | Data folder: ./data
2025-01-10 22:15:50.56 | INFO | Database: PostgreSQL
2025-01-10 22:15:50.56 | INFO | Service fee: 0.0
2025-01-10 22:15:50.56 | INFO | Service fee max: 0
2025-01-10 22:15:50.56 | INFO | Service fee wallet: None
2025-01-10 22:15:50.56 | INFO | NWCConnection is ready. relay: wss://relay.getalby.com/v1 account: XXXXX service: xxxxx
2025-01-10 22:15:50.56 | INFO | Connecting to backend NWCWallet...
2025-01-10 22:15:50.56 | DEBUG | Waiting for connection...
2025-01-10 22:15:50.56 | DEBUG | Connecting to NWC relay wss://relay.getalby.com/v1
2025-01-10 22:15:50.56 | DEBUG | Creating new connection...
2025-01-10 22:15:51.56 | DEBUG | Waiting for connection...
2025-01-10 22:15:52.56 | DEBUG | Waiting for connection...
2025-01-10 22:15:53.56 | DEBUG | Waiting for connection...
2025-01-10 22:15:54.57 | DEBUG | Waiting for connection...
2025-01-10 22:15:55.57 | DEBUG | Waiting for connection...
2025-01-10 22:15:56.57 | DEBUG | Waiting for connection...
2025-01-10 22:15:57.57 | DEBUG | Waiting for connection...
2025-01-10 22:15:58.57 | DEBUG | Waiting for connection...
2025-01-10 22:15:59.57 | DEBUG | Waiting for connection...
2025-01-10 22:16:00.56 | ERROR | Error connecting to NWC relay: 
2025-01-10 22:16:00.56 | DEBUG | Reconnecting to NWC relay in 5 seconds...
2025-01-10 22:16:00.57 | DEBUG | Waiting for connection...
`

It's then stuck in this loop until the service is force stop.

[...]

PR #2885, opened 2025-01-13
fix: Add timeout to wait_for_connection

riccardobl:
Gives up connecting to the nwc service if it doesn't respond for 2 minutes.

Closes https://github.com/lnbits/lnbits/issues/2883 .

### DIFF
--- a/lnbits/wallets/nwc.py
+++ b/lnbits/wallets/nwc.py
@@ -440,11 +440,14 @@ class NWCConnection:
                         logger.error("Error closing subscription: " + str(e))
         return subscription
 
-    async def _wait_for_connection(self):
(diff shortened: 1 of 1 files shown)


motorina0 (approved):

lnbits/wallets/nwc.py
```diff
@@ -440,11 +440,14 @@ async def _close_subscription_by_eventid(
                         logger.error("Error closing subscription: " + str(e))
         return subscription
 
-    async def _wait_for_connection(self):
+    async def _wait_for_connection(self, timeout=60 * 2):
```
dni, line 443:
```suggestion
    async def _wait_for_connection(self, timeout: int = 60 * 2):
```

PR #2898, opened 2025-01-18
feature: Default theme for new users

arcbtc:
Super user can set color, bg image, border, and a default reaction.
User theme settings override the global theme settings.

https://github.com/user-attachments/assets/006966de-be9d-4782-a0b4-27f209e302b0


talvasconcelos (approved):
tested

motorina0 (approved):

PR #2900, opened 2025-01-20
Translate expiry to expirySeconds for Phoenixd

blackcoffeexbt:
Translate expiry to expirySeconds for Phoenixd and add a default of 1 hour if not set


blackcoffeexbt:
Gah. Didn't rebase. Closing. Will reopen

PR #2903, opened 2025-01-21
change UI for create invoice dialog

talvasconcelos:
Change the create invoice dialog

image

image

image

image

If wallet has a tracking fiat currency, when creating the currency now defaults to that tracking currency


arcbtc:
This looks quite heavy. What about something like this, to avoid adding all the custom css?
https://github.com/lnbits/lnbits/pull/2913

PR #2914, opened 2025-01-28
fix: router force update

talvasconcelos:
The loading of the routes dynamically is using `$forceUpdate` every time the component is loaded. After the first render, if the user goes to an already loaded component (for example open server, then open users, open server again) it makes a weird re render, caused by the forceUpdate, regardless of the component was already loaded!

This fix, adds an array that stores already loaded components and prevents the forceUpdate if it already been visited/loaded before, preventing the re render and weird behaviour!!


motorina0:
Lint fails

arcbtc (approved):

motorina0 (approved):

PR #2917, opened 2025-01-28
Feat: Adds wallet icon/color select

arcbtc:


lnbits/core/migrations.py
```diff
@@ -692,3 +692,14 @@ async def m030_add_user_api_tokens_column(db: Connection):
...
         """
     )
+
+
+async def m031_add_color_and_icon_to_wallets(db: Connection):
+    """
+    Adds icon and color columns to wallets.
+    """
+    try:
+        await db.execute("ALTER TABLE wallets ADD COLUMN icon TEXT DEFAULT 'flash_on'")
```
motorina0, line 702: - `wallets` table should have an `extra` column where we put these additional fields (and future ones)
- see `User` class:
```py
class User(BaseModel):
    id: str
    created_at: datetime
    ...
```

lnbits/core/models/wallets.py
```diff
@@ -34,6 +34,8 @@ class Wallet(BaseModel):
     updated_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
     currency: Optional[str] = None
     balance_msat: int = Field(default=0, no_database=True)
+    icon: str = "flash_on"
```
motorina0, line 37: should go into `extra`, see: https://github.com/lnbits/lnbits/pull/2917/files#r1933635201

lnbits/core/migrations.py
arcbtc: Why, are they not just two different approaches with similar weight pros/cons? As by same logic something currency would go there as well?

lnbits/core/models/wallets.py
arcbtc: Icon could one day be an image blob, so not sure a json makes sense

motorina0 (approved):

PR #2949, opened 2025-02-10
feat: modernize error page

talvasconcelos:
Hopefully a better and cleaner error page

image

image


talvasconcelos:
@motorina0 , check only works for installed extensions!

lnbits/templates/error.html
```diff
@@ -23,7 +23,7 @@
             class="q-mb-lg full-width"
           ></q-btn>
           <br />
-          <q-btn color="primary" to="/" label="Go Home"></q-btn>
+          <q-btn color="primary" href="/" label="Go Home"></q-btn>
```
talvasconcelos, line 26: the `to` is a vue router thing, hfer also works

lnbits/core/views/generic.py
```diff
@@ -50,12 +50,9 @@ async def home(request: Request, lightning: str = ""):
...
 async def first_install(request: Request):
     if not settings.first_install:
-        return template_renderer().TemplateResponse(
-            request,
-            "error.html",
-            {
-                "err": "Super user account has already been configured.",
-            },
+        raise HTTPException(
+            status_code=HTTPStatus.SERVICE_UNAVAILABLE,
```
dni, line 54: Service Unvailable is the wrong statuscode. it should only be used if the webserver is not reachable.
maybe `bad request`?

lnbits/exceptions.py
```diff
@@ -119,3 +130,43 @@ async def invoice_error_handler(request: Request, exc: InvoiceError):
...
+            status_code = HTTPStatus.FORBIDDEN
+            message = f"Extension '{path}' not installed. Ask the admin to install it."
+
+        return template_renderer().TemplateResponse(
+            request,
+            "error.html",
+            {"status_code": int(status_code), "message": message},
+            int(status_code),
+        )
+
```
dni, line 158: why are we returning forbidden on an 404 exception handler?

lnbits/core/tasks.py
```diff
@@ -62,6 +63,13 @@ async def run_by_the_minute_tasks():
             except Exception as ex:
                 logger.error(ex)
 
+        if minute_counter % 60 == 0:
+            try:
+                # initialize the list of all extensions
+                await InstallableExtension.get_installable_extensions()
+            except Exception as ex:
+                logger.error(ex)
+
```
dni, line 72: why do we do this every hour?

lnbits/exceptions.py
talvasconcelos: I'm guessing because the path isn't allowed for the user, something along that logic! @motorina0 ?

that is specifically for the extensions, so the error is 404 but since it's related to an extension.... bla bla bla.... should i remove it and keep it as 404?

dni: i think 404 here is also fine

arcbtc (approved):

PR #2982, opened 2025-02-21
bug: settings fundingtab errors

dni:
funding tab was not working if you had node ui enabled


talvasconcelos (approved):
that was probably me...

PR #2999, opened 2025-02-26
fix: paying an invoice does not send a webhook. #2472

### LINKED ISSUE #2472: Paying an invoice does not send a webhook.
When paying an invoice a webhook is not sent. But websockets updates are working.
I checked the source and on the [payinvoice function](https://github.com/lnbits/lnbits/blob/c04c13b2f8cfbb625571a07dfddeb65ea6df8dac/lnbits/core/services.py#L183C11-L183C22) on line [319](https://github.com/lnbits/lnbits/blob/c04c13b2f8cfbb625571a07dfddeb65ea6df8dac/lnbits/core/services.py#L319) and line [363](https://github.com/lnbits/lnbits/blob/c04c13b2f8cfbb625571a07dfddeb65ea6df8dac/lnbits/core/services.py#L363) invoices notifications with websockets are being sent. 

Should we send also the webhook with the [dispatch_webhook](https://github.com/lnbits/lnbits/blob/c04c13b2f8cfbb625571a07dfddeb65ea6df8dac/lnbits/core/tasks.py#L111) function?

dni:
closes #2472

`send_payment_notification` now does all the actions, send ws, send push notifications, send webhook, send chat


lnbits/core/services/payments.py
```diff
@@ -288,6 +289,13 @@ async def send_payment_notification(wallet: Wallet, payment: Payment):
         send_chat_payment_notification(wallet, payment)
     except Exception as e:
         logger.error("Error sending chat payment notification", e)
+    try:
+        await send_payment_push_notification(payment)
```
motorina0, line 293: should take `wallet` as an optional parameter

dni: fixed

motorina0 (approved):

PR #3002, opened 2025-02-26
[fix]: various issues discovered during testing

motorina0:
See comments next to code.


lnbits/middleware.py
```diff
@@ -134,8 +134,9 @@ async def dispatch(self, request: Request, call_next) -> Response:
             assert response
             return response
         finally:
-            duration = (datetime.now(timezone.utc) - start_time).total_seconds()
-            await self._log_audit(request, response, duration, request_details)
+            if request_details:
```
motorina0, line 137: only if audit enabled

lnbits/settings.py
```diff
@@ -695,12 +695,14 @@ def _is_http_request_path_auditable(self, path: Optional[str]):
                 if _re_fullmatch_safe(exclude_path, path):
                     return False
 
-        if len(self.lnbits_audit_include_paths) != 0:
-            if not path:
-                return False
-            for include_path in self.lnbits_audit_include_paths:
-                if _re_fullmatch_safe(include_path, path):
-                    return True
+        if len(self.lnbits_audit_include_paths) == 0:
```
motorina0, line 698: empty path means all paths

lnbits/static/js/base.js
```diff
@@ -357,6 +357,7 @@ window.LNbits = {
       }
     },
     prepareFilterQuery(tableConfig, props) {
+      tableConfig.filter = tableConfig.filter || {}
```
motorina0, line 360: defensive for when a table config is not properly initialized

lnbits/utils/exchange_rates.py
```diff
@@ -187,6 +187,9 @@ def allowed_currencies():
 
 
 async def btc_rates(currency: str) -> list[tuple[str, float]]:
+    if currency.upper() not in allowed_currencies():
+        raise ValueError(f"Currency '{currency}' not allowed.")
+
```
motorina0, line 192: prevent calls to exchanges APIs

PR #3006, opened 2025-02-27
fix: python versions were defined incorrectly and remove 3.9

dni:
python3.13 does fail on building greenlet.

https://python-poetry.org/docs/1.8/dependency-specification/#tilde-requirements

screenshot-1740649150

- removed python3.9 (had issues aswell)
- now poetry will throw errors when installing on wrong versions
- RUF007 formatting now works without python3.9


motorina0 (approved):

tlindi:
I'm not exactly anygood with pyenv etc "things". But FYI: 

I tried to update my python to 3.12 and LNbits make fails with error described here:
https://github.com/python/cpython/issues/124241

PR #3010, opened 2025-02-27
Fix: don't show register if new users not allowed

talvasconcelos:
Enforce the `LNBITS_NEW_ACCOUNTS_ALLOWED` flag/setting on the UI. 

Don't show the register form.

image


lnbits/core/templates/core/index.html
```diff
@@ -160,15 +160,19 @@
                     class="text-center text-grey-6"
                     v-if="authAction !== 'reset'"
                   >
-                    <p v-if="authAction === 'login'" class="q-mb-none">
+                    <p
+                      v-if="authAction === 'login' && {{ LNBITS_NEW_ACCOUNTS_ALLOWED }}"
```
motorina0, line 164: wouldn't this be true when `LNBITS_NEW_ACCOUNTS_ALLOWED` is `"False"`?

on the same file this is used:
 :allowed_new_users="{{ LNBITS_NEW_ACCOUNTS_ALLOWED | tojson }}"

talvasconcelos: good one

motorina0 (approved):

lnbits/templates/components.vue
```diff
@@ -1302,6 +1302,9 @@
         </div>
       </q-form>
     </q-card-section>
+    <q-card-section v-else-if="!allowed_new_users && authAction === 'register'">
+      <h5 class="text-center" v-text="$t('new_user_not_allowed')"></h5>
+    </q-card-section>
```
dni, line 1307: can we just hide the register button and don't even show the `new_user_not_allowed` card section. i think it would be cleaner.

lnbits/core/templates/core/index.html
```diff
@@ -160,15 +160,19 @@
                     class="text-center text-grey-6"
                     v-if="authAction !== 'reset'"
                   >
-                    <p v-if="authAction === 'login'" class="q-mb-none">
+                    <p
+                      v-if="authAction === 'login' && {{ LNBITS_NEW_ACCOUNTS_ALLOWED | tojson }}"
```
dni, line 164: can you introduce a this? 
```suggestion
                      v-if="authAction === 'login' && allowed_new_users"
```

lnbits/templates/components.vue
talvasconcelos: ok, now by default we don't show the register card, if register is not allowed. I kept it in the markup, just in case someone, somehow ends up there!

motorina0 (approved):

dni (approved):

dni (approved):

PR #3013, opened 2025-03-03
[fix] settings cleanup

motorina0:
- add validation for `int` and `float` settings
- remove SaaS settings and logic (use `first_install`) instead
- default to stronger default `auth_secret`
- create temporary `auth_secret` and store it in a file

### DIFF
--- a/lnbits/settings.py
+++ b/lnbits/settings.py
@@ -11,6 +11,7 @@ from os import path
 from pathlib import Path
 from time import gmtime, strftime, time
 from typing import Any
+from uuid import uuid4
 
 from loguru import logger
 from pydantic import BaseModel, BaseSettings, Extra, Field, validator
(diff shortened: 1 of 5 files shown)


talvasconcelos (approved):

arcbtc (approved):

lnbits/settings.py
```diff
@@ -792,24 +795,25 @@ class EnvSettings(LNbitsSettings):
...
-    def lnbits_server_up_time(self) -> str:
-        up_time = int(time() - self.server_startup_time)
-        return strftime("%H:%M:%S", gmtime(up_time))
-
-
-class SaaSSettings(LNbitsSettings):
-    lnbits_saas_callback: str | None = Field(default=None)
-    lnbits_saas_secret: str | None = Field(default=None)
-    lnbits_saas_instance_id: str | None = Field(default=None)
+    def check_auth_secret_key(self):
```
dni, line 806:
```suggestion
    def check_auth_secret_key(self) -> None:
```

PR #3014, opened 2025-03-03
refactor: render variables for jinja, components and window vars

dni:
now you can use all the helpers constants in the vue context via the windowMixin and all the properties are parsed consistently.


lnbits/helpers.py
```diff
@@ -65,59 +65,45 @@ def template_renderer(additional_folders: Optional[list] = None) -> Jinja2Templa
     t = Jinja2Templates(loader=jinja2.FileSystemLoader(folders))
     t.env.globals["static_url_for"] = static_url_for
 
-    if settings.lnbits_ad_space_enabled:
```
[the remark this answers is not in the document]
dni: because we need it in the templating anyway

talvasconcelos:
I like it

talvasconcelos (changes requested):
I like the concept but it's not working, needs further debugging

talvasconcelos:
Tested, it works and fixes the print pages!

talvasconcelos (approved):

lnbits/templates/components.vue
```diff
@@ -1443,7 +1445,7 @@
         </q-tooltip>
       </q-btn>
       <q-btn
-        v-if="authMethods.includes('keycloak-auth')"
+        v-if="LNBITS_AUTH_METHODS.includes('keycloak-auth')"
```
motorina0, line 1448: why mix globals here?

dni: i think i removed the property authMethods, now parsing is done globally, and before that it was done inside the component

motorina0: The question is "why"?
It was better before.

dni: why passing around properties when they are static in globals?

motorina0: Not in scope of this PR

dni: reverted all refactors from components

lnbits/static/js/index.js
```diff
@@ -101,8 +101,8 @@ window.app = Vue.createApp({
     }
   },
   created() {
-    this.description = SITE_DESCRIPTION
-    this.allowedRegister = ALLOWED_REGISTER
+    this.description = this.SITE_DESCRIPTION
```
motorina0, line 104: is `this.SITE_DESCRIPTION` correct (using `this`)?

dni: this uses windowMixing instead of the global, i think yes, this is cleaner

lnbits/templates/base.html
```diff
@@ -390,42 +390,38 @@ <h6 class="q-my-none ellipsis full-width">
...
     </div>
 
+    <script type="text/javascript">
+      const WINDOW_SETTINGS = {{ WINDOW_SETTINGS | tojson }}
+      Object.keys(WINDOW_SETTINGS).forEach(key => {
+        window[key] = WINDOW_SETTINGS[key]
+      })
+      if (LNBITS_THEME_OPTIONS && LNBITS_THEME_OPTIONS.length) {
+        window.allowedThemes = LNBITS_THEME_OPTIONS.map(str => str.trim())
+      }
```
motorina0, line 400: can this be in helper?

PR #3018, opened 2025-03-04
feat: add test email and improve on email notifications

dni:
- added email_username (can be different than email)
- send test email function / api endpoint
- move send_mail to helper

screenshot-1741086412


lnbits/core/services/notifications.py
```diff
@@ -111,41 +108,26 @@ async def send_telegram_message(token: str, chat_id: str, message: str) -> dict:
...
+            settings.lnbits_email_notifications_server,
+            settings.lnbits_email_notifications_port,
+            settings.lnbits_email_notifications_username,
+            settings.lnbits_email_notifications_password,
+            settings.lnbits_email_notifications_email,
+            settings.lnbits_email_notifications_to_emails,
+            subject,
+            message,
+        )
+        return {"status": "ok"}
```
motorina0, line 127: use:
```py
class SimpleStatus(BaseModel):
    success: bool
    message: str
```

lnbits/helpers.py
```diff
@@ -348,3 +351,33 @@ def normalize_path(path: Optional[str]) -> str:
 
 def normalized_path(request: Request) -> str:
     return "/" + "/".join(path_segments(request.url.path))
+
+
+async def send_email(
```
motorina0, line 356: this is a notification service, not a helper

dni: i would argue this is a smtp helper, you see it by the imports

arcbtc: Id argue the same. I will use for extensions at some point.

arcbtc (approved):

lnbits/helpers.py
arcbtc: Nostr would be helpful as a helper as well, give it a privkey, pubkey and message.

motorina0: A helper something you call from crud, models, services, views or api.
A service is something you call from other services or from api.
`send_email` IS a service.

motorina0 (approved):

PR #3029, opened 2025-03-10
tweak: fixes service fee badge

arcbtc:


dni (changes requested):
you should add denomination to the i18n var `service_fee_max`

lnbits/static/i18n/en.js
```diff
@@ -166,6 +166,7 @@ window.localisation.en = {
   payment_hash: 'Payment Hash',
   fee: 'Fee',
   amount: 'Amount',
+  max: 'Max',
```
dni, line 169: `service_fee_max` add another var with denom to there, its cleaner

talvasconcelos (approved):

dni (approved):

dni:
the issue was that the label of the badge was duplicated

PR #3031, opened 2025-03-10
tweak: use qr code image for favicon as well

arcbtc:
Screenshot from 2025-03-10 21-56-56


dni:
i dont like this, i think favicon and qrcode icon should be different.

dni (approved):
screenshot-1741775992

talvasconcelos (approved):

PR #3049, opened 2025-03-13
feat: denom in the service fee badge, makes no sense

arcbtc:
Its always sats or always some fakewallet denom.
Neither user base needs to be aware

### DIFF
--- a/lnbits/templates/base.html
+++ b/lnbits/templates/base.html
@@ -89,7 +89,7 @@
             >
               {% if LNBITS_SERVICE_FEE_MAX > 0 %}
               <span
-                v-text='$t("service_fee_max_badge", { amount: "{{ LNBITS_SERVICE_FEE }}", max: "{{ LNBITS_SERVICE_FEE_MAX }}", denom: "{{ LNBITS_DENOMINATION }}"})'
+                v-text='$t("service_fee_max_badge", { amount: "{{ LNBITS_SERVICE_FEE }}", max: "{{ LNBITS_SERVICE_FEE_MAX }} {{ LNBITS_DENOMINATION }}"})'
               ></span>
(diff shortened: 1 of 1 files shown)


lnbits/templates/base.html
```diff
@@ -89,7 +89,7 @@
             >
               {% if LNBITS_SERVICE_FEE_MAX > 0 %}
               <span
-                v-text='$t("service_fee_max_badge", { amount: "{{ LNBITS_SERVICE_FEE }}", max: "{{ LNBITS_SERVICE_FEE_MAX }}", denom: "{{ LNBITS_DENOMINATION }}"})'
+                v-text='$t("service_fee_max_badge", { amount: "{{ LNBITS_SERVICE_FEE }}", max: "{{ LNBITS_SERVICE_FEE_MAX }} {{ LNBITS_DENOMINATION }}"})'
```
dni, line 92: i dont understand this, change its equivalent, if you combine the 2 variable into 1 replacement for the i18n string. you have to remove it from the i18n labels, but why?

arcbtc:
image

No reason to add "denom:" here.

Neither userbase fakewallet/sats users (userbase = users of the install beyond admin) need to know denoms are chnageable.

dni:
your change has the `LNBITS_DENOMINATION` still included, did you mean to remove it?

motorina0:
> No reason to add "denom:" here.

in `en.js`:
```json
  "service_fee_max_badge":  "Service fee: {amount} % per transaction (max {max} {denom})"
```

`service_fee_max_badge` expects an `denom` key. So:
 - either remove the `denom` placeholder from `service_fee_max_badge` (in all i18n files)
 - or leave the codebase as is

PR #3053, opened 2025-03-17
hide 'Runs On' scroll if title is not LNbits

talvasconcelos:
Don't show the runs on scroll if the title is different than `LNbits` and denomination is different than `sats`


motorina0 (approved):
LGTM

PR #3055, opened 2025-03-17
[Fix] default to sats if wallet not fakewallet

talvasconcelos:
denomination defaults to sats if the wallet is not fakewallet

To test:
- change fakewallet funding denomination to XXX
- save and refresh
- denomination should be XXX
- change funding source to whatever
- denomination should now default to sats
- change back to fakewallet, XXX should be the denomination


motorina0 (approved):
LGTM

dni (approved):

Issue #2274, opened 2024-02-13
[BUG] after delete-settings lost admin user name cant be reused


tlindi:
**Describe the bug**
lnbits-cli tool resets settings so they are re-read from .env and first_start is shown.
If one wants to use same super admin username, it is impossible cause username is reserved.

**To Reproduce**

1. Install lnbits and goto first_install. create superuser "lnbits" and password
2. run poetry run -C /home/setup/lnbits lnbits-cli delete-settings
3. restart LNbits
4. goto first_install page and try to create user lnbits
5. You will get error: Username already in exists. 403 FORBIDDEN
If you have lost superuser UID / username&pwd there's no avail.
Even if you create new superadmin, it cant see old super admin with usermanager. And it can not see users the old superuser created.

**Expected behavior**
"delete settings" should set superuser name and password so it is possible to first_install to reset username (not "usr") and password.

**Screenshots**
If needed.

**Desktop (please complete the following information):**

- LNbits version: 0.12.1
- Database: sqlite (this template is missings ":" after "Database")

**Additional context**
(User management is tuff issue to handle.)

dni:
you could just set a new username, login into the old one and change it there then again in the new one... but i agree its not ideal.

tlindi:
> you could just set a new username, login into the old one and change it there then again in the new one... but i agree its not ideal.

I could - but new superuser cant see old superuser on usermanager. And say I did reset settings cause I lost superuser username / password .. this seems not to work at all. I need to use cli to see superuser "usr".

And even with it I can't reset password cause MyAccount can not be accessed with http://IPaddress and there is no existing link to MyAccount page for legacy usr= "authentication".

tlindi:
Would it be good solution on first_install to be able to recover existing superuser by authenticating with superuser usr id?

motorina0:
> Even if you create new superadmin, it cant see old super admin with usermanager. And it can not see users the old superuser created.

This is now possible with the `/users` console. See https://github.com/lnbits/lnbits/pull/2139

Now the new `superuser` can 
 - manage the wallets of the old `superuser` (delete wallet or transfer funds)
 - delete old `superuser` (if it has no wallets)
 - recycle the old `superuser` (not necessarily a best practice)

Closing now. Please re-open the issue if the new `user` feature are not sufficient.

tlindi:
SuperUser "recovery" URL seems to work as expected on 1.0.0-rc9.
[...]

PR #3065, opened 2025-03-22
feat: dont reset superuser on delete settings endpoint

arcbtc:
✅ Deletes settings and not reset super_user
✅ Deletes localstorage settings
✅ Already resarts, so removes need to restart.
Unsure if this is also used to reset super_user somewhere?

### DIFF
--- a/lnbits/core/crud/settings.py
+++ b/lnbits/core/crud/settings.py
@@ -61,7 +61,7 @@ async def update_super_user(super_user: str) -> SuperSettings:
 
 
 async def delete_admin_settings(tag: Optional[str] = "core") -> None:
-    await db.execute("DELETE FROM settings WHERE tag = :tag", {"tag": tag})
+    await db.execute("DELETE FROM system_settings WHERE tag = :tag", {"tag": tag})
 
 
 async def create_admin_settings(super_user: str, new_settings: dict) -> SuperSettings:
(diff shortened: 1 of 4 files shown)


tlindi:
> Unsure if this is also used to reset super_user somewhere?
>
Was used here: https://github.com/lnbits/lnbits/issues/2274
And now no recovery is not possible: https://github.com/lnbits/lnbits/issues/3096

lnbits/core/crud/settings.py
```diff
@@ -61,7 +61,21 @@ async def update_super_user(super_user: str) -> SuperSettings:
...
 
 async def delete_admin_settings(tag: Optional[str] = "core") -> None:
-    await db.execute("DELETE FROM settings WHERE tag = :tag", {"tag": tag})
+    await db.execute(
+        """
+        DELETE FROM system_settings
+        WHERE tag = :tag
+        AND (
+            tag != 'core'
+            OR id NOT IN (
```
dni, line 70: don't put logic in crud, just delete the settings and update the newly created settings with the old user

lnbits/static/js/admin.js
```diff
@@ -546,10 +546,10 @@ window.AdminPageLogic = {
               Quasar.Notify.create({
                 type: 'positive',
                 message:
-                  'Success! Restored settings to defaults, restart required!',
+                  'Success! Restored settings to defaults. Restarting...',
                 icon: null
               })
-              this.needsRestart = true
+              this.$q.localStorage.clear()
```
dni, line 552: `needsRestart` marker was deleted

motorina0 (approved):

dni (approved):

PR #3071, opened 2025-03-29
[CHORE] add responsive text scaling with clamp() for wallet balance display

PatMulligan:
Implemented font size scaling using CSS clamp() function to make wallet balance
text responsive to viewport width. The balance now properly fills available
space while shrinking when constrained by smaller viewports, maintaining readability across all device sizes without text overflow issues.

- Min size: 0.75rem (12px) prevents text from becoming too small on mobile
- Scaling: 10vw provides smooth resizing relative to viewport width
- Max size: 3rem (48px) caps growth on larger screens

### Notes

- you may want to play with the values of the clamp function [mdn web docs clamp() fct](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp)

### Before
- mobile
image
- desktop
image

### After
- mobile
image
- desktop
image

### DIFF
--- a/lnbits/core/templates/core/wallet.html
+++ b/lnbits/core/templates/core/wallet.html
@@ -59,9 +59,11 @@
                   <div class="col-7">
                     <div class="row">
                       <div class="col-auto">
-                        <div class="text-h3 q-my-none text-no-wrap">
+                        <div class="text-h3 q-my-none full-width">
                           <strong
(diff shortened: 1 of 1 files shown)


motorina0 (approved):

talvasconcelos (approved):
I like it.
There's no Quasar class to do this, @arcbtc ! Let's merge it

PR #3090, opened 2025-04-04
[FIX] Fix database compatibility issue in search query builder for payments filter

PatMulligan:
## Description
This PR resolves an issue with the search functionality that was failing in SQLite environments due to the use of the `concat()` function, which is not available in SQLite. The fix replaces this with the SQL standard string concatenation operator `||`, which is supported across SQLite, PostgreSQL, and CockroachDB.

## Changes
- Modified the `where()` method to use the `||` operator for string concatenation
- Added handling for NULL values using COALESCE to prevent NULL propagation in search expressions

## Issue
When attempting to perform searches on SQLite databases, the application was throwing the following error:
```
(sqlite3.OperationalError) no such function: concat 
[SQL: SELECT * FROM apipayments WHERE wallet_id = ? AND status != ? AND lower(concat(memo, amount, wallet_id, tag, status, time)) LIKE ? ORDER BY time desc LIMIT 10 ] 
[parameters: ('<redacted>', 'failed', '%be%')]
```

## Testing
- [x] Tested with SQLite database - search now works correctly
- [ ] Verified compatibility with PostgreSQL
- [ ] Verified compatibility with CockroachDB

### DIFF
--- a/lnbits/db.py
+++ b/lnbits/db.py
@@ -554,9 +554,9 @@ class Filters(BaseModel, Generic[TFilterModel]):
             for page_filter in self.filters:
                 where_stmts.append(page_filter.statement)
         if self.search and self.model and self.model.__search_fields__:
-            where_stmts.append(
-                f"lower(concat({', '.join(self.model.__search_fields__)})) LIKE :search"
-            )
+            # Use COALESCE to handle NULL values and || for cross-database compatible string concatenation
+            search_expr = " || ".join(f"COALESCE({field}, '')" for field in self.model.__search_fields__)
+            where_stmts.append(f"lower({search_expr}) LIKE :search")
 
         if where_stmts:
             return "WHERE " + " AND ".join(where_stmts)
--- a/lnbits/db.py
+++ b/lnbits/db.py
@@ -557,7 +557,9 @@ class Filters(BaseModel, Generic[TFilterModel]):
             # Use `COALESCE` to handle `NULL` values and `||`
             # for cross-database compatible string concatenation
             _fields = self.model.__search_fields__
-            search_expr = " || ".join(f"COALESCE({field}, '')" for field in _fields)
+            search_expr = " || ".join(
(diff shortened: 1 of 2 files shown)


motorina0:
Thank you!
What `SQLite` version are you using? This worked fine previously on `SQLite`.

Note:
 - `lint` fails (run `make all` and push to fix)
 - we are on code freeze for `v1`, this PR will be merged after the official release

PatMulligan:
Running `v3.37.2 2022-01-06` on this machine

dni:
we have an test failing here with a database error concerning bigint.
```
 FAILED tests/api/test_api.py::test_get_payments - sqlalchemy.exc.DBAPIError: 
(sqlalchemy.dialects.postgresql.asyncpg.Error) <class 'asyncpg.exceptions.InvalidTextRepresentationError'>: 
invalid input syntax for type bigint: ""

[SQL: 
            SELECT * FROM apipayments
            WHERE wallet_id = %s AND (status = 'success' OR status = 'pending') AND lower(COALESCE(memo, '') || COALESCE(amount, '') || COALESCE(wallet_id, '') || COALESCE(tag, '') || COALESCE(status, '') || COALESCE(time, '')) LIKE %s
            
            ORDER BY time desc
            
            ]
[parameters: ('32579a7fce7e48dc9460dff844d615ba', '%xxx%')]
```

`invalid input syntax for type bigint: ""` seems like input "" for bigint is not allowed, so `COALESCE(amount, '')` i guess fails.

hmmm

PatMulligan:
It looks like what's happening is COALESCE is trying to convert '' to **bigint** to match the type of `amount`. Let's see if this works!

dni:
check passed once :) nice

dni:
is there any tests we can write for this? that would be amazing

dni (approved):
i would love some tests, but ok for now, manually tested and i didn run into issues on sqlite

PR #3092, opened 2025-04-06
Initial Support to Receive Payments using Bolt12 Offers.

21M4TW:
-I started modifying the code to add Bolt12 offer support to LNBits.
-It is now possible to receive payments to invoices associated to a known offer, when using the CLN RPC wallet.
-There is a new apioffers table that was created for the DB. Some offer-related fields were also added to apipayments,
-A create_offer service was added, but currently it is not hooked to any UI or extension (there is however a test for it).
-A fetch_invoice function was added to the CLN RPC wallet, but otherwise I have not yet added support to pay to Bolt12 offers.


21M4TW:
The bolt12 python package from Rusty is obsolete and is incompatible with the new bolt12 strings. The best I was able to find is to use the CLN RPC to decode bolt12 invoices. When sending to bolt12 offers, a decoded format of the offers will also be required within LNbits. Is it reasonable to assume that wallets can provide bolt12 encoding/decoding functionalities?

ZenenTreadwell:
I want to see bolt12 integrated into Lnbits but I have a really hard time believing that all of these changes / additions are necessary in order to make that happen.

21M4TW:
> I want to see bolt12 integrated into Lnbits but I have a really hard time believing that all of these changes / additions are necessary in order to make that happen.

Well, everything about offers has to be added, and bolt11 decoding functionalities in LNBits run pretty deep in the code base, so this requires some changes as well. As of now the API allows to create offers, list them, enable and disable them. It is also possible to fetch invoices and pay invoices using bolt12 (support only with CLN RPC). The code is still using these "fake" bolt11 invoices for the payment functionalities that rely on them (the bolt11 field of the apipayments table is still storing bolt11 invoices, not bolt12). I would be cleaner to accept bolt12 data for this field as well, but this would likely break more things.

21M4TW:
If any reviewer could give me some initial feedback about this support for bolt12 it would be great! Thanks!

lnbits/core/services/payments.py
```diff
@@ -45,6 +57,206 @@
...
+        if settings.is_wallet_max_balance_exceeded(
+            user_wallet.balance_msat / 1000 + amount_sat
+        ):
+            raise OfferError(
+                f"Wallet balance cannot exceed "
+                f"{settings.lnbits_wallet_limit_max_balance} sats.",
+                status="failed",
+            )
+
+    while True:
```
ZenenTreadwell, line 99: What's going on here? Is there a reason this needs to happen in an infinite loop?

lnbits/core/crud/offers.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+async def enable_offer(offer_id: str) -> None:
+    await db.execute(
+        f"""
+        UPDATE apioffers SET active = :active, updated_at = {db.timestamp_placeholder("now")}
+        WHERE offer_id = :offer_id
+        """,
+        {"now": time(), "offer_id": offer_id, "active": OfferState.TRUE},
+    )
```
ZenenTreadwell, line 208: I'm suspicious of any f-strings that are being sent to the database; is there a way to compose these queries in a way that is more guaranteed to be sanitary?

lnbits/core/crud/offers.py
```diff
@@ -0,0 +1,228 @@
...
+    wallet_id: Optional[str] = None,
+) -> Optional[Offer]:
+    clause: str = "offer_id = :offer_id"
+    values = {
+        "wallet_id": wallet_id,
+        "offer_id": offer_id,
+    }
+
+    if wallet_id:
+        clause = f"({clause}) AND wallet_id = :wallet_id"
```
ZenenTreadwell, line 36: ; drop tables

lnbits/core/crud/offers.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+
+    values: dict = {
+        "wallet_id": wallet_id,
+        "created_at": since,
+    }
+    clause: list[str] = []
+
+    if since is not None:
+        clause.append(f"created_at > {db.timestamp_placeholder('time')}")
```
ZenenTreadwell, line 70: another f-string to the database

lnbits/tasks.py
```diff
@@ -200,14 +205,60 @@ async def invoice_callback_dispatcher(checking_id: str, is_internal: bool = Fals
...
-        payment.preimage = status.preimage
-        payment.status = PaymentState.SUCCESS
-        await update_payment(payment)
-        internal = "internal" if is_internal else ""
-        logger.success(f"{internal} invoice {checking_id} settled")
-        for name, send_chan in invoice_listeners.items():
-            logger.trace(f"invoice listeners: sending to `{name}`")
-            await send_chan.put(payment)
+    if payment:
+        if payment.is_in:
```
ZenenTreadwell, line 209: nitpick: I prefer the `payment and payment.is_in` syntax. Reads clearly and has one less indent

lnbits/wallets/corelightning.py
ZenenTreadwell, line 1: Seems solid, I would use this code with my node

tests/regtest/test_services_create_offer.py
```diff
@@ -0,0 +1,36 @@
...
+        memo=description,
+        single_use=True,
+    )
+    """
+    offer = decode(invoice_offer.bolt12)
+    assert offer.amount_msat == 1000000
+    assert offer.description == description
+    """
+
+    await disable_offer(wallet_id="10c8094f9d1c4ab4b9e3dfe964527297",offer_id=invoice_offer.offer_id)
```
ZenenTreadwell, line 30: Why is this wallet_id hard-coded? I don't see this in the rest of the test suite, usually the from_wallet is passed as a parameter to the test function

ZenenTreadwell (commented):
not a fan of f-strings in database queries, it seems like an unnecessary risk. Once again, lots of code here and it should definitely be combed over thoroughly. That said, integrating bolt12 is a large and important task, and this PR seems to have done a significant chunk of the work

21M4TW:
> not a fan of f-strings in database queries, it seems like an unnecessary risk. Once again, lots of code here and it should definitely be combed over thoroughly. That said, integrating bolt12 is a large and important task, and this PR seems to have done a significant chunk of the work

Ok thank you for the feedback. I can replace the f-strings

dni:
thanks for your contribution, i like it :D bolt12 is a huge topic and we need to make sure we implement it the right way. in some places your pr feels a bit heavy weight, but in general its good. someone in the lnbits core need to fully champion the bolt12 integration and this is a good place to start.

21M4TW:
> thanks for your contribution, i like it :D bolt12 is a huge topic and we need to make sure we implement it the right way. in some places your pr feels a bit heavy weight, but in general its good. someone in the lnbits core need to fully champion the bolt12 integration and this is a good place to start.

Thank you for looking at it. I am working on other components around this PR (dependencies, extension) to make it fully usable. Please let me know if there are other things I could do to help.

21M4TW:
> I think it might be a good idea to keep those components separate so that this PR can be processed in bite-sized pieces. Honestly, if there was a way to break this apart into a few different PRs that might help move this along. I know that the LNBits folks have it on their radar, though.

Thank you. Yes I was not planning to lump the new stuff with this PR. Glad to hear that this is on their radar! :)

21M4TW:
I have been working at implementing support for this in an extension, and I am now considering to add a bolt12 column to the apipayments table, so the invoice string is directly accessible from a Payment class instance if desired. In multiple places LNbits uses bolt11_decode to extract the payment_hash and/or the description from the bolt11 string, so I was thinking of maybe leaving in the fake bolt11 invoice I have been generating in order to avoid breaking backward compatibility, but this bolt12 column could be used by the code that supports it. What do you think? Thanks

21M4TW:
I have found another thing. Internal invoices are not working with the current version of this PR, as invoices are only discovered once a payment is received, due to the inexistence of push notification from CLN regarding invoice creation. So the best way I see for internal bolt12 payments in _pay_internal_invoice is to search for an offer with a matching offer_id in the apioffers table, then if there is one, look for a matching invoice at the funding source level using the payment_hash, then tag the payment as internal if this is the case.

21M4TW:
> I have found another thing. Internal invoices are not working with the current version of this PR, as invoices are only discovered once a payment is received, due to the inexistence of push notification from CLN regarding invoice creation. So the best way I see for internal bolt12 payments in _pay_internal_invoice is to search for an offer with a matching offer_id in the apioffers table, then if there is one, look for a matching invoice at the funding source level using the payment_hash, then tag the payment as internal if this is the case.

It is now fixed: https://github.com/lnbits/lnbits/pull/3092/commits/ff11d955da719097271349d790e36c942b2630df

21M4TW:
I just sent a PR to add bolt12 support to the NWC Provider extension, building on the work to add bolt12 support to lnbits: https://github.com/riccardobl/nwcprovider/pull/12

21M4TW:
I just sent a PR to add bolt12 support to NWC (NIP-47): https://github.com/nostr-protocol/nips/pull/1952

21M4TW:
What is the status on this? Anything I can do to help it move forward? Thanks!

ZenenTreadwell:
I don't think that I'm the person to review this. I'm not an official lnbits dev, but I will say that this PR is absolutely massive. It might help the team if you add a more comprehensive test suite? Aside from that, it's probably going to have to wait until the lnbits team feels like they actually need to integrate bolt12.

21M4TW:
> I don't think that I'm the person to review this. I'm not an official lnbits dev, but I will say that this PR is absolutely massive. It might help the team if you add a more comprehensive test suite? Aside from that, it's probably going to have to wait until the lnbits team feels like they actually need to integrate bolt12.

I can work on more tests. Besides @dni that had rebased this PR, I have not heard from the lnbits team though. Is there interest for it? I have been using it since I created the PR. I can rebase it to the latest version and look at creating more tests if others are interested.

21M4TW:
There is now a new rebased and cleaned up version: https://github.com/lnbits/lnbits/pull/3663

PR #3094, opened 2025-04-08
[Fix] - Bug on switching wallets and payment details

talvasconcelos:
There is a bug when switching wallets and checking payment details.

When selecting a new wallet, `window.history.replaceState()` was used which screwed vue router and the dynamic component stuff


dni (approved):

lnbits/static/js/wallet.js
```diff
@@ -1112,15 +1112,7 @@ window.WalletPageLogic = {
     }
   },
   watch: {
-    'g.wallet.id'(newVal, oldVal) {
-      try {
-        this.fetchChartData()
```
motorina0, line 1117: is the chart data. still updated when switching wallets?
why was this removed, does not look related to the original issue?

talvasconcelos: strangelly it does!

talvasconcelos: i'll test a bit more, i can add it back in!

motorina0: if its not related to this PRs fix then please do

motorina0 (approved):

arcbtc (approved):

Issue #3103, opened 2025-04-11
[BUG] Local time for invoices shown incorrectly and error shown


jjim1982:
The local time for invoices seems to be displayed incorrectly. The UTC time shown in the popup is correct, but the one above that says 9 hours ago when it should be a few seconds ago. It seems to be treating the UTC time as my local time.

<img width="566" alt="Image" src="https://github.com/user-attachments/assets/441ae4b9-ff83-4b0d-b485-465071ba6de6" />

I also get the error "**_can't compare offset-naive and offset-aware datetimes 500_**" for one of my wallets. However the incorrect time is shown in all wallets.

- LNbits version: 1
- Database postgres

Log:

`Apr 11 04:35:41 vps poetry[131837]:   + Exception Group Traceback (most recent call last):
Apr 11 04:35:41 vps poetry[131837]:   |   File "/home/ubuntu/.cache/pypoetry/virtualenvs/lnbits-FjepO-aV-py3.12/lib/python3.12/site-packages/starlette/_utils.py", line 76, in collapse_excgroups
Apr 11 04:35:41 vps poetry[131837]:   |     yield
Apr 11 04:35:41 vps poetry[131837]:   |   File "/home/ubuntu/.cache/pypoetry/virtualenvs/lnbits-FjepO-aV-py3.12/lib/python3.12/site-packages/starlette/middleware/base.py", line 186, in __call__
Apr 11 04:35:41 vps poetry[131837]:   |     async with anyio.create_task_group() as task_group:
Apr 11 04:35:41 vps poetry[131837]:   |   File "/home/ubuntu/.cache/pypoetry/virtualenvs/lnbits-FjepO-aV-py3.12/lib/python3.12/site-packages/anyio/_backends/_asyncio.py", line 815, in __aexit__
Apr 11 04:35:41 vps poetry[131837]:   |     raise BaseExceptionGroup(
Apr 11 04:35:41 vps poetry[131837]:   | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
Apr 11 04:35:41 vps poetry[131837]:   +-+---------------- 1 ----------------
Apr 11 04:35:41 vps poetry[131837]:     | Traceback (most recent call last):
Apr 11 04:35:41 vps poetry[131837]:     |   File "/home/ubuntu/.cache/pypoetry/virtualenvs/lnbits-FjepO-aV-py3.12/lib/python3.12/site-packages/starlette/middleware/errors.py", line 165, in __call__
[...]

PR #3104, opened 2025-04-11
[fix] offset naive dates error

motorina0:
Fixes error reported here: https://github.com/lnbits/lnbits/issues/3103
image

Also fixes wrong order of labels:
image

### DIFF
--- a/lnbits/static/js/wallet.js
+++ b/lnbits/static/js/wallet.js
@@ -1112,13 +1112,6 @@ window.WalletPageLogic = {
     }
   },
   watch: {
-    'g.wallet.id'(newVal, oldVal) {
-      try {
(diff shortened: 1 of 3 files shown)


lnbits/static/js/wallet.js
```diff
@@ -1112,13 +1112,6 @@ window.WalletPageLogic = {
     }
   },
   watch: {
-    'g.wallet.id'(newVal, oldVal) {
```
motorina0, line 1115: no longer required, `created()` function is now called when wallets change

dni (approved):

AxelHamburch:
I tested the PR, after restarting my LNbits instance it worked! 👍

PR #3107, opened 2025-04-12
[doc] Update installation.md to match v1.0.x requirements

tlindi:
Please comment and amend. Many users are now days reading docs page.

### DIFF
--- a/docs/guide/installation.md
+++ b/docs/guide/installation.md
@@ -78,11 +78,22 @@ poetry run lnbits-cli --help
 
 ```sh
 cd lnbits
-# Stop LNbits with `ctrl + x`
-git pull
-# Keep your poetry install up to date, this can be done with `poetry self update`
+# Stop LNbits with `ctrl + x` or with service manager
+# sudo systemctl stop lnbits
+
+# Update LNbits
+git pull --rebase
+
+# To keep your poetry up to date, use
+poetry env use python3.12
+poetry env remove python3.9
+poetry env list
 poetry install --only main
-# Start LNbits with `poetry run lnbits`
+
+# Start LNbits with
+poetry run lnbits
+
+# use LNbits admin UI Extensions page function "Update All" do get extensions onto proper level
 ```
 
 ## Option 2: Install script (on Debian/Ubuntu)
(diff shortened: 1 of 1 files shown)


tlindi:
same file is passing my LNbits prettier check.

docs/guide/installation.md
```diff
@@ -78,11 +78,34 @@ poetry run lnbits-cli --help
...
-# Stop LNbits with `ctrl + x`
-git pull
-# Keep your poetry install up to date, this can be done with `poetry self update`
+# Stop LNbits with `ctrl + x` or with service manager
+# sudo systemctl stop lnbits
+
+# To keep your poetry up to date, use
+pyenv install 3.12
+pyenv local 3.12
+pyenv rehash
```
dni, line 87: we are not using `pyenv`

docs/guide/installation.md
```diff
@@ -78,11 +78,34 @@ poetry run lnbits-cli --help
...
+poetry env use python3.12
+poetry env list
+poetry self update
```
dni, line 90: not sure if `poetry self update` is a good idea. we are not compatible wit hthe new poetry v2.x because of nix

tlindi: Would it be reasonable to define exact poetry version supported
with `poetry self update <version>`?

docs/guide/installation.md
tlindi: removed

docs/guide/installation.md
tlindi: > not sure if `poetry self update` is a good idea. we are not compatible wit hthe new poetry v2.x because of nix

removed.

dni (approved):

docs/guide/installation.md
```diff
@@ -78,11 +78,22 @@ poetry run lnbits-cli --help
...
+# Stop LNbits with `ctrl + x` or with service manager
+# sudo systemctl stop lnbits
+
+# Update LNbits
+git pull --rebase
+
+# To keep your poetry up to date, use
+poetry env use python3.12
+poetry env remove python3.9
+poetry env list
```
motorina0, line 90: Maybe this should be listed a separate prerequisite (not to be misunderstood that it has to be ran each time)

tlindi: Rephrased instructions

dni (approved):

tlindi:
Thanks

Issue #3114, opened 2025-04-13
[BUG] Webhook "not sent yet"


DoktorShift:
**Describe the bug**
When i pay an invoice the attached webhook does not get called.

**To Reproduce**

1. Create an invoice and attach webhook or use lnurlp to attach a webhook in the UI.
2. Pay this paycode/invoice
3. Go to payed invoice "details" 
4. See badge in webhook column "not sent yet"

**Expected behavior**
Sent out a webhook to the desired location

**Screenshots**

![Image](https://github.com/user-attachments/assets/30662441-f177-407e-a15f-8226b6ab5166)

**Software:**

- LNbits v1
- Database sqlite

**Additional context**
1) I came to this error trough extension SatsPayServer. Tested afterwards with lnurlp and finaly also with creating invoice via curl and attach webhook.

2) I also changed the webhook destination multiple times to exclude its a receiver problem.

3) I recognized a new feature called "Payment wait time". i set the default 5 seconds to 0 seconds to test if this new function is causing errors. No Success.

honzagalajda:
I came across the same error when working with demo.lnbits.com…

honzagalajda:
So, I found a way around it by using the LNbits WebSocket `wss://demo.lnbits.com/api/v1/ws/{invoice_key}`

motorina0:
Fixed here: https://github.com/lnbits/lnbits/pull/3119

PR #3119, opened 2025-04-17
fix: webhook call on invoice pay

motorina0:
Reported here: https://github.com/lnbits/lnbits/issues/3114
- `webhook_status` is `TEXT` in DB
- use `payment.json()` instead of `payment.dict()`
- fix error logging (include error in message, not as extra param)
- add webhook unit test

### DIFF
--- a/lnbits/core/services/notifications.py
+++ b/lnbits/core/services/notifications.py
@@ -205,24 +205,23 @@ async def dispatch_webhook(payment: Payment):
     logger.debug("sending webhook", payment.webhook)
 
     if not payment.webhook:
-        return await mark_webhook_sent(payment.payment_hash, -1)
+        return await mark_webhook_sent(payment.payment_hash, "-1")
 
     headers = {"User-Agent": settings.user_agent}
     async with httpx.AsyncClient(headers=headers) as client:
-        data = payment.dict()
         try:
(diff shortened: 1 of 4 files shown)


lnbits/core/services/notifications.py
```diff
@@ -205,43 +205,45 @@ async def dispatch_webhook(payment: Payment):
...
-        logger.error("Error sending push payment notification", e)
+        logger.error(f"Error sending push payment notification {e!s}")
 
-    if payment.webhook and not payment.webhook_status:
-        await dispatch_webhook(payment)
+    try:
+        if payment.webhook and not payment.webhook_status:
+            await dispatch_webhook(payment)
+    except Exception as e:
+        logger.error(f"Error dispatching webhook: {e!s}")
```
dni, line 246:
```suggestion
    if payment.webhook and not payment.webhook_status:
        try:
            await dispatch_webhook(payment)
        except Exception as e:
            logger.error(f"Error dispatching webhook: {e!s}")
```

dni: only try catch what can fail

dni (approved):
not in love with saving an int as string in db, but ok :) that one nitpick and its fine.

talvasconcelos (approved):

PR #3127, opened 2025-04-25
security: update package `h11`

dni:
updated uvicorn as its the package requiring h11, also bumped h11 to 0.16.
https://github.com/lnbits/lnbits/security/dependabot/86


dni:
@prusnak :wave:

prusnak:
> @prusnak 👋

as always, do not use `poetry lock` but `poetry lock --no-update`

force pushed the fix which builds with nix (no changes needed in nix files)

motorina0 (approved):

prusnak (approved):

PR #3139, opened 2025-05-02
fix: better differentiation between `UNAUTHORIZED` and `FORBIDDEN`

motorina0:
### Summary
 - Use 401 when credentials are missing or incorrect
 - Use 403 when credentials are correct, but access is not allowed.

**Problem**:
When the server auth secret is changed, all the client side tokens are invalidated.
This lead to an infinite loop from the `wallet` -> `error` -> `login` -> `wallet` -> ....
 - user token is invalidated (by server side secret update)
 - user goes to `wallet` page
 - user is redirected to `error` page (invalid token)
 - error page on `401` redirects to `login` page
 - `login` page finds a token (though invalid) and redirects to `wallet` page

**Solution**:
 - narrow the `401 UNAUTHORIZED` returned codes
 - use `403 FORBIDDEN` where possible
 - on error page logout if `401 UNAUTHORIZED` is retuned

### DIFF
--- a/lnbits/core/views/auth_api.py
+++ b/lnbits/core/views/auth_api.py
@@ -74,7 +74,7 @@ async def get_auth_user(user: User = Depends(check_user_exists)) -> User:
 async def login(data: LoginUsernamePassword) -> JSONResponse:
     if not settings.is_auth_method_allowed(AuthMethods.username_and_password):
         raise HTTPException(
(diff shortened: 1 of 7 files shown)


dni (approved):
love it, thats a nice PR!

talvasconcelos (approved):
tested! fixes infinite loop

PR #3089, opened 2025-04-03
refactor: fundingsource Invoice-, PaymentResponses

dni:
- refactor the wallets to kwargs instead of arguments, to make them more readable and flexible
- should be the same as before

PR #3150, opened 2025-05-09
feat: Add Strike Wallet Integration

santyr:
This pull request introduces a new wallet integration for LNbits, enabling users to leverage the Strike API for Lightning Network payments.

**Key Features:**

*   **Strike API Integration:** Implements a `StrikeWallet` class that interacts with the Strike API to manage Lightning invoices and payments.
*   **Core Wallet Functionality:** Supports essential wallet operations, including:
    *   Retrieving wallet balance.
    *   Creating Lightning invoices (receive requests).
    *   Paying Lightning invoices.
    *   Checking the status of invoices and payments.
*   **Rate Limiting and Retries:** Incorporates rate limiting to comply with Strike API usage policies and implements retry mechanisms to handle transient errors.
*   **Error Handling:** Includes robust error handling to manage API errors and other potential issues.
*   **Asynchronous Implementation:**  Uses `asyncio` and `httpx` for efficient asynchronous operation within the LNbits environment.
*   **Configuration:**  Relies on LNbits settings for API endpoint and key configuration.

**Implementation Details:**

*   The `StrikeWallet` class inherits from the LNbits `Wallet` base class and implements the required interface methods.
*   Rate limiting is achieved using a `TokenBucket` to control request frequency to different API endpoint categories.

This new wallet integration expands the options available to LNbits users, providing a convenient way to manage Lightning payments through the Strike platform.


santyr:
Currently used in production for the Lightning Goats Project.

https://lightning-goats.com
https://primal.net/p/nprofile1qqsxd84men85p8hqgearxes2az8azljn08nydeqa0s3klayk8u7rddshkxjm3

-S

motorina0:
This is GREAT! 
Thank you for opening this PR!

I had a quick look, looks fine. I will review in detail next week.

lnbits/wallets/strike.py
```diff
@@ -0,0 +1,467 @@
...
+
+            if state in {"SUCCEEDED", "COMPLETED"}:  # If payment succeeded.
+                preimage = data.get("preimage") or data.get("preImage")  # Get payment preimage.
+                return PaymentResponse(True, payment_id, fee_msat, preimage, None)  # Return successful payment response.
+
+            # Strike often returns 202/PENDING immediately; treat as “still working”.
+            if state in {"PENDING", "QUEUED", "READY_TO_SETTLE", ""}:  # If payment is still pending.
+                return PaymentResponse(None, payment_id, None, None, None)  # Return pending payment response.
+
+            return PaymentResponse(False, payment_id, None, None, f"State: {state}")  # Return failed payment response with state.
```
motorina0, line 322: Isn't there an explicit list of states for 'FAILED'?
If a new `pending` status is introduced by the API then it will not be recognized and the payment will be marked as failed (which has an impact on the wallet balance).

It is better to check explicitly for `failed` and leave the default branch for `pending`.

lnbits/wallets/strike.py
```diff
@@ -0,0 +1,467 @@
...
+
+        except httpx.HTTPStatusError as e:
+            msg = e.response.json().get("message", e.response.text)  # Get error message from response.
+            return PaymentResponse(False, None, None, None, f"Strike API error: {msg}")  # Return payment response with error.
```
motorina0, line 326: There was a refactoring in this PR https://github.com/lnbits/lnbits/pull/3089 where explicit named parameters are used. Can you please use that form.

Example:
https://github.com/lnbits/lnbits/pull/3089/files#diff-dd5221a5b7e9defe1cbdc1f3e3e0d709cdea670356f838088ee785950af45bedR175

santyr: Will do.

motorina0:
Can you please also add the UI for the config.

Example:
image

santyr:
Where do I add the UI for the config?

motorina0:
> Where do I add the UI for the config?

See: `lnbits/static/js/components/lnbits-funding-sources.js`

If its not your area, I can add it later, just let me know.

santyr:
I've updated funding sources by adding the following, but still don't see ui form elements when I run locally.  I'm probably missing something obvious:

breez_greenlight_device_cert: 'Greenlight Device Cert',
breez_greenlight_invite_code: 'Greenlight Invite Code'
}
        ],
        [
          'StrikeWallet',
          'Strike',
          {
            strike_api_endpoint: 'API Endpoint',
            strike_api_key: 'API Key'
          }
]
]
}

motorina0:
> I've updated funding sources by adding the following, but still don't see ui form elements when I run locally. I'm probably missing something obvious:

You need to run `make bundle`.

Then hard refresh in the browser.

santyr:
Nice!  Thank you.

image

lnbits/wallets/strike.py
```diff
@@ -0,0 +1,475 @@
...
+            return StatusResponse("Connection error", 0)
+
+    async def create_invoice(
+        self,
+        amount: int,
+        memo: Optional[str] = None,
+        description_hash: Optional[bytes] = None,
+        unhashed_description: bytes | None = None,  # Add this parameter
+        **kwargs,
+    ) -> InvoiceResponse:
```
dni, line 250: is there a way to get the preimage of a created invoice? or pass in a custom preimage to keep track of it?

dni (approved):
looks pretty good! only thing i would add is having a preimage when creating an invoice.

lnbits/wallets/strike.py
santyr: The create_invoice method in the StrikeWallet class only allows you to specify the amount, memo (description), and optionally a description hash. The preimage is handled internally by Strike and is only revealed after a successful payment.

Strike returns the preimage in the response to the payment execution request. Specifically, within the pay_invoice method:

1) After creating a payment quote, the code executes it with a PATCH request to /payment-quotes/{quote_id}/execute.

2) The response from this request (stored in the e variable) is parsed as JSON.

3) The code then attempts to extract the preimage from the JSON data using these lines:

`preimage = data.get("preimage") or data.get("preImage")`

If a preimage is found, it's included in the PaymentResponse object returned by the pay_invoice method.

As I understand it, the preimage serves as proof of payment. It's only revealed after the payment has been successfully claimed by the recipient.  Anyone who knows the preimage can verify that the payment was made by hashing it and comparing it to the hash lock in the original invoice.  The preimage is unique to each invoice, ensuring that it corresponds to a specific payment.

dni: is that some AI :D?
in some fundingsources we are able to provide a payment_hash/preimage up front when creating an invoice, that is how we keep track of the preimage when creating an invoice. like this: https://github.com/lnbits/lnbits/blob/dev/lnbits/wallets/lndgrpc.py#L152

santyr: I did use an AI to help me write that, since I wasn't clear on the use of a preimage until you brought it up actually.  

In some cases, it's easier to get a preimage, such as when you're running your own node, and you are taking less of a risk in using it, since you're not revealing it to anyone before your Sats arrive.  

 As I understand it, the preimage is never revealed until the payment is received, then it back propagates along the payment route until it reaches the whoever paid the invoice.  To quote the AI, 'In essence, the preimage is the secret "key" that, once revealed by the recipient, cascades backward along the payment path, unlocking each HTLC and allowing funds to flow securely and atomically to the recipient while fees flow to intermediaries.'

On the other hand the payment hash, or the hash lock, is always available in a bolt 11 invoice, and it can be used as an identifier when you're tracking a payment and to verify it's legit.  The payment hash (hash lock)  is a the preimage in hashed form.

I pretty new to coding anything for Lightning, but couldn't you just extract the payment hash from the invoice and then confirm it against the preimage when the recipient (Strike in this case) reveals it as part of the regular payment flow?

dni: thank you for not using AI :D it seemed a bit like it, sorry for accusing.
you are completely right. if you have the preimage of the payment  it is the proof of the payment. so you pay the invoice, you get the preimage.
so basically until recently we did not keep track of preimages when creating invoices / payments we receive, only payments we pay. but keeping track of preimages for invoices we give out can also be very useful and even to just have consistent state of the payments. so most nodes give you the option to provide a hash or preimage upfront to create the invoice which makes it easy for lnbits to keep track of, its not 100% necessary in the context of receiving, but  for lnbits as accounting software its beneficial.

dni: they payment_hash in the invoice is not the be trusted :D you can just change the whole invoice from a 100k sats to a 1 sat invoice and keep the payment_hash the same if your account system trust just the hash you got a problem.

santyr: I see.  The payment hash is really just the hashed preimage, and could be used to verify if the backend (Strike for example)  received the payment, but the preimage doesn't have any payment amount data in it.  Out of curiosity, in the code you shared whats the reason for supplying a preimage to LND rather than letting LND generate it or only supplying the payment hash to LND while holding onto the preimage in LNbits?  (hold invoices?)

lnbits/wallets/strike.py
```diff
@@ -0,0 +1,551 @@
...
+        while self._running and settings.lnbits_running:
+            loop_start = time.time()
+            had_activity = False
+
+            req_budget = max(
+                1, rate_limit * sleep_s // 60
+            )  # Calculate request budget based on sleep time.
+            processed = 0
+
+            for inv in list(self.pending_invoices):  # Iterate through pending invoices.
```
motorina0, line 475: After a server restart this list will be empty.

lnbits/wallets/strike.py
```diff
@@ -0,0 +1,551 @@
...
+                },
+                "targetCurrency": "BTC",
+            }
+            if description_hash:
+                payload["bolt11"]["descriptionHash"] = description_hash.hex()
+
+            r = await self._post(
+                "/receive-requests",  # Create a receive request (invoice) on Strike.
+                json=payload,
+                headers={**self.client.headers, "idempotency-key": idem},
```
motorina0, line 277: Where is this `idempotency-key` header documented.
I could not find it here https://docs.strike.me/api/create-a-receive-request

motorina0:
I have did some code clean-up and refactorings.
@santyr please check this PR in your repo https://github.com/santyr/lnbits/pull/2/files

I have left a few `#todo` comments for you

lnbits/wallets/strike.py
santyr: I've removed the idempotency-key headers these were holdovers from early development.  They're totally unneeded and could be harmful.

https://github.com/santyr/lnbits/commit/3db1133e68df0d0b2b73356d1a7940947d174920

motorina0:
@santyr I have simplified the retry logic, it is better to be handled at the business logic level

There is only one `#todo` left: `pending_payments` will be empty after a server restart. I think another endpoint must be used if now `quote_id` found (instead of returning `pending`). See `#todo` in code.

Also, can you please test that my refactoring did not broke anything.

lnbits/wallets/strike.py
motorina0: LNbits does a regular check (every `30 min`) on `payments` like so:

```py
  funding_source = get_funding_source()
  if self.is_out:
      status = await funding_source.get_payment_status(self.checking_id)
  else:
      status = await funding_source.get_invoice_status(self.checking_id)
```

So these two functions need to make the right API calls, return the correct status and LNbits takes care of the rest.

santyr:
@motorina0  your changes seem to be running fine. (they're in being used by the Lightning Goats right now.) I did end up fixing an url to remove an extra /v1 and reworking get_invoice_status to prevent fallback to wrong endpoint after seeing the initial 404 caused by the extra /v1 in the path.

See: 
https://github.com/santyr/lnbits/commit/64ffb26e96f10543cf008743d94ad8cd9b8cb868 
https://github.com/santyr/lnbits/commit/ca06c08c7ea77ff6476a4cda251500f83dcd00ef

santyr:
I've modified get_payment_status() to address the #todo regarding `pending_payments` being lost on server restart.  This should do it:

https://github.com/santyr/lnbits/commit/13776c5e15df0f2788ac0d3e808b9ef4b3ad2180

motorina0:
@santyr thank you for pushing the changes.

- I have did a bit of refactoring (no functional changes). Please have a look and test.
- After the CI regtest issue is fixed (unrelated to this PR) I will merge this PR.
- I have also renamed `Strike` to `Strike (alpha)` in the funding source list. I will remove the `alpha` after this feature is part of at least one minor release. We will be waiting confirmation from you that all works as expected with the official release.

Thank you again for your confirmation.

PR #3165, opened 2025-05-19
[fix] timezone for payment list

motorina0:
- use `.local()` for date
- use `created_at` instead of date string


dni (approved):

kallerosenbaum:
Thanks for fix. I triggered the bug on a local instance v1.1.0 using FakeWallet, then I switched to dev branch and tested again. It's fixed there!

PR #3192, opened 2025-06-13
 CI: lnbits-boltz docker, for those who want instant payments + boltz funding source UX tweak

arcbtc:
This pr does a few things:
* creates an additional docker for lnbits with boltz built in 
* automatically creates a wallet in boltz if boltz funding source is selected and makes mnemonic fetchable from frontend
* adds an extra funding source prop function to allow readonly files that can also be rendered as a qr (useful for boltz and also phoenixd)


Dockerfile
```diff
@@ -50,9 +50,42 @@ COPY --from=builder /app/.venv .venv
...
+    tar -xvzf boltz-release.tar.gz && \
+    cd bin/linux_amd64 && \
+    chmod +x boltzd boltzcli && \
+    mv boltzd /usr/local/bin/ && \
+    mv boltzcli /usr/local/bin/
+
+WORKDIR /app
+
+# === END: Add Boltz ===
+
```
jackstar12: Its simpler to use the boltz-client docker image for getting the binaries. There's also no need for including boltzcli.

At the top, add
`FROM boltz/boltz-client:latest AS boltz`
So you can change it down here to
```suggestion
COPY --from=boltz /bin/boltzd /usr/local/bin/
```

lnbits/wallets/boltz.py
```diff
@@ -80,6 +80,15 @@ async def status(self) -> StatusResponse:
 
         return StatusResponse(None, response.balance.total * 1000)
 
+    async def create_wallet(self):
```
jackstar12: how will the user be able to see the mnemonic?

arcbtc: I plan on adding in the funding source UI + qr code, which can be scanned in greenwallet.
So flow will be, select boltz (wallet is made or fetched by name "lnbits" via grpc), user can scan a qr code in greenwallet. On our saas we will set boltz as the default in env, and this will happen on boot.

arcbtc: I feel you are way more qualified to make this pr than me 😂, so thanks for pointing me in the right direction, alternatively feel free take over the pr.

lnbits/wallets/boltz.py
```diff
@@ -225,3 +249,25 @@ async def paid_invoices_stream(self) -> AsyncGenerator[str, None]:
...
+
+    async def _fetch_wallet(self, wallet_name: str) -> Optional[str]:
+        try:
+            request = boltzrpc_pb2.GetWalletRequest(name=wallet_name)
+            response = await self.rpc.GetWallet(request, metadata=self.metadata)
+            logger.info(f"Wallet '{wallet_name}' already exists with ID {response.id}")
+            return settings.boltz_mnemonic
+        except AioRpcError as exc:
+            details = exc.details() or "unknown error"
+            if "not found" not in details.lower():
```
jackstar12: You can also check for the NOT_FOUND status code here: https://grpc.io/docs/guides/status-codes/

motorina0:
disable nix until poetry/uv ready

motorina0:
- [x] check cannot connect errors in docker logs
- [x] how to restart boltz client (or the docker container) is something goes wrong
- [x] test docker container re-create (same green wallet config)

motorina0 (approved):

motorina0 (approved):

PR #3204, opened 2025-06-19
Exclude gbp from Binance exchange rates

blackcoffeexbt:
For an unknown reason, the Binance API returns the BTCGBP pair at around 40% of the value of other exchanges which results in an incorrect average BTCGBP value for fiat calculations.


motorina0 (approved):

motorina0:
Will be superseded by:
https://github.com/lnbits/lnbits/pull/3206

PR #3209, opened 2025-06-23
We need to use a nostr library nostr-sdk works well

arcbtc:
Having a decent helper for nostr in core will encourage more nostr-centric extensions, and less code duplication
Tested with nostr-sdk 0.42.1

UPDATE: everyone in core hated the class, so ill just make some examples somewhere using the sdk directly


bitkarrot:
Cool. No frontend? If front end for the vue.js sections, I really like [nostr-tools](https://github.com/nbd-wtf/nostr-tools)

motorina0 (approved):

PR #3213, opened 2025-06-24
chore: update poetry, disable nix workflow and python package updates

dni:
- disable nix workflow
- update to poetry v2
- update python packages


.github/workflows/nix.yml
```diff
@@ -25,6 +25,7 @@ on:
 
 jobs:
   nix:
+    if: false
```
motorina0: Add an explainer

dni: done

PR #3214, opened 2025-06-24
fix: error handling

### LINKED ISSUE #3188: [BUG] response is undefined when lnurlscan times out
**Describe the bug**
When lnurlscan times out, LNBits has an uncaught exception at https://github.com/lnbits/lnbits/blob/e750165cc3eb7e611b0a8df8df44d3749c4cd94c/lnbits/static/js/wallet.js#L345

The `response` variable is not defined in this case.

Furthermore, there is no clear way to change the timeout for servers which are slow (or lower the timeout in order to reproduce errors like this with fast servers).

**To Reproduce**

1. Make sure the .well-known/lnurlp you'll be hitting takes >5 seconds before it responds to requests
2. Go to "Paste Invoice"
3. Type in someone's lighting address & click READ
4. See generic error in the UI and in the browser's Dev Console

**Expected behavior**
An error is shown saying the back end timed out and instructions on how to adjust the length of time LNBits will wait for a response are provided.

**Screenshots**

Image

- LNbits version: 1.1.0
- Database: sqlite

**Additional context**

Self-hosted servers can sometimes be slower than we'd like, but by giving good error messages and allowing the user to define how long they want to wait, this can be mitigated while users try to improve the performance of their servers.

talvasconcelos:
Handle server errors and be more informative. I think 5 sec is more than enough as timeout. Should we change it? 

Fixes #3188

### DIFF
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -136,14 +136,14 @@ async def api_lnurlscan(
             except httpx.HTTPStatusError as exc:
                 if exc.response.status_code == 404:
                     raise HTTPException(HTTPStatus.NOT_FOUND, "Not found") from exc
-                else:
-                    raise HTTPException(
-                        status_code=HTTPStatus.SERVICE_UNAVAILABLE,
-                        detail={
-                            "domain": domain,
-                            "message": "failed to get parameters",
-                        },
-                    ) from exc
+
+                raise HTTPException(
+                    status_code=HTTPStatus.SERVICE_UNAVAILABLE,
+                    detail={
+                        "domain": domain,
+                        "message": "failed to get parameters",
+                    },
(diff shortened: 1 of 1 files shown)


lnbits/core/views/api.py
```diff
@@ -130,14 +130,20 @@ async def api_lnurlscan(
...
-                    detail={"domain": domain, "message": "failed to get parameters"},
-                )
-
+            try:
+                r = await client.get(url, timeout=5)
+                r.raise_for_status()
+            except httpx.HTTPStatusError as exc:
+                if exc.response.status_code == 404:
+                    raise HTTPException(HTTPStatus.NOT_FOUND, "Not found") from exc
+                else:
```
dni: that else is not required

PR #3221, opened 2025-06-26
chore: update python packages + formatting + cleanup

dni:
some packages are not updated.
- sqlalchemy
- pydantic
- lnurl
- httpx required by lnurl

interestingly updating the linters now required formatting.

removed `lnbits/requestvars.py` it also had a new weird linting error and by checking it out, it seemed to be used in only 1 place, i think it was some legacy code.

had to change import from websockets library, to add a legacy namespace

### DIFF
--- a/lnbits/core/services/payments.py
+++ b/lnbits/core/services/payments.py
@@ -20,7 +20,7 @@ from lnbits.exceptions import InvoiceError, PaymentError
 from lnbits.fiat import get_fiat_provider
 from lnbits.helpers import check_callback_url
 from lnbits.settings import settings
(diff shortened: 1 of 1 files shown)


lnbits/core/crud/__init__.py
```diff
@@ -85,24 +85,16 @@
 )
 
 __all__ = [
-    # payments
```
motorina0: these are useful, why remove?

dni: now this exports are sorted by ruff, which made the comments bad, imho its better for them to be sorted automatically anyways

PR #3236, opened 2025-07-02
feat: Allow custom memo on pay invoice

### LINKED ISSUE #2416: [Feature request] Allow to set memo when paying a lightning invoice
It would be great if we can set the memo to show in lnbits when we are paying a lightning invoice using lnurl. At the moment, I am using /api/v1/payments call to pay for an invoice but I can see that the "memo" value gets ignored in that api call if the "out" is true. It seems to use the description from the invoice but I am hoping to pass in the separate memo value in the api call so it's easy for us to tell what this payment was for in lnbits.

talvasconcelos:
Allows for an optional custom internal memo when paying an invoice. If payment has an internal memo, it's stored in `extra.internal_memo` for the payment and displayed in the payments list.

Pay Invoice Dialog:
image

Pay LNURL Dialog:
image

Payment List:
image

Closes #2416


lnbits/core/templates/core/wallet.html
```diff
@@ -848,6 +848,20 @@ <h5 class="q-my-none text-bold">
             </div>
             <q-separator></q-separator>
             <h6 class="text-center" v-text="parse.invoice.description"></h6>
+            <q-input
+              filled
+              dense
+              v-model="parse.data.internalMemo"
+              label="Internal memo"
+              class="q-mb-md"
+              :rules="[ val => !val || val.length <= 64 || 'Please use maximum 64 characters' ]"
```
motorina0: lets make it `512`

lnbits/core/templates/core/wallet.html
```diff
@@ -1079,6 +1093,20 @@ <h6 class="text-center" v-text="parse.invoice.description"></h6>
                     label="Comment (optional)"
                     :maxlength="parse.lnurlpay.commentAllowed"
                   ></q-input>
+                  <q-input
+                    filled
+                    dense
+                    v-model="parse.data.internalMemo"
+                    label="Internal memo (optional)"
+                    class="q-my-md"
+                    :rules="[ val => !val || val.length <= 64 || 'Please use maximum 64 characters' ]"
```
motorina0: 512

lnbits/static/js/base.js
```diff
@@ -31,11 +31,17 @@ window.LNbits = {
...
     },
-    payInvoice(wallet, bolt11) {
-      return this.request('post', '/api/v1/payments', wallet.adminkey, {
+    payInvoice(wallet, bolt11, internalMemo = null) {
+      const data = {
         out: true,
         bolt11: bolt11
-      })
+      }
+      if (internalMemo && internalMemo !== '') {
```
motorina0: empty string is falsly: https://developer.mozilla.org/en-US/docs/Glossary/Falsy
so use of `if (internalMemo){` is sufficient

lnbits/static/js/base.js
```diff
@@ -44,16 +50,28 @@ window.LNbits = {
...
         callback,
         description_hash,
         amount,
         comment,
         description,
         unit
-      })
+      }
+
+      if (internalMemo && internalMemo !== '') {
```
motorina0: https://developer.mozilla.org/en-US/docs/Glossary/Falsy

motorina0:
The user should have the option to add an `internal memo` also when it creates an invoice: add a subtle way to show/toggle the internal memo.

Make the internal memo input multi line (default 2 rows) just like on create invoice dialog.

UI Suggestion:
Internal memo button:
image

Show internal memo text area
image

lnbits/core/views/payment_api.py
```diff
@@ -347,6 +344,8 @@ async def api_payments_pay_lnurl(
     if data.unit and data.unit != "sat":
         extra["fiat_currency"] = data.unit
         extra["fiat_amount"] = data.amount / 1000
+    if data.internal_memo:
+        extra["internal_memo"] = data.internal_memo
```
motorina0, line 359: length should be checked server-side too

talvasconcelos:
> The user should have the option to add an `internal memo` also when it creates an invoice: add a subtle way to show/toggle the internal memo.
> 
> Make the internal memo input multi line (default 2 rows) just like on create invoice dialog.


Honestly, i'm not too fond of the toggle show/hide internal memo input, it's one more extra click, on a process that should be quick and straightforward. I think it looks better to have it, shown, small (dense) with autogrow flag, so it add more lines as the user types. What do you think?

image

image

motorina0:
> it's one more extra click, on a process that should be quick and straightforward. I do think `Internal Memo` should be on-demand only.
 - the internal memo is optional and only set occasionally (not with every payment)
 - for the old flow the UI and UX is the same
 - can we try this inside the memo input (just a suggestion)?
 ```
 <template v-slot:append>
                  <q-icon
                    name="..."
                    @click="..."
                    class="cursor-pointer"
                  />
</template>
```
                

>  I think it looks better to have it, shown, small (dense) with autogrow flag, so it add more lines as the user types.
- I agree

File: lnbits/static/js/wallet.js [hunk 6]
```diff
@@ -10,10 +10,12 @@ window.WalletPageLogic = {
         invoice: null,
         lnurlpay: null,
         lnurlauth: null,
+        toggleInternalMemo: false,
```
motorina0: is this still needed?

lnbits/static/js/wallet.js
```diff
@@ -36,9 +38,11 @@ window.WalletPageLogic = {
         units: ['sat'],
         unit: 'sat',
         fiatProvider: '',
+        toggleInternalMemo: false,
```
motorina0: is this still needed?

lnbits/core/views/payment_api.py
```diff
@@ -338,6 +302,44 @@ async def api_payments_pay_lnurl(
                 f" {amount_msat} msat, got {invoice.amount_msat}."
             ),
         )
+    return invoice
+
+
+async def _fetch_lnurl_params(client, data, amount_msat, domain):
```
motorina0: add types to params
add return type

lnbits/core/views/payment_api.py
```diff
@@ -285,50 +282,17 @@ async def api_payments_fee_reserve(invoice: str = Query("invoice")) -> JSONRespo
...
-            r.raise_for_status()
-        except (httpx.HTTPError, ssl.SSLError) as exc:
-            logger.warning(exc)
-            raise HTTPException(
-                status_code=HTTPStatus.BAD_REQUEST,
-                detail=f"Failed to connect to {domain}.",
-            ) from exc
-
-    params = json.loads(r.text)
+def _validate_lnurl_response(params, domain, amount_msat):
```
motorina0: - add types to params
- add return type

lnbits/core/views/payment_api.py
```diff
@@ -338,6 +304,46 @@ async def api_payments_pay_lnurl(
...
+    return json.loads(r.text)
+
+
+@payment_router.post("/lnurl")
+async def api_payments_pay_lnurl(
+    data: CreateLnurl, wallet: WalletTypeInfo = Depends(require_admin_key)
+) -> Payment:
+    domain = urlparse(data.callback).netloc
+    headers = {"User-Agent": settings.user_agent}
+    async with httpx.AsyncClient(headers=headers, follow_redirects=True) as client:
```
motorina0: why not move `async with httpx.AsyncClient` in `_fetch_lnurl_params`?

PR #3245, opened 2025-07-07
feat: `run_interval` in tasks and refactor check_pending_payments

dni:


lnbits/tasks.py
```diff
@@ -153,51 +151,18 @@ async def wrapper() -> None:
...
-                    logger.debug(f"{prefix} pending {payment.checking_id}")
-                await asyncio.sleep(0.01)  # to avoid complete blocking
-            logger.info(
-                f"Task: pending check finished for {count} payments"
-                f" (took {time.time() - start_time:0.3f} s)"
-            )
-        await asyncio.sleep(sleep_time)
+    async def wrapper() -> None:
+        while settings.lnbits_running:
+            await func()
```
motorina0, line 162: this should be in a try-catch so the loop does not stop in case of exception

dni: this is already done by create_permanent_task

PR #3257, opened 2025-07-09
chore: get rid of mypy notes and typecheck more functions

dni:
screenshot-1752046323


lnbits/core/tasks.py
```diff
@@ -13,7 +13,7 @@
 from lnbits.core.crud.payments import get_payments_status_count
 from lnbits.core.crud.users import get_accounts
 from lnbits.core.crud.wallets import get_wallets_count
-from lnbits.core.models import AuditEntry
```
motorina0, line 16: why change this?

dni: just importing it more specifically

PR #3279, opened 2025-07-17
chore: replace passlib with bcrypt

dni:
passlib does not seemed to be maintained for a longer time and runs into some combatibility issues soon and already throws warning


dni:
tested

PR #3298, opened 2025-07-31
feat: allow creating/update user with nostr `npub1` pubkey

### LINKED ISSUE #3296: Support Nostr npub format for public key input in account creation
Currently, when creating an account or updating user settings in LNbits, the "Public Key" field only accepts Nostr public keys in raw hexadecimal format (64 hex characters). This creates a poor user experience because most Nostr users are familiar with their public key in the `npub` format (bech32-encoded), which is the standard way Nostr clients display public keys.

### Current Behavior
- The field is labeled simply as "Public Key" 
- Only accepts 64-character hexadecimal strings
- Users must manually convert their `npub` key to hex format using external tools like https://nostrcheck.me/converter/

### Expected Behavior
- The field should be clearly labeled as "Nostr Public Key (npub or hex)"
- Accept both formats:
  - `npub` format (e.g., `npub180cvv07tjdrrgpa0j7j7tmnyl2yr6yr7l8j4s3evf6u64th6gkwsyjh6w6`)
  - Hex format (e.g., `3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d`)
- Automatically convert `npub` to hex format internally when storing

### Technical Details
According to [NIP-19](https://github.com/nostr-protocol/nips/blob/master/19.md), `npub` is the bech32-encoded format of Nostr public keys with the human-readable prefix "npub". This is the format that users see and copy from their Nostr clients.

### Proposed Solution
1. Update the input label to clarify it accepts Nostr keys
2. Add validation to check if the input starts with "npub1"
3. If it's an npub, decode it using bech32 to get the hex format
4. Store the hex format in the database (maintaining backward compatibility)
5. Update the validation function `is_valid_pubkey()` to handle both formats

### Benefits
- Better user experience - users can paste their key directly from Nostr clients
- No need for external conversion tools
- Follows Nostr ecosystem standards
- Maintains backward compatibility

### Implementation Notes
- The `nostr-tools` library already included in the project (see package.json) likely has bech32 encoding/decoding utilities
- Only the input validation and conversion logic needs to be updated
- The database can continue storing hex format for compatibility

talvasconcelos:
Closes #3296

### DIFF
--- a/lnbits/helpers.py
+++ b/lnbits/helpers.py
@@ -18,7 +18,6 @@ from pydantic.schema import field_schema
 from lnbits.jinja2_templating import Jinja2Templates
 from lnbits.settings import settings
 from lnbits.utils.crypto import AESCipher
-from lnbits.utils.nostr import normalize_public_key
 
 from .db import FilterModel
 
@@ -203,10 +202,13 @@ def is_valid_external_id(external_id: str) -> bool:
 
 
 def is_valid_pubkey(pubkey: str) -> bool:
-    key = normalize_public_key(pubkey)
-    if key:
+    if len(pubkey) != 64:
+        return False
+    try:
+        int(pubkey, 16)
         return True
-    return False
+    except Exception as _:
+        return False
 
 
 def create_access_token(data: dict, token_expire_minutes: int | None = None) -> str:
--- a/lnbits/helpers.py
+++ b/lnbits/helpers.py
@@ -18,6 +18,7 @@ from pydantic.schema import field_schema
 from lnbits.jinja2_templating import Jinja2Templates
 from lnbits.settings import settings
 from lnbits.utils.crypto import AESCipher
+from lnbits.utils.nostr import normalize_public_key
 
 from .db import FilterModel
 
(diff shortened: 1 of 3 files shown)


lnbits/helpers.py
```diff
@@ -206,13 +207,10 @@ def is_valid_external_id(external_id: str) -> bool:
 
 
 def is_valid_pubkey(pubkey: str) -> bool:
```
motorina0: this should not be changed, the function is nostr agnostic
the call to this function should be made with the already normalized pubkey

motorina0:
please add a test to tests/api/test_users.py

tests/api/test_users.py
```diff
@@ -380,3 +379,62 @@ async def test_update_superuser_only_allowed_by_superuser(
     )
 
     assert resp.json()["detail"] == "Action only allowed for super user."
+
+
```
motorina0, line 383: please also test for invalid hex or npub

tests/api/test_users.py
```diff
@@ -380,3 +379,85 @@ async def test_update_superuser_only_allowed_by_superuser(
...
+@pytest.mark.anyio
+async def test_create_user_invalid_npub(http_client: AsyncClient, superuser_token):
+    tiny_id = shortuuid.uuid()[:8]
+    invalid_pubkeys = [
+        "npub1flrz7qu87n8y04jwy6r74z44pczcwaesumth08uxrusv4sm7efs83zq8z",
+        "4fc62f0387f4ce47d64e2687ea89f5a8702c3bb98736bbbcf30f906561bf653",
+    ]
+    for invalid_pubkey in invalid_pubkeys:
```
dni: put the invalid pubkeys as decorator `@pytest.mark.parametrize` instead of the `for`

PR #3299, opened 2025-07-31
feat: add batch invoice polling and persistence for StrikeWallet

santyr:
Add batching for invoice polling: Invoice status checks are now performed in batches using a single API call, improving efficiency and reducing the number of requests.

Persist pending invoices across restarts: Pending invoice IDs are saved to disk and reloaded on startup, ensuring no invoice status is lost if the server restarts.


santyr:
Yikes.  I failed to upload my changes....

PR #3300, opened 2025-07-31
feat: add batch invoice polling and persistence for StrikeWallet

santyr:
Add batching for invoice polling: Invoice status checks are now performed in batches using a single API call, improving efficiency and reducing the number of requests.

Persist pending invoices across restarts: Pending invoice IDs are saved to disk and reloaded on startup, ensuring no invoice status is lost if the server restarts.

### DIFF
--- a/lnbits/wallets/strike.py
+++ b/lnbits/wallets/strike.py
@@ -111,7 +111,16 @@ class StrikeWallet(Wallet):
         )
 
         # runtime state
+        import os, json
         self.pending_invoices: list[str] = []  # Keep it as a list
+        # path for persisting pending invoices
+        self.state_path = os.path.join(settings.lnbits_data_folder, "strike_pending_invoices.json")
+        # load persisted pending invoices
+        try:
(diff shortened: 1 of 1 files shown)


motorina0:
CI fails
please run `make all` and fix the reported issues

lnbits/wallets/strike.py
```diff
@@ -111,7 +111,20 @@ def __init__(self) -> None:
         )
 
         # runtime state
+        import json
+        import os
```
dni: imports on the top of the file.

lnbits/wallets/strike.py
```diff
@@ -120,6 +133,15 @@ def __init__(self) -> None:
         self._cached_balance_ts: float = 0.0
         self._cache_ttl = 30  # seconds
 
+    def _persist_pending(self):
+        import json
```
dni: imports on the top of the file.

dni (approved):

PR #3304, opened 2025-08-03
feat: CLN Socket Funding Source

21M4TW:
This is a funding source that uses CLN's socket (Commando). It is based on https://github.com/jb55/lnsocket , and a python wrapper (https://github.com/21M4TW/pylnsocket). It supports multiple CLN runes similar to the CLNRest funding source. It also uses a rune for the CLN xpay command.


motorina0:
@dni regtest

pyproject.toml
```diff
@@ -65,8 +64,10 @@ jsonpath-ng = "^1.7.0"
 pynostr = "^0.6.2"
 python-multipart = "^0.0.20"
 filetype = "^1.2.0"
-nostr-sdk = "^0.42.1"
 bcrypt = "^4.3.0"
+fastapi = "^0.116.1"
+greenlet = "^3.2.3"
+pylnsocket = {git = "https://github.com/21M4TW/pylnsocket"}
```
dni, line 70: if we use the pylnsocket wrapper, it has to have proper pypi releases. are you planning to do those?

dni:
i am a bit confused about `lnsocket`,  do you have to configure anything on the cln node for this to work? and help me understand what are the benefits of this or why would you do this?

21M4TW:
> i am a bit confused about `lnsocket`, do you have to configure anything on the cln node for this to work? and help me understand what are the benefits of this or why would you do this?

No it is the beauty of it, it goes directly through the CLN Lightning port. You only need to create a CLN rune that gives the desired access. Don't need SSL or anything.

pyproject.toml
21M4TW: Yes I can take care of it

motorina0:
Will not be added.

PR #3316, opened 2025-08-14
fix(corelightning): Explicitly set "pay" method in RPC call

santyr:
When paying an invoice, the CoreLightning wallet was making an RPC call without a method name, causing the node to return an "Unknown command ''" error.

This was because the method name was being read from a configurable setting that was not always present. The fix hardcodes the RPC method to "pay", ensuring the call is always correctly formatted and removing the dependency on a potentially missing configuration.

### DIFF
--- a/lnbits/wallets/corelightning.py
+++ b/lnbits/wallets/corelightning.py
@@ -174,7 +174,8 @@ class CoreLightningWallet(Wallet):
                 "description": invoice.description,
             }
 
-            r = await run_sync(lambda: self.ln.call(self.pay, payload))
+            # FIX: Use the explicit "pay" command instead of the configurable self.pay
+            r = await run_sync(lambda: self.ln.call("pay", payload))
 
             fee_msat = -int(r["amount_sent_msat"] - r["amount_msat"])
             return PaymentResponse(


lnbits/wallets/corelightning.py
```diff
@@ -174,7 +174,8 @@ async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse
                 "description": invoice.description,
             }
 
-            r = await run_sync(lambda: self.ln.call(self.pay, payload))
+            # FIX: Use the explicit "pay" command instead of the configurable self.pay
+            r = await run_sync(lambda: self.ln.call("pay", payload))
```
dni: people are using xpay though including me. add an if and check if its empty a string and if its empty use `pay` as default.

santyr: I've updated the code as requested, and enabled xpay on my c-lightning node for testing.

lnbits/wallets/corelightning.py
```diff
@@ -174,7 +174,9 @@ async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse
                 "description": invoice.description,
             }
 
-            r = await run_sync(lambda: self.ln.call(self.pay, payload))
+            # Use configured pay command (e.g. xpay), but default to "pay"
+            pay_command = self.pay or "pay"
```
dni, line 178: ```    corelightning_pay_command: str = Field(default="pay")```
self.pay should never be none, maybe check some like` if self.pay not in ["pay", "xpay", "mpay"]: raise` or create a `self.valid_pay_commands` or even better let the config by a select box instead of a string

PR #3317, opened 2025-08-17
feat: lud11 disposable and storeable payRequests.

dni:
https://github.com/lnurl/luds/blob/luds/11.md

- add stored links to wallet model and frontend page
- change `CreateLnurlPayment` its now optional `res` and `lnurl` you need one of them for the api to work and you also can provide both, if you want to save on the first lnurl request and fully store the lnurl.
- change fetch_lnurl_pay function
- remove dead lnurlauth code

screenshot-1755444823

screen recording here: https://x.com/dnilabs/status/1957324824541278293


arcbtc:
In the screen recording, i noticed a page refresh?
It looked manual. When the invoice is returned from the listener if it has the lnurlp reusable stuff you could run whatever runs on refesh, we want to keep refeshes and need for a refresh out of core.

talvasconcelos (approved):

PR #3324, opened 2025-08-19
UV and Nix flake working 🥳

arcbtc:
uv, uv2nix and our nix flake are working!
```
uv venv
uv sync
uv run lnbits
```
uv2nix is also working so our nix build can keep up with latest version
```
nix build
nix run
```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,88 +1,90 @@
 {
   "nodes": {
+    "build-system-pkgs": {
+      "inputs": {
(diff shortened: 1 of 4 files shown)


arcbtc:
Tests failing, but i think thats a different issue on dev

dni:
cherrypicked nix related commit onto https://github.com/lnbits/lnbits/pull/3325

PR #3325, opened 2025-08-20
feat: use `uv` instead of `poetry` for CI, docker and development

dni:
based on @arcbtc initial pr. https://github.com/lnbits/lnbits/pull/3324

- most of the file diff is caused by `ruff` removing the `Optional` type in this commit. https://github.com/lnbits/lnbits/pull/3325/commits/bf3cbeda10ae7be10c64afb37a699035a8f43ae0 - somehow using `uv` made ruff notice those things

- still keep compatibility for poetry (still used for releasing on pypi)

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ${{ matrix.os-version }}
     steps:
       - uses: actions/checkout@v4
-      - uses: lnbits/lnbits/.github/actions/prepare@feat/uv
+      - uses: lnbits/lnbits/.github/actions/prepare@dev
         with:
           python-version: ${{ inputs.python-version }}
           node-version: ${{ matrix.node-version }}


.github/workflows/make.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     runs-on: ${{ matrix.os-version }}
     steps:
       - uses: actions/checkout@v4
-      - uses: lnbits/lnbits/.github/actions/prepare@dev
+      - uses: lnbits/lnbits/.github/actions/prepare@feat/uv
```
dni: TODO: remove after CI completes successfully

dni:
closes #3324

.github/workflows/make.yml
dni:
```suggestion
      - uses: lnbits/lnbits/.github/actions/prepare@dev
```

PR #3332, opened 2025-08-22
refactor: move from settings.py model to fiat_provider utils

dni:
just to keep `settings.py` and the configuration simpler


motorina0:
I see no need for this PR.

dni:
it only moves the logic bit from settings to utils because settings is already pretty massive and the class FiatSettings got a bit simpler. if you dont like it close it, but i think it a good code cleanup and it was bothering me multiple times at this point so i just cleaned it up a bit.

PR #3333, opened 2025-08-22
fix: start background tasks when activating extensions at runtime

echennells:
When extensions are activated after boot (not during startup), their background tasks were not being started. This caused extensions to only partially work - HTTP endpoints would function but background processing (like payment listeners) would not run.

This fix adds the missing call to start_extension_background_work() in the activate_extension function, ensuring extensions activated at runtime have full functionality.


motorina0:
Thanks!
Looks like a valid issue, I will test it.

dni (approved):

PR #3335, opened 2025-08-26
feat: add copy, download and write NFC to qrcode components

dni:
- add write nfc when available enable it with `nfc="true"`

screenshot-1756228109
screenshot-1756228082
mobile: 
photo_2025-08-26_19-23-39

### DIFF
--- a/lnbits/templates/components.vue
+++ b/lnbits/templates/components.vue
@@ -1104,18 +1104,18 @@
               </div>
               <div class="row q-mt-md">
                 <q-btn
-                  v-close-popup
-                  flat
-                  color="grey"
-                  class="q-mr-auto"
-                  :label="$t('close')"
-                ></q-btn>
-                <q-btn
                   outline
                   color="grey"
                   @click="checkPayment(props.row.payment_hash)"
                   icon="refresh"
                   :label="$t('payment_check')"
+                ></q-btn>
+                <q-btn
+                  v-close-popup
+                  flat
+                  color="grey"
+                  class="q-ml-auto"
+                  :label="$t('close')"
                 ></q-btn>
               </div>
             </q-card-section>


lnbits/templates/components.vue
```diff
@@ -631,6 +632,9 @@
       class="qrcode__image"
       :src="custom.logo"
       alt="qrcode icon"
+      @click="downloadSVG"
```
talvasconcelos: some QR codes are wrapped in an <a> tag!

won't this cause issues?

dni: now there if `:href=""` where you can pass a url

lnbits/static/js/components/lnbits-qrcode.js
```diff
@@ -9,19 +9,123 @@ window.app.component('lnbits-qrcode', {
       type: String,
       required: true
     },
-    options: Object
+    nfc: {
+      type: Boolean,
+      default: false
+    },
+    buttons: {
```
motorina0: ambiguous naming

motorina0:
who need the SVG of the QR code?

lnbits/static/js/components/lnbits-qrcode.js
dni: yes, its the buttons on the qrcode components, what wrong with that? `<div v-if="buttons" class="qrcode__buttons row q-gutter-x-sm">`

dni:
> who need the SVG of the QR code?

everyone who wants to do something like a flyer, promotions or just want to print out a qrcode with those sticker printers.

lnbits/static/js/components/lnbits-qrcode.js
motorina0: > its the buttons on the qrcode components
 - it is not clear what those buttons are and what they do
 - it should be boolean-like name, something like `show-buttons`

motorina0:
These buttons look unacceptably ugly:

screenshot-1756228082

lnbits/static/js/components/lnbits-qrcode.js
dni: good catch!

dni:
screenshot-1756802319

PR #3339, opened 2025-09-01
[feat] Extension Builder

motorina0:
### Extension Builder

Go to `Extensions` -> `Create Extension`
image

The wizard for building a basic extension will be presented:

image


pyproject.toml
```diff
@@ -250,6 +250,7 @@ classmethod-decorators = [
 # TODO: remove S101 ignore
 "lnbits/*" = ["S101"]
 "lnbits/core/views/admin_api.py" = ["S602", "S603", "S607"]
+"lnbits/core/services/extensions_builder.py" = ["S701"]
```
dni, line 253: what is `S701` ? maybe add a comment

dni:
ext_id and ext_name should be required on the first page and you should not be allowed to advance in the form.
screenshot-1758780037

dni:
screenshot-1758780408
> The owner of the extension can view this data, `but should not modify it.`
if he should not modify it why is there an edit and delete button. i think the comment is a bit off

dni:
is it possible to have the client data optional? maybe that will also immediately get rid of the public page. i see some use cases where i only need the ownerdata

dni:
screenshot-1758780824
i think public page title and public page description should be plain textfields, so you can enter what you want or if empty use the defaults from lnbits

lnbits/core/views/extension_api.py
```diff
@@ -115,6 +121,120 @@ async def api_install_extension(data: CreateExtension):
...
 
 
+@extension_router.post(
+    "/builder/zip",
+    summary="Build and download extension zip.",
+    description="""
+        This endpoint generates a zip file for the extension based on the provided data.
+    """,
+    dependencies=[Depends(check_user_exists)],
+    response_model=None,
```
dni:
```suggestion
```
the response is a `FileResponse` no need to set it to none here

lnbits/core/views/extension_api.py
```diff
@@ -115,6 +121,120 @@ async def api_install_extension(data: CreateExtension):
...
+
+
+@extension_router.post(
+    "/builder/deploy",
+    summary="Build extension based on provided config.",
+    description="""
+        This endpoint generates a zip file for the extension based on the provided data.
+        If `deploy` is set to true, the extension will be installed and activated.
+    """,
+    response_model=SimpleStatus,
```
dni:
```suggestion
```
redundant fastapi inherits from return type

lnbits/core/views/extension_api.py
```diff
@@ -115,6 +121,120 @@ async def api_install_extension(data: CreateExtension):
...
+        user_ext.active = True
+    await update_user_extension(user_ext)
+
+    return SimpleStatus(success=True, message=f"Extension '{data.id}' deployed.")
+
+
+@extension_router.post(
+    "/builder/preview",
+    summary="Build and preview the extension ui.",
+    response_model=SimpleStatus,
```
dni:
```suggestion
```

lnbits/core/views/extension_api.py
```diff
@@ -115,6 +121,120 @@ async def api_install_extension(data: CreateExtension):
...
+
+
+@extension_router.delete(
+    "/builder",
+    summary="Clean extension builder data.",
+    description="""
+        This endpoint cleans the extension builder data.
+    """,
+    dependencies=[Depends(check_admin)],
+    response_model=SimpleStatus,
```
dni:
```suggestion
```

motorina0:
> is it possible to have the client data optional? maybe that will also immediately get rid of the public page. i see some use cases where i only need the ownerdata

possible, but not in this iteration

motorina0:
> ext_id and ext_name should be required on the first page and you should not be allowed to advance in the form.

It is ok for the user to see the other pages. The message is there. He can not `Preview` or `Deploy`.

motorina0:
> screenshot-1758780824 i think public page title and public page description should be plain textfields, so you can enter what you want or if empty use the defaults from lnbits

lnbits/core/views/extension_api.py
motorina0: `response_model=None` is required, otherwise there is an error at server startup

lnbits/core/views/extension_api.py
motorina0: `response_model=None` is required, otherwise there is an error at server startup

lnbits/core/views/extension_api.py
dni: not for me... i tried it and all other endpoints also dont use it.

motorina0: fixed

talvasconcelos (approved):
Tested. :star_struck: 

Failing jmeter test

PR #3351, opened 2025-09-05
chore: qr button tweak (quasar classes should always be used)

arcbtc:
buttons also need to be a block to avoid elements leaking into them


motorina0:
- lint fails
- add a screenshot (before & after) please

arcbtc:
result is the same. will fix lint

motorina0:
Fixed lint.
@arcbtc  Can this be out of draft?

dni (approved):

PR #3363, opened 2025-09-16
FIX:  Flake and Module for NixOS deployment

PatMulligan:
# Fix LNBits NixOS Service Module for Python Virtual Environment Package

## Problem

The current LNBits NixOS service module (`nix/modules/lnbits-service.nix`) has several issues that prevent it from working with the flake-built Python virtual environment package:

1. **Missing `src` attribute**: The module expects `cfg.package.src` but the flake builds a Python virtual environment package that doesn't have a `src` attribute
2. **Incorrect binary path**: `lib.getExe` resolves to the wrong binary name (`lnbits-env` instead of `lnbits`)
3. **Wrong working directory**: The service needs to run from where the Python package's static files and templates are located

## Solution

This PR fixes the service module to work correctly with the virtual environment package structure produced by the flake.

### 1. Removed dependency on non-existent `src` attribute
```diff
- LNBITS_PATH = "${cfg.package.src}";
```
The `LNBITS_PATH` environment variable was removed as it references a non-existent `src` attribute.

### 2. Fixed working directory to point to installed Python package
```diff
- WorkingDirectory = "${cfg.package.src}";
+ WorkingDirectory = "${cfg.package}/lib/python3.12/site-packages";
```
LNBits needs to run from the directory containing its static files and templates. In the virtual environment package, these are located in the site-packages directory.

### 3. Fixed binary path resolution
```diff
- ExecStart = "${lib.getExe cfg.package} --port ${toString cfg.port} --host ${cfg.host}";
+ ExecStart = "${cfg.package}/bin/lnbits --port ${toString cfg.port} --host ${cfg.host}";
```
The `lib.getExe` function was incorrectly resolving to `lnbits-env` (the package name) instead of the actual `lnbits` binary. Using the explicit path ensures the correct binary is executed.

### 4. Fixed StateDirectory to use relative path
```diff
- StateDirectory = "${cfg.stateDir}";
+ StateDirectory = "lnbits";
```
Changed to use the relative path "lnbits" which systemd will automatically create under `/var/lib/`.

### 5. Proper data folder structure
```diff
+ systemd.tmpfiles.rules = [
+   "d ${cfg.stateDir}                            0700 ${cfg.user} ${cfg.group} - -"
+   "d ${cfg.stateDir}/data                       0700 ${cfg.user} ${cfg.group} - -"
+ ];

  environment = lib.mkMerge [
    {
-     LNBITS_DATA_FOLDER = "${cfg.stateDir}";
-     LNBITS_EXTENSIONS_PATH = "${cfg.stateDir}/extensions";
-     LNBITS_PATH = "${cfg.package.src}";
+     LNBITS_DATA_FOLDER = "${cfg.stateDir}/data";
+     # LNBits automatically appends '/extensions' to this path
+     LNBITS_EXTENSIONS_PATH = "${cfg.stateDir}";
    }
```
Fixed the folder structure to properly separate data and extensions:
- Data files (database.sqlite3, images/, logs/, upgrades/) now go in `${stateDir}/data`
- Extensions go in `${stateDir}/extensions` (LNBits automatically appends `/extensions` to the path)
- Added tmpfiles rules to ensure directories are created with proper permissions
- Note: `LNBITS_EXTENSIONS_PATH` is set to `${stateDir}` because LNBits internally appends `/extensions` to this path

## Testing

The fixed service module has been tested in a NixOS VM with the following results:

✅ Service starts successfully
✅ Web interface accessible at configured port
✅ Database migrations run correctly
✅ Static files and templates load properly

### Test Configuration Used
```nix
services.lnbits = {
  enable = true;
  host = "0.0.0.0";
  port = 5000;
  openFirewall = true;
  env = {
    LNBITS_ADMIN_UI = "true";
  };
};
```

### Verification Commands
```bash
# Service status
systemctl status lnbits
# ● lnbits.service - lnbits
#      Active: active (running)

# Web interface test
curl -I http://localhost:5000/
# HTTP/1.1 307 Temporary Redirect
# location: /first_install

# Process verification
ps aux | grep lnbits
# python3.12 /nix/store/.../bin/lnbits --port 5000 --host 0.0.0.0

# Verify folder structure
ls -la /var/lib/lnbits/
# drwx------ 5 lnbits lnbits 4096 data
# drwxr-xr-x 2 lnbits lnbits 4096 extensions

ls -la /var/lib/lnbits/data/
# -rw-r--r-- 1 lnbits lnbits 143360 database.sqlite3
# drwxr-xr-x 2 lnbits lnbits   4096 images
# drwxr-xr-x 2 lnbits lnbits   4096 logs
# drwxr-xr-x 2 lnbits lnbits   4096 upgrades
```

## Impact

This fix enables users to:
- Successfully deploy LNBits on NixOS using the flake
- Use the NixOS service module for production deployments
- Configure LNBits declaratively through NixOS configuration

## Backwards Compatibility

These changes maintain backwards compatibility with existing configurations. The service interface remains unchanged - only the internal implementation is fixed to work with the virtual environment package structure.

## Related Issues

This fixes the error encountered when using the flake's NixOS module:
```
error: attribute 'src' missing
at /etc/nixos/sources/lnbits/nix/modules/lnbits-service.nix:111:31
```

## Additional Notes

The Python version (3.12) is currently hardcoded in the working directory path. A future improvement could make this dynamic based on the Python version used by the package.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -139,9 +139,10 @@
             replaceVars = prev.replaceVars or (path: vars: prev.substituteAll ({ src = path; } // vars));
           };
 
+          # System-specific nixos modules to avoid circular dependency
           nixosModules.default = { pkgs, lib, config, ... }: {
             imports = [ "${./nix/modules/lnbits-service.nix}" ];
-            nixpkgs.overlays = [ self.overlays.default ];
+            nixpkgs.overlays = [ self.overlays.${system}.default ];
           };
 
           checks = { };
--- a/nix/modules/README.md
+++ b/nix/modules/README.md
@@ -1,0 +1,241 @@
+# LNBits NixOS Installation Guide
+
+This guide shows how to install LNBits on a fresh NixOS system.
+
+## Quick Start (Recommended)
+
+Add this to your NixOS configuration (`/etc/nixos/configuration.nix`):
+
+```nix
+{ config, lib, pkgs, ... }:
+
+let
+  lnbitsFlake = builtins.getFlake "github:lnbits/lnbits";
+in
+{
+  imports = [
+    # Import LNBits service module directly from GitHub
+    "${lnbitsFlake}/nix/modules/lnbits-service.nix"
+  ];
+
+  # Enable flakes (required)
+  nix.settings.experimental-features = [ "nix-command" "flakes" ];
+
+  # Configure LNBits service
+  services.lnbits = {
+    enable = true;
+    host = "0.0.0.0";        # Listen on all interfaces
+    port = 5000;             # Default port
+    openFirewall = true;     # Open firewall port automatically
+
+    # Use package from the same flake (adjust system architecture as needed)
+    package = lnbitsFlake.packages.x86_64-linux.lnbits;
+
+    env = {
+      LNBITS_ADMIN_UI = "true";
+      # Configure your Lightning backend:
+      # LNBITS_BACKEND_WALLET_CLASS = "LndRestWallet";
+      # LND_REST_ENDPOINT = "https://localhost:8080";
+      # LND_REST_CERT = "/path/to/tls.cert";
+      # LND_REST_MACAROON = "/path/to/admin.macaroon";
+    };
+  };
+
+  # Rebuild and switch
+  # sudo nixos-rebuild switch
+}
+```
+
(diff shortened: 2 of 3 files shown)


arcbtc:
Tested and works. I used these install guide tweaks, would they need to be different?
https://github.com/lnbits/lnbits/pull/3361/files

dni (approved):

PR #3364, opened 2025-09-16
feat: stripe recurring payments

arcbtc:
We will use for our saas payments, but another interesting use case could be for dca on sellcoins extension.
Tested for tap-to-pay, payment link and recurring payment link.


motorina0:
Closed in favor of: https://github.com/lnbits/lnbits/pull/3369

PR #3367, opened 2025-09-24
ci: lock boltz client to latest working version

dni:


Dockerfile.boltz
```diff
@@ -1,4 +1,4 @@
-FROM boltz/boltz-client:latest AS boltz
+FROM boltz/boltz-client:2.8.3 AS boltz
```
michael1011, line 1: `latest` has long been fixed: https://github.com/BoltzExchange/boltz-client/pull/540
`2.8.3` is already 5 releases behind

dni: should we still lock to one release just in case latest breaks or has breaking changes? 
what would be the correct version for that to not be surprised?

kilrau: > should we still lock to one release just in case latest breaks or has breaking changes?

No. There won't be breaking changes and latest is fine.

> what would be the correct version for that to not be surprised?

Latest.

kilrau:
Please move back to latest @dni 

When you opened this PR it was already fixed btw

PR #3369, opened 2025-10-01
[feat] Stripe subscription

motorina0:


lnbits/core/views/callback_api.py
```diff
@@ -33,3 +41,104 @@ async def api_generic_webhook_handler(
         success=False,
         message=f"Unknown fiat provider '{provider_name}'.",
     )
+
+
+async def handle_stripe_event(event: dict):
```
dni, line 46: i think this code should be somewhere under `services` not `views`

dni: maybe `services/callback.py` ?

motorina0: `services/callback.py` doesn;t say much. Any other suggestion?

Issue #3365, opened 2025-09-19
[BUG] Extension “New Version” display error


AxelHamburch:
I am using LNbits version v1.2.1, and the Admin UI shows under Extensions that there is a new version available. 

<img width="548" height="233" alt="Image" src="https://github.com/user-attachments/assets/61f015b9-a0ba-4c17-b036-9235e1eb3506" />

However, I can't install it because it's only for version 1.3.0.

<img width="505" height="354" alt="Image" src="https://github.com/user-attachments/assets/23765fe8-4b8a-4a39-8b84-faf58b0ec2f2" />

It's not great that I'm being told I have a new version that I can't install, but it's still manageable.

However, the notification is also displayed under Extension Management. There, 11 possible updates are listed. 

<img width="1197" height="606" alt="Image" src="https://github.com/user-attachments/assets/e8f16547-5d2c-4ab2-b145-a1f02d0e5f5f" />

When I try to update, I get a series of error messages. 

<img width="520" height="768" alt="Image" src="https://github.com/user-attachments/assets/cbd6ac10-6256-4dc3-86ee-ffbadc32ad99" />

That shouldn't happen. Why is it trying to update when the extension versions don't match my LNbits version?

But what's even worse is that errors occur during the attempted installation.

<img width="966" height="427" alt="Image" src="https://github.com/user-attachments/assets/e0433966-e549-4815-a00c-b953bf21738b" />

These errors then cause subsequent errors, so that no extension can be called up and only this screen is visible.

<img width="850" height="476" alt="Image" src="https://github.com/user-attachments/assets/4f5f89dd-eb20-4a71-a8c7-dd5dbb8e4ca3" />

The logs show:

<img width="1652" height="234" alt="Image" src="https://github.com/user-attachments/assets/23e82f7b-9189-4283-8f14-8d2884f341b0" />

Only restarting the server fixes the error. However, something strange can be seen at the beginning of the logs:

<img width="1224" height="593" alt="Image" src="https://github.com/user-attachments/assets/f893efe4-e958-453f-ba2b-e511daea9de3" />

If I am on version v1.2.1, why am I getting an extension update notification for updates that are not even valid for my version, and more importantly, why is it trying to perform updates and then crashing my extensions?

I've noticed this error before. When we jumped from v0.12.12 to v1.0, there was a similar phenomenon. It tried to perform an update, but it didn't work. However, I can't remember if anything broke in the process.

motorina0:
Thanks. I will check this.

[...]

PR #3385, opened 2025-10-06
feat:  show badge if an extension has paid releases

motorina0:
image


motorina0:
Also fixes: https://github.com/lnbits/lnbits/issues/3365

lnbits/core/templates/core/extensions.html
```diff
@@ -137,8 +137,40 @@
               @click="showExtensionDetails(extension.id, extension.details_link)"
               v-text="extension.name"
             ></div>
-            <div>
+            <div style="justify-content: space-between; display: flex">
```
talvasconcelos, line 140: i think there's a quasar class for these styles

motorina0: which one?
I would have used it.

PR #3390, opened 2025-10-07
feat: allow query param for qrcode

motorina0:
Allow for URLs to be shown in the QR code

Before this returned 404:
http://localhost:5000/api/v1/qrcode/https%3A%2F%2Fcheckout.stripe.com

After, this can be used:
http://localhost:5000/api/v1/qrcode?data=https%3A%2F%2Fcheckout.stripe.com

### DIFF
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -115,7 +115,7 @@ async def api_fiat_as_sats(data: ConversionData):
 
 @api_router.get("/api/v1/qrcode", response_class=StreamingResponse)
 @api_router.get("/api/v1/qrcode/{data}", response_class=StreamingResponse)
-async def img(data):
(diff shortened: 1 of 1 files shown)


lnbits/core/views/api.py
```diff
@@ -113,6 +113,7 @@ async def api_fiat_as_sats(data: ConversionData):
         return output
 
 
+@api_router.get("/api/v1/qrcode", response_class=StreamingResponse)
 @api_router.get("/api/v1/qrcode/{data}", response_class=StreamingResponse)
 async def img(data):
```
dni: can we add a endpoint description here? its not very clear from the code that an url parameter or a query parameter can be used.

```suggestion
async def img(data: str):
```

PR #3414, opened 2025-10-16
refactor: move `/payments` into vue component

dni:
first step as an attempt to decouple the frontend more from the backend. as talked with @talvasconcelos. this gets rid of the DynamicComponent loading of the payments page because there is nothing that has to be rendered on the server side for it to work.

- create  generic empty route with just `base.html` for reloads
- move into /static/js/pages / static/templates/pages


arcbtc:
Wont this get bloaty rather than dynamically loading, adding all the components to package.json?

dni:
> Wont this get bloaty rather than dynamically loading, adding all the components to package.json?

this improves the speed noticeable for all users, and its only a couple of pages currently. the first page load will be minutly slower but the navigation around the site will be faster because it saves 1 ajax call for the html on each navigation.

in the future the quasar app could have a build step and those components and templates can get lazy loaded, but for now this is just a cleanup that get us into that direction step by step. then we have both speed in the ui + speed on the first load.

lnbits/templates/base.html
```diff
@@ -424,8 +424,8 @@ <h6 class="q-my-none ellipsis full-width">
         window[key] = WINDOW_SETTINGS[key]
       })
     </script>
-    {% include('components.vue') %} {% block vue_templates %}{% endblock %} {%
-    for url in INCLUDED_JS %}
+    {% include('components.vue') %} {% include('pages/payments.vue') %} {% block
```
talvasconcelos, line 487: this will include the payments component on every page that extends base.html ?

dni: yes, that is done once instead of everytime we do an ajax

motorina0 (approved):

PR #3419, opened 2025-10-16
feat: add verify preimage form to wallet page

dni:
- added hex functions to js utils
screenshot-1760622600


motorina0:
Concept NACK
Why is this feature on the wallet page?

Issue #3426, opened 2025-10-18
[BUG] Deleted wallet can be used for transactions


lightningcheckout:
**Describe the bug**
We are running an instance of LNbits with the Partytap addon enabled. A user deleted his wallet but not his partytap devices which was linked to it. He continued using it, which resulted in transactions being processed by LNbits, but the sats were 'gone' from a user perspective. 

Checkin as an admin, I saw the transactions registered under the deleted wallet. 

**Expected behavior**
I would expected that a deleted wallet will never accept a transaction to it.

**Desktop (please complete the following information):**
- LNbits version: 1.2.1 
- Database: Postgres

lightningcheckout:
By 'undelete' the wallet the balance of the wallet showed up again with the correct value and became visible for the user.

motorina0:
I cannot reproduce this issue on LNbits core (by calling the LNbits API).
I am not aware of the `PartyApp` extension, that is not an LNbits vetted extension. 
Is the source code public? The issue might be there.

Trying to pay an invoice from a deleted wallet return 404.
**Request:**
```curl
curl --location 'http://localhost:5000/api/v1/payments' \
--header 'x-api-key: 598962f1b75342d5b39c306854aca202' \
--header 'Content-Type: application/json' \
--data '{"out":true,"bolt11":"lnbc1u1p50xfyrpp5kspd3hcv2tn8z2uq60dt0x56d2v9kxmsaqhm0wn23u0ueeq0rp4qcqzyssp5ra0p5nllfzsdp2rtu38q5hrufqczxyjjfy3u8xmjdtsd5u3aq8ws9q7sqqqqqqqqqqqqqqqqqqqsqqqqqysgqdq2f38xy6t5wvmqz9gxqrrssrzjqwryaup9lh50kkranzgcdnn2fgvx390wgj5jd07rwr3vxeje0glclll4ttz7sp6kpvqqqqlgqqqqqeqqjqmwmu3nk96dxqweq2vnpcj5ugmex07ah0yzlta88p62nf99kfy7ekygggfutqv9zw5g8gpl00c3gk0q0f63dvnnj5v6py9hp6qcdslxsqw3my84"
}'
```

**Response:**
```json
{
    "detail": "Wallet not found."
}
```

lightningcheckout:
Sorry, maybe my descrption wasn't clear enough. But I meant that a user is able to receive funds in his 'deleted' wallet. 

[...]

PR #3439, opened 2025-10-20
fix: exclude (soft) deleted wallets when creating an invoice

motorina0:
Reported here: https://github.com/lnbits/lnbits/issues/3426

The `deleted` state was correctly checked for API calls.
However for the lower level `services/payments` the `deleted` flag for getting the wallet was relaxed and it included deleted wallets.
This allowed some extensions to still receive payments after the wallet was `soft` deleted.


talvasconcelos (approved):
Checks are failing

DoktorShift:
Tested as mentioned to reproduce here: https://github.com/lnbits/lnbits/issues/3426#issuecomment-3418320195

It failed. Means PR is working.

PR #3454, opened 2025-10-29
fix: add dependency-groups get rid of warning

dni:
getting rid of this deprecation warning: 
```
Oct 28 14:57:34 mynode lnbits[57529]: warning: The `tool.uv.dev-dependencies` field (used in `pyproject.toml`) is deprecated and will be removed in a future release; use `dependency-groups.dev` instead
```


motorina0:
Linting fails.

motorina0:
LGTM

PR #3455, opened 2025-10-29
feat: NWC use coincurve instead of secp

dni:


lnbits/utils/nostr.py
```diff
@@ -105,24 +105,22 @@ def verify_event(event: dict) -> bool:
     if event_id != event["id"]:
         return False
     pubkey_hex = event["pubkey"]
-    pubkey = secp256k1.PublicKey(bytes.fromhex("02" + pubkey_hex), True)
-    if not pubkey.schnorr_verify(
-        bytes.fromhex(event_id), bytes.fromhex(event["sig"]), None, raw=True
-    ):
+    pubkey = coincurve.PublicKeyXOnly(bytes.fromhex(pubkey_hex))
+    if not pubkey.verify(bytes.fromhex(event["sig"]), bytes.fromhex(event_id)):
```
motorina0, line 109: `verify` and `schnorr_verify` are two different functionalities
Does `pubkey.verify` do schnorr verify by default (because its created with `PublicKeyXOnly`)?

lnbits/utils/nostr.py
```diff
@@ -19,22 +19,22 @@ def generate_keypair() -> tuple[str, str]:
 
 
 def encrypt_content(
```
motorina0, line 21: nice to have: a small unit test for `encrypt_content` and `decrypt_content`

lnbits/utils/nostr.py
dni: yes, i had to look up the docs myself thats why i use PublicKeyXOnly.
doc links:

verify for normal public key using ECDSA: https://ofek.dev/coincurve/api/#coincurve.PublicKey.verify
verify for pub key XOnly using schnorr: https://ofek.dev/coincurve/api/#coincurve.PublicKeyXOnly.verify

not 100% why that is like that, but it works :)

motorina0 (approved):
LGTM

PR #3409, opened 2025-10-13
fix websocket scope

talvasconcelos:
#### Issue
The original `onInvoicePaid` returned `ws.onclose`, which was ineffective:
- If `ws.onclose` was `null` (default case), calling the returned value (`cleanup()`) threw a `TypeError: cleanup is not a function`.
- If `ws.onclose` was a function, it didn’t close the WebSocket, leaving connections open and contributing to global scope pollution (`window.ws`).
- This caused potential memory leaks and cross-tab interference when multiple WebSocket instances overwrote `window.ws`.

#### Testing Scenarios
The issue (#3308) occurs randomly (e.g., multiple tabs or long-running tabs), so testing focuses on preventing global scope pollution, ensuring proper cleanup, and verifying WebSocket behavior.

1. **No Global Scope Pollution**:
   - Open http://127.0.0.1:5000, log in, and navigate to a wallet page (`/wallet/<id>`).
   - Create an invoice to trigger `onInvoicePaid`.
   - In DevTools Console, type `window.ws` and press Enter.
   - **Expected**: `undefined` (no global `ws`).
   - Test in 2-3 tabs: Create invoices in each; confirm `window.ws` is always `undefined`.
2. **Payment Handling**:
   - Create an invoice and pay it (e.g., via another LNbits wallet or mobile wallet).
   - **Expected**: UI updates (invoice marked "Paid", confetti, etc...).
3. **WebSocket Error Handling**:
   - On wallet page, toggle network offline (DevTools → Network → Offline).
   - **Expected**: Console logs "WebSocket error, reconnecting...". No new connection until network is restored (handled by `paymentEvents`).
4. **Server-Side Closure**:
   - Restart server, load wallet page, wait 10s.
   - **Expected**:  Console logs "WebSocket closed: code=XXX, reason=XXX" (e.g., code 4001).

#### Notes
- The original `return ws.onclose` was non-functional, causing cleanup failures and contributing to global scope issues.
- `paymentEvents` now uses `Map` to efficiently track and clean up WebSocket listeners, preventing duplicates.
- Reconnect logic is managed in `paymentEvents` (not `onInvoicePaid`) to centralize wallet listener management.
- The issue’s randomness (multiple tabs, long-running tabs) is mitigated by proper cleanup and `Map`-based tracking.

PR #3459, opened 2025-10-30
Fix: Websocket reconnection and UI notification

talvasconcelos:
### What changed

* Replaced the single global `ws` variable with **per-wallet WebSocket instances** to prevent cross-wallet interference and dropped connections.
* Added proper **`onopen`**, **`onerror`**, and **`onclose`** handlers with structured console logging.
* Implemented **automatic reconnect** every 5 s when a connection is lost.
* Introduced a **retry cap** (12 attempts ≈ 60 s) to avoid infinite reconnect loops if the server is down.
* Added a global flag **`g.connectionWarning`** and a **`g.disconnectedWallets`** set to track connection status across wallets.
* Updated the `mounted()` hook to include a **periodic health check** (every 60 s) that re-triggers `paymentEvents()` if any wallets remain disconnected.
* No changes to existing balance or payment update logic — only connection management was refactored.
* Added UI badge and notification for connection lost or restored

Got it — you just need a short, developer-friendly test plan that could go in the PR description or review notes.
Here’s a concise, realistic version that matches how LNbits devs usually write them:

---

### Steps to reproduce the issue

1. Launch LNbits and log in with a user that has **two or more wallets**.
2. Open the browser console to watch WebSocket activity.
3. Leave the browser tab open for some time, or switch between wallets/pages several times (issue is hard to reproduce)
4. Stop or restart the LNbits server (or disconnect the network) to simulate a lost WebSocket connection.
5. Observe that:
   * WebSocket connections sometimes stop receiving events.
   * The global `ws` variable is reused incorrectly between wallets.
   * No reconnection occurs, sometimes, after connection loss.
   * There is no user feedback when connections are down.

---

### Steps to test the fix

1. Run LNbits with this branch and log in with **multiple wallets**.
2. Confirm that each wallet opens its own WebSocket connection in the console.
3. Simulate a connection loss:
   * Stop the server, block the network, or close/reopen the browser tab.
4. Verify that:
   * Each wallet reconnects automatically every 5 s.
   * After 60 s or 12 failed attempts, retries stop with a clear log message.
   * When any wallet is disconnected, `g.connectionWarning` becomes `true`; it resets to `false` once all are reconnected.
   * Notifications on the UI indicate lost connection.
5. Restart the server or restore the network:
   * Connections are re-established automatically (within 60 s).
   * Incoming payments still update balances correctly.
   * Notification about reconnection / badge disappears

Closes #3409 
Closes #3308


lnbits/static/js/base.js
```diff
@@ -179,14 +179,79 @@ window.LNbits = {
...
-      ws = new WebSocket(`${websocketUrl}/${wallet.inkey}`)
-      ws.onmessage = ev => {
-        const data = JSON.parse(ev.data)
-        if (data.payment) {
-          cb(data)
+      const reconnectInterval = 5000 // 5 seconds
+      const maxRetries = 12 // ~60 seconds
+      const websocketUrlFull = `${websocketUrl}/${wallet.inkey}`
+
+      if (!g.activeWebsockets) g.activeWebsockets = {}
```
dni, line 186: why not using `data()` variable of the component instead of this magic `g` variable?

lnbits/static/js/base.js
```diff
@@ -179,14 +179,79 @@ window.LNbits = {
...
+      if (!g.disconnectedWallets) g.disconnectedWallets = new Set()
+
+      const updateConnectionWarning = () => {
+        g.connectionWarning = g.disconnectedWallets.size > 0
+      }
+
+      const connect = (retryCount = 0) => {
```
dni, line 193: this should be a method of the component, not a variable inside an event callback function

lnbits/static/js/base.js
```diff
@@ -494,7 +561,8 @@ window.windowMixin = {
       bgimageChoice: this.$q.localStorage.has('lnbits.backgroundImage')
         ? this.$q.localStorage.getItem('lnbits.backgroundImage')
         : USE_DEFAULT_BGIMAGE,
-      ...WINDOW_SETTINGS
+      ...WINDOW_SETTINGS,
+      silentConnectionCleanup: false
```
dni, line 565: what is this variable and why is it in windowMixin?

lnbits/static/js/base.js
```diff
@@ -802,6 +870,57 @@ window.windowMixin = {
...
+
+  beforeUnmount() {
+    this.silentConnectionCleanup = true
+    // Prevent background timers when component is destroyed
+    if (this.connectionRetryTimer) {
+      clearInterval(this.connectionRetryTimer)
+      this.connectionRetryTimer = null
+    }
+  },
+  watch: {
```
dni, line 892: i dont think, watching a variable is good here, you know when its disconnected and just should send the notifications / set variables there instead of creating a variable and watching it.

lnbits/static/js/base.js
talvasconcelos: where is the `data()` of the component?

lnbits/static/js/base.js
talvasconcelos: So we don't trigger notifications when switching pages! 

hmm... maybe the right place to put it, i agree

dni:
will soon open a new pr

PR #3460, opened 2025-10-31
 refactor: move /node into vue component

dni:
follow up on https://github.com/lnbits/lnbits/pull/3414

1. instead of checking if node public is enabled on the generic route, its done on the api request and the node public button isnt shown, if you still navigate to there you will see this message, after a request which will return `503 not enabled`
screenshot-1761895644

2. there is no need to check to node status on the generic at `/node` the status will be checked anyways by lnbits itself and on the api request getting the node data


lnbits/core/templates/empty_public.html
```diff
@@ -0,0 +1,3 @@
+{% extends "public.html" %} {% from "macros.jinja" import window_vars with
+context %} {% block page %} {% endblock %} {% block scripts %} {{
+window_vars(user) }} {% endblock %}
```
motorina0:
```suggestion
 }} {% endblock %}
```

why does it need `window_vars(user)` if its public?

lnbits/static/js/pages/node-public.js
```diff
@@ -0,0 +1,74 @@
+window.PageNodePublic = {
```
motorina0, line 1:
```suggestion
window.pages.PageNodePublic = {
```

maybe its a good idea not to put these pages in the global scope

dni: no

lnbits/core/templates/empty_public.html
dni: open issue

PR #3461, opened 2025-10-31
 refactor: move /audit into vue component

dni:


lnbits/core/views/audit_api.py
```diff
@@ -25,6 +28,11 @@
 async def api_get_audit(
     filters: Filters = Depends(parse_filters(AuditFilters)),
 ) -> Page[AuditEntry]:
+    if not settings.lnbits_audit_enabled:
```
motorina0: `lnbits_audit_enabled` is a flag for collecting data
if disabled it should not prevent users from seeing previously collected data

motorina0 (approved):

PR #3466, opened 2025-10-31
refactor: move /admin into vue component

dni:
- rename empty -> index


lnbits/templates/pages/admin.vue
```diff
@@ -0,0 +1,3167 @@
+<template id="page-admin">
```
motorina0, line 1: this file is now huge

PR #3468, opened 2025-11-01
Fix LND payment fee updates after settlement

ziggie1984:
## Summary

Fixes three issues preventing proper fee reconciliation for LND backend payments:

1. **Streaming API returns too early** - Both gRPC and REST implementations exit on first status update instead of waiting for terminal state
2. **Wrong fee field in gRPC** - Uses `htlcs[-1].route.total_fees_msat` instead of top-level `fee_msat`
3. **Deprecated REST API** - Documents that `/v1/channels/transactions` should be migrated to `/v2/router/send`

## Problem

When paying invoices with LND backends, LNbits reserves a 1% routing fee upfront. After payment settlement, this fee should be updated to the actual routing cost. However, **fees remain stuck at the 1% reserve** for most payments.

### Root Causes

**Issue #1: Streaming APIs return too early**

Both `TrackPaymentV2` (gRPC) and `/v2/router/track` (REST) stream multiple updates as a payment progresses:
- `NON_EXISTENT` / `UNKNOWN` → `IN_FLIGHT` → `SUCCEEDED` / `FAILED`

The code was returning on the **first update**, often while `IN_FLIGHT`, never receiving the final update with actual fees. This meant:
- `get_payment_status()` returns `PaymentPendingStatus()` on first IN_FLIGHT update
- Background task `check_pending_payments()` never updates the payment to SUCCESS
- Fee remains at the reserved 1% forever

**Issue #2: Wrong gRPC fee field**

Code accessed `payment.htlcs[-1].route.total_fees_msat` which:
- Only shows fee from last HTLC (wrong for multi-path payments)
- May use wrong attempt's fee (multi-attempt scenarios)
- Doesn't aggregate properly across all payment paths

Should use `payment.fee_msat` which correctly aggregates all fees.

**Issue #3: Deprecated REST endpoint**

`/v1/channels/transactions` is the legacy Lightning RPC API. Modern `/v2/router/send` (Router RPC) provides better MPP support and fee reporting, and is actively maintained.

## Impact

**Before:**
- Payments > 5 seconds (most real-world payments): ❌ Fee stuck at 1%
- Payments < 5 seconds: ✅ Fee correct
- Multi-path payments (gRPC): ❌ Wrong fee calculation

**After:**
- All payments: ✅ Fee correctly updated to actual routing cost
- Multi-path payments: ✅ Correct aggregated fee

### Fast Payments (< 5 sec timeout)
```
User pays invoice
  ↓
LNbits reserves 1% fee
  ↓
[...]

PR #3469, opened 2025-11-01
refactor: lndgrpc, update grpcs file, use types and enums, LookupInvoiceV2

dni:
we can use the track payment v2 top level fees and preimage instead of the htlc's

https://lightning.engineering/api-docs/api/lnd/router/track-payment-v2/#lnrpcpayment

inspiration taken from here: https://github.com/lnbits/lnbits/pull/3468/commits/23438789f6f0d11ca7f191702a8c80f7aa7d0ca5

- added script to update grpcs files
- removed type ignore
- update grpc files
- LookupInvoiceV2

### DIFF
--- a/lnbits/wallets/lndgrpc.py
+++ b/lnbits/wallets/lndgrpc.py
@@ -191,6 +191,7 @@ class LndWallet(Wallet):
             payment_request=bolt11,
             fee_limit_msat=fee_limit_msat,
             timeout_seconds=30,
+            no_inflight_updates=True,
         )
         try:
             res: Payment = await self.router_rpc.SendPaymentV2(req).read()


tests/wallets/fixtures/json/fixtures_rpc.json
```diff
@@ -2160,22 +2191,6 @@
                     }
                   }
                 },
-                {
```
motorina0, line 2163: why remove this test?

dni: i think its an invalid test, it will never return no status

motorina0 (approved):

PR #3473, opened 2025-11-03
refactor: simplify base.js

dni:
- api / windowMixin / utils each its file
- added globals.js and moved bits from base.html


lnbits/static/js/base.js
```diff
@@ -1,182 +1,7 @@
...
-      })
-    },
-    getCurrencies() {
-      return this.request('GET', '/api/v1/currencies').then(response => {
-        return ['sats', ...response.data]
-      })
-    }
-  },
+  utils: window.lnbitsUtils,
+  api: window.lnbitsApi,
```
motorina0:
```suggestion
  api: window._lnbitsApi,
```

Otherwhise we will have `window.LNbits.api` and `window.lnbitsApi` used interchangeably

lnbits/static/js/utils.js
```diff
@@ -0,0 +1,148 @@
+window.lnbitsUtils = {
```
motorina0:
```suggestion
window._lnbitsUtils = {
```

then 
```js
window.LNbits.utils
```

lnbits/static/js/lnurl.js
```diff
@@ -0,0 +1,23 @@
+window.decryptLnurlPayAES = (success_action, preimage) => {
```
motorina0, line 1: why not put this in `utils`?
It should not pollute the global namespace

lnbits/static/js/api.js
```diff
@@ -0,0 +1,173 @@
+window.lnbitsApi = {
```
motorina0:
```suggestion
window.LNbits.api = {
```

why go trough the intermediate variable?

dni: order of initialisation, it will be a reference, so no overhead by just referencing

lnbits/static/js/lnurl.js
dni: that would be a breaking change

PR #3481, opened 2025-11-07
refactor: use decorators for disabled endpoints

dni:
- `check_extension_builder`
- `check_admin_ui`


lnbits/decorators.py
```diff
@@ -359,3 +359,20 @@ def normalize_url(self, *args, **kwargs):
...
+
+
+async def check_admin_ui() -> None:
+    if not settings.lnbits_admin_ui:
+        raise HTTPException(
+            status_code=HTTPStatus.SERVICE_UNAVAILABLE, detail="Admin UI is disabled."
+        )
+
+
+async def check_extension_builder(
```
motorina0, line 371: this is too specific to be a decorator

PR #3431, opened 2025-10-18
fix: improve mobile responsiveness for admin settings

BenGWeeks:
## Summary

Fixes #3400 - Improves mobile responsiveness for admin settings on narrow screens (335px width).

## Changes

- **Mobile layout**: Horizontal tabs at top instead of vertical sidebar to save vertical space
- **Responsive columns**: Changed `col-6` to `col-12 col-md-6` to prevent overflow on mobile
- **Overflow handling**: Added max-width and overflow-wrap to inputs, chips, and long content
- **CSS fixes**: Added mobile-specific CSS for chip wrapping and field hint text spacing
- **Scrolling fix**: Removed swipeable attribute that was interfering with vertical scrolling

## Test plan

- [x] Test on mobile viewport at 335px width
- [x] Verify all tabs are accessible via horizontal tab navigation
- [x] Verify content doesn't overflow or overlap
- [x] Verify chips with long regex patterns wrap properly
- [x] Verify hint text doesn't overlap with content below
- [x] Verify vertical scrolling works correctly within tabs
- [x] Verify desktop layout unchanged and working correctly
- [x] Run make format (all linters passed)

## Screenshots

https://github.com/user-attachments/assets/e1a0edd3-eef1-4700-8542-0e5d5a031ff0

Tested on mobile screens at 335px width - content now properly stacks and wraps without overflow or overlap.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3486, opened 2025-11-08
fix: incorrect styles in `base.html` from #3431

dni:
references here. https://github.com/lnbits/lnbits/pull/3431#discussion_r2480809503


lnbits/static/scss/base.scss
```diff
@@ -132,3 +132,11 @@ video {
...
   padding: 12px;
 }
+
+.wallet-list-card {
+  margin-top: 1px;
+  margin-right: 1rem;
+}
+.wallet-list-card:first-child {
+  margin-left: 1px;
+}
```
talvasconcelos, line 179: Isn't it missing the media queries?

dni: yes, that was the bug and removed on purpose. read here. https://github.com/lnbits/lnbits/pull/3431#discussion_r2480809503

motorina0 (approved):

PR #3501, opened 2025-11-10
fix: vertically center first install

### LINKED ISSUE #3497: [BUG] yertically  center first_install form
Image

talvasconcelos:
main html tag (`<main></main>`) needs "centering"

Style is applied on the template and scoped... so it doesn't affect other tags! 

image

Closes #3497

### DIFF
--- a/lnbits/templates/pages/first-install.vue
+++ b/lnbits/templates/pages/first-install.vue
@@ -1,83 +1,78 @@
 <template id="page-first-install">
-  <div class="row q-col-gutter-md justify-center main first-install">
-    <div class="col-10 col-md-8 col-lg-6 q-gutter-y-md">
-      <q-card>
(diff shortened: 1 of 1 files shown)


talvasconcelos:
IMHO, this is a one time page, it's fine to have the style tag with the styling... I can change it if needed! @dni @motorina0

lnbits/templates/pages/first-install.vue
```diff
@@ -73,3 +73,11 @@
     </div>
   </div>
 </template>
+
+<style scoped>
```
dni: are you sure the scoped, does not work for other pages? not sure if that works because we  dont use the vue templates as we should.

talvasconcelos: Checking... apparently it leaks! 

Let me try another approach

PR #3504, opened 2025-11-11
[feat] user assets

motorina0:
### Summary

Allow users to upload assets (images but also other types).
The assets can be visible only for the user that owns them or made visible for the public.
Removed old `Library` logic.

### User assets
image

### Settings
image


dni (approved):
LGTM, nice

PR #3526, opened 2025-11-13
refactor: `lnbits-wallet-charts` and clean up paymentFiltering

dni:
- debounce rendering of charts. fixes issues if filter is changed too fast and the chart rendering runs into itself destroying and creating thechart at the same time
- pass paymentFilter prop to payment-list and use it
- pass paymentFilter prop to charts and use it
- optional filter argument for utils.js prepareFilter
- fixes unhandled cases ins `handleFilterChanged()` in payment-list


lnbits/core/templates/core/wallet.html
```diff
@@ -550,37 +549,9 @@
...
-            </q-card>
-          </div>
-          <div v-if="hasChartActive && !chartDataPointCount">
-            <q-card>
-              <q-card-section> No chart data available</q-card-section>
-            </q-card>
-          </div>
+          <lnbits-wallet-chart
+            :paymentFilter="paymentFilter"
+          ></lnbits-wallet-chart>
```
motorina0:
```suggestion
          <lnbits-wallet-charts
            :paymentFilter="paymentFilter"
            :show-balance-chart="..."
            :show-...
            :show-...
          ></lnbits-wallet-charts>
```

PR #3527, opened 2025-11-14
feat: add `external_data` field to `User`

dni:
closes: https://github.com/lnbits/lnbits/issues/3530

### DIFF
--- a/lnbits/core/models/users.py
+++ b/lnbits/core/models/users.py
@@ -152,6 +152,7 @@ class UserAcls(BaseModel):
 class Account(BaseModel):
     id: str
     external_id: str | None = None  # for external account linking
+    external_data: str | None = None
     username: str | None = None
     password_hash: str | None = None
     pubkey: str | None = None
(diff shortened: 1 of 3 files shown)


BenGWeeks:
Previously, you could pass in any additional field into the `extra` property:

```
extra: Optional[dict[str, str]] = Field(default=None)
```

However, the extra field is now defined as:

``` 
```

Where `UserExtra` has:

```
    email_verified: bool | None = False
    first_name: str | None = None
    last_name: str | None = None
    display_name: str | None = None
    picture: str | None = None
```

I'm not 100% sure why fields such as the following have been moved to extra, but it means we can no longer store additional user information in there that we used to. For example, we were using this to store items such as:

```
aadObjectId (Azure Active Directory Object ID)
userType (teammate/agent, etc.)
privateWalletId
allowanceWalletId
```
 
If you prefer to keep `extra` strongly defined to store e-mail, first name, last name, etc. perhaps rather than doing (a single long string):

```
external_data: str | None = None
```

Could we please create `external data` as the following (like `extra` used to be):
 
 ```
external_data: Optional[dict[str, str]] = Field(default=None)
```

PR #3537, opened 2025-11-17
[feat] Payment labels

motorina0:
### Labels for payments
 - manage user labels
 - add/remove label to payments
 - search for labels

https://github.com/user-attachments/assets/327f60e8-5615-49a0-84dc-89bb54e6568f


lnbits/core/models/users.py
```diff
@@ -36,6 +37,12 @@ class WalletInviteRequest(BaseModel):
     to_wallet_name: str
 
 
+class UserLabel(BaseModel):
+    name: str | None = None
```
dni: shouldnt thks be `str` instead of `str | None` if it cannot be empty?

lnbits/core/models/users.py
```diff
@@ -55,6 +62,8 @@ class UserExtra(BaseModel):
 
     wallet_invite_requests: list[WalletInviteRequest] = []
 
+    labels: list[UserLabel] = []
```
dni, line 67: is there a reason this in extra? i think i should be on the `User`

lnbits/core/views/auth_api.py
```diff
@@ -422,15 +423,6 @@ async def update(
 ) -> User | None:
     if data.user_id != user.id:
         raise HTTPException(HTTPStatus.BAD_REQUEST, "Invalid user ID.")
-    if data.username and not is_valid_username(data.username):
```
dni, line 425: was this validated on the model already?

lnbits/helpers.py
```diff
@@ -201,6 +201,11 @@ def is_valid_username(username: str) -> bool:
     return re.fullmatch(username_regex, username) is not None
 
 
+def is_valid_label(label: str) -> bool:
```
dni, line 204: this regex should be a condition in the pydantic model.

lnbits/db.py
```diff
@@ -431,6 +431,8 @@ class Operator(Enum):
     INCLUDE = "in"
     EXCLUDE = "ex"
     LIKE = "like"
+    EVERY = "every"
+    ANY = "any"
```
dni, line 435: hmm, i really dont immediatly understand the changes in `db.py` in this big pr, can we somehow isolate this changes and maybe add some tests?

lnbits/core/views/payment_api.py
```diff
@@ -260,6 +263,27 @@ async def api_payments_create(
     return await create_payment_request(wallet_id, invoice_data)
 
 
+@payment_router.put("/{payment_hash}/labels")
```
dni, line 267: dont we already have and enpoint like `update payment details` ? i think it should be reused.

lnbits/core/views/generic.py
```diff
@@ -183,7 +183,7 @@ async def manifest(request: Request, usr: str):
                 "src": (
                     settings.lnbits_custom_logo
                     if settings.lnbits_custom_logo
-                    else "https://cdn.jsdelivr.net/gh/lnbits/lnbits@main/docs/logos/lnbits.png"
+                    else "images/logos/lnbits.png"
```
dni, line 186: unrelated? but i lets that one slide ;)

lnbits/core/migrations.py
```diff
@@ -778,3 +778,11 @@ async def m037_create_assets_table(db: Connection):
         );
         """
     )
+
+
+async def m038_add_labels_for_payments(db: Connection):
```
dni, line 783: `UserLabel` should also be added on db level

lnbits/core/models/users.py
```diff
@@ -202,6 +223,8 @@ def validate_fields(self):
         if user_uuid4.hex != self.id:
             raise ValueError("User ID is not valid UUID4 hex string.")
 
+        self.extra.validate_labels()
```
dni, line 228: here, aswell i think the labels should not be in extra.

lnbits/core/models/payments.py
```diff
@@ -198,6 +207,7 @@ class PaymentFilters(FilterModel):
     preimage: str | None
     payment_hash: str | None
     wallet_id: str | None
+    labels: str | None
```
dni, line 211: this is this `str | None`? this is inconsitent with the other type `list[str] | None`

lnbits/core/models/payments.py
```diff
@@ -320,3 +331,7 @@ class CancelInvoice(BaseModel):
     def check_hex(cls, v):
         _ = bytes.fromhex(v)
         return v
+
+
+class UpdatePaymentLabels(BaseModel):
```
dni, line 337: if we reuse that updatee payment details endpoint we dont need this model and enpoint.

lnbits/core/models/users.py
motorina0: No reason to be in `User`. Extra is for data that we do not index or search for.

lnbits/core/views/payment_api.py
motorina0: There is no update endpoint and I don't think there should be one.

lnbits/core/migrations.py
motorina0: What is the benefit in doing that?

lnbits/core/models/payments.py
motorina0: It is a query param, only string form is accepted. I have tried `list` and the query param is ignored.

lnbits/core/models/payments.py
motorina0: what endpoint?

lnbits/core/views/auth_api.py
motorina0: it is validated at the service level (which is called below)

lnbits/core/models/users.py
```diff
@@ -36,6 +37,12 @@ class WalletInviteRequest(BaseModel):
     to_wallet_name: str
 
 
+class UserLabel(BaseModel):
+    name: str = Field(regex=r"([A-Za-z0-9 ._-]{1,100}$)")
+    description: str | None = Field(default=None, max_length=250)
+    color: str | None = Field(default=None, max_length=7)  # e.g., "#RRGGBB"
```
dni: trhere is a rtegex for this color also :) `^#(?:[0-9a-fA-F]{3}){1,2}$`

motorina0: done

PR #3559, opened 2025-11-24
feat: move `wallet.html` to vue component. FINAL DynamicComponent pr

dni:
this is the last of the extensive frontend refactoring getting rid of `Dynamic Component`

difference in generic /wallet api.
- lastActiveWallet is now saved into localstorage
- there is no wallet creation if a user has no wallets now
- wallets check are in the frontend now done from `g.user.wallets`


lnbits/core/views/generic.py
```diff
@@ -42,54 +35,6 @@ async def favicon():
...
-)
-async def get_user_wallet(
-    request: Request,
-    lnbits_last_active_wallet: Annotated[str | None, Cookie()] = None,
-    user: User = Depends(check_user_exists),
-    wal: UUID4 | None = Query(None),
-):
-    if wal:
-        wallet = await get_wallet(wal.hex)
-    elif len(user.wallets) == 0:
```
motorina0, line 58: This check is still important. An user should ALWAYS have at least one wallet.

dni: how about we always create 1 wallet when creating an user and an user can only delete his last wallet when he deletes his account?

dni: a wallet is already created within the `create_user_account` service and with this would that be good? https://github.com/lnbits/lnbits/pull/3564

PR #3562, opened 2025-11-24
feat: check_new_accounts_allowed decorator for register api

dni:
- `create_user_account_no_ckeck` was unecessary. the check is  only necessary in the the `/register` and post `/api/v1/account` endpoint. created a small decorator for that to check.

### DIFF
--- a/lnbits/decorators.py
+++ b/lnbits/decorators.py
@@ -558,5 +558,5 @@ async def check_new_accounts_allowed():
     if not settings.new_accounts_allowed:
         raise HTTPException(
             status_code=HTTPStatus.BAD_REQUEST,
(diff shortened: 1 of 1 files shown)


motorina0:
what is the `feat` here?

dni:
> what is the `feat` here?

i guess the new decorator? feel free to change the description if you have a better idea.

motorina0:
Out of scope.

dni:
can you fix the typo at least? `create_user_account_no_ckeck`.  
i don't think it is out of scope, the service should not check if we have public user creation disabled (thats only for registering from the frontend / public api), because the service should always work so this `create_user_account_no_ckeck` should actually be the default behaviour.

PR #3565, opened 2025-11-25
[perf] pending payments check

motorina0:
### Summary

- Do not re-check paiment payments by default on search (too slow). But allow a recheck if explicitly required.

- On the wallet page recheck in the `background` the pending payments. This way the wallet payments load faster and the user is going to get an update once the backgorund operation is done.

- fix: double query for payments (once in `created` and once in `watch wallet`). Removed the one created.

- fix: `INCLUDE` and `ECVLUDE` operators.


dni (approved):
tested, lgtm

PR #3567, opened 2025-11-25
feat: move qrcode scanner into reuseable component

dni:
@talvasconcelos can you test that an take it over?


dni (commented):
looks good tal

talvasconcelos (approved):

PR #3580, opened 2025-11-26
chore: cleanup `LNBITS_DENOMINATION` global var

dni:
use already existing `g.denomination` which is now also reactive


lnbits/static/js/components/lnbits-payment-list.js
```diff
@@ -18,7 +17,7 @@ window.app.component('lnbits-payment-list', {
           {
             name: 'amount',
             align: 'right',
-            label: this.$t('amount') + ' (' + LNBITS_DENOMINATION + ')',
```
dni, line 21: we dont need the denom in the labels of the payment csv

lnbits/static/js/pages/node.js
```diff
@@ -91,14 +91,14 @@ window.PageNode = {
           {
             name: 'sat',
             align: 'right',
-            label: this.$t('amount') + ' (' + LNBITS_DENOMINATION + ')',
+            label: this.$t('amount'),
```
dni, line 94: we dont need denomination on node page!

dni: custom denom is only for fake wallet, so no node manager :)

arcbtc (approved):

PR #3583, opened 2025-11-26
feat: change utils.formatDate and utils.formatTimestamp

dni:
- added dateFrom and timestampFrom
- use utils everywhere


lnbits/static/js/components.js
```diff
@@ -749,24 +749,6 @@ window.app.component('lnbits-channel-balance', {
   `
 })
 
-window.app.component('lnbits-date', {
```
dni, line 752: this was a too small component to be useful and it was only used on node manager page

PR #3584, opened 2025-11-26
refactor: move copyText and formatBalance into utils

dni:
add utils to mixin and a preferred way for the future should be `utils.copyText` so you can easily differentiate between a core function and a component scoped function

TODO: follow up PR renaming all the `copyText` inside core, not doing that in this pr for clarity.


talvasconcelos (approved):

motorina0:
Any extensions using these functions?

lnbits/static/js/init-app.js
```diff
@@ -107,6 +109,11 @@ window.app.mixin({
       const isVueRoute = matchedRoute?.matched?.length > 0
       return isVueRoute
     }
+  },
+  // backwards compatibility for extensions, should not be used in the future
+  methods: {
+    copyText: window._lnbitsUtils.copyText,
```
dni, line 115: > Any extensions using these functions?

yes, but its 100% backwards compatible with this

PR #3570, opened 2025-11-25
fix: cleanup paymentEvents, simplify websockets

dni:
i dont like it that its still in the windowMixin `created()` but the code is much cleaner now and works pretty well as of my tests. code from created will eventually move to the place where `g.user` is mapped

PR #3587, opened 2025-11-26
fix: regression receive dialog was not closing

dni:
accidently remived here, while testing. https://github.com/lnbits/lnbits/pull/3570


lnbits/static/js/globals.js
```diff
@@ -23,6 +23,7 @@ window.g = Vue.reactive({
   showNewWalletDialog: false,
   newWalletType: 'lightning',
   updatePayments: false,
+  updatePaymentsHash: false,
```
motorina0: this is a good place to leave a comment with what this global variables are for

dni: done

motorina0 (approved):

PR #3596, opened 2025-11-27
feat: use latest fonts, roboto and material icons

dni:
as recommended here: https://quasar.dev/start/umd#installation


lnbits/static/scss/base.scss
```diff
@@ -68,22 +68,6 @@ video {
   border-radius: 3px;
 }
 
-.material-icons {
-  font-family: 'Material Icons';
-  font-weight: normal;
-  font-style: normal;
-  font-size: 24px;
```
dni, line 75: this was overwritting default styline of icons, which makes for example the <lnbits-manage-wallet> icons look a bit smaller.

PR #3599, opened 2025-11-27
perf: optimisation to avoid double db calls

motorina0:
### Summary
The get paginated payments for wallet API is one of the most called ones.

This PR reduces one call to get the wallet.


lnbits/core/crud/payments.py
```diff
@@ -133,8 +135,9 @@ async def get_payments_paginated(  # noqa: C901
     if since is not None:
         clause.append(f"time > {db.timestamp_placeholder('time')}")
 
-    if wallet_id:
-        wallet = await get_wallet(wallet_id)
+    if wallet_id or wallet:
```
dni, line 138: this if can be simplified

dni: pushed simplified version @motorina0

PR #3600, opened 2025-11-27
perf: use `check_account_exists` decorator

motorina0:
### Summary
The `check_user_exists` decorator fetches data that is rarely needed in the API function.
This new decorator does the minimum required work, and significantly improves performance.

  More performant version of `check_user_exists`.
  Unlike `check_user_exists`, this function:
    - does not check extension access
    - does not fetch the user wallets

`check_account_id_exists` decorator caches the result for better performance

### DIFF
--- a/lnbits/decorators.py
+++ b/lnbits/decorators.py
@@ -155,7 +155,7 @@ async def check_account_id_exists(
     if access_token:
         cache_key = f"auth:access_token:{sha256s(access_token)}"
     elif usr:
-        cache_key = f"auth:user_id:{usr.hex}"
+        cache_key = f"auth:user_id:{sha256s(usr.hex)}"
 
     if cache_key and settings.auth_authentication_cache_minutes > 0:
         account_id = cache.get(cache_key)


lnbits/core/views/extensions_builder_api.py
```diff
@@ -128,10 +129,10 @@ async def api_deploy_extension(
 )
 async def api_preview_extension(
     data: ExtensionData,
-    user: User = Depends(check_user_exists),
+    account: AccountId = Depends(check_account_id_exists),
```
dni:
```suggestion
    account_id: AccountId = Depends(check_account_id_exists),
```

lnbits/core/views/payment_api.py
```diff
@@ -118,14 +118,14 @@ async def api_payments_history(
 async def api_payments_counting_stats(
     count_by: PaymentCountField = Query("tag"),
     filters: Filters[PaymentFilters] = Depends(parse_filters(PaymentFilters)),
-    user: User = Depends(check_user_exists),
+    account: AccountId = Depends(check_account_id_exists),
```
dni: lets call it `account_id` everywhere to not confuse it

PR #3601, opened 2025-11-28
fix: g.wallet artifacts in lnbits-payment-list and lnbits-manage-wall...

dni:
…et-list

- leftovers ins payment-list
- use `g.lastActiveWallet` inside lnbits-drawer instead


lnbits/templates/components/lnbits-manage-wallet-list.vue
```diff
@@ -11,24 +11,16 @@
       )"
       :key="walletRec.id"
       clickable
-      :active="g.wallet && g.wallet.id === walletRec.id"
+      :active="g.lastActiveWallet === walletRec.id"
```
dni: hm, it stays active when not on wallet page now aswell... i dont like it

dni: changed it from `g.lastActiveWallet` to a reactive var inside lnbits-manage-wallet and that component watches to `$route` and sets the activeWalletId from the $route `params.id`

PR #3603, opened 2025-11-28
[perf] Faster require invoice key

motorina0:
### Summary
The `require_admin_key` and `require_invoice_key` are too heavy. They also return the wallet with all its data. The wallet balance is rarely used and it takes time to computed.

This PR introduces `require_base_admin_key` and `require_base_invoice_key` decorators which fetch a light version of the wallet (`BaseWallet`) and caches the result.

If the wallet is deleted or the api-keys are reset, then the cache is cleared.

The old `BaseWallet` has been renamed to `WalletBasicInfo`.

### DIFF
--- a/lnbits/core/models/wallets.py
+++ b/lnbits/core/models/wallets.py
@@ -11,7 +11,7 @@ from lnbits.db import FilterModel
 from lnbits.settings import settings
 
 
-class WalletBasicInfo(BaseModel):
+class WalletInfo(BaseModel):
     id: str
     name: str
     adminkey: str
--- a/lnbits/core/views/api.py
+++ b/lnbits/core/views/api.py
@@ -12,7 +12,6 @@ from lnbits.core.models import (
     CreateWallet,
     User,
     Wallet,
-    WalletBasicInfo,
 )
 from lnbits.core.models.users import AccountId
 from lnbits.decorators import (
@@ -71,7 +70,6 @@ async def health_check(
     "/api/v1/wallets",
     name="Wallets",
     description="Get basic info for all of user's wallets.",
-    response_model=list[WalletBasicInfo],
 )
 async def api_wallets(user: User = Depends(check_user_exists)) -> list[Wallet]:
     return user.wallets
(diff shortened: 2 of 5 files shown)


lnbits/core/views/api.py
```diff
@@ -71,7 +71,7 @@ async def health_check(
     "/api/v1/wallets",
     name="Wallets",
     description="Get basic info for all of user's wallets.",
-    response_model=list[BaseWallet],
+    response_model=list[WalletBasicInfo],
```
dni: redundant, is return type = repsonse model

lnbits/core/models/wallets.py
```diff
@@ -230,6 +233,12 @@ class WalletTypeInfo:
     wallet: Wallet
 
 
+@dataclass
+class BaseWalletTypeInfo:
```
dni, line 237:
```suggestion
class BaseWalletKeyInfo:
```

lnbits/decorators.py
```diff
@@ -102,16 +128,50 @@ async def __call__(self, request: Request) -> WalletTypeInfo:
...
 
         request.scope["user_id"] = wallet.user
-        if self.expected_key_type is KeyType.admin and wallet.adminkey != key_value:
+        await _check_user_access(request, wallet.user)
+
+        key_type = await self._extract_key_type(key_value, wallet)
+        return WalletTypeInfo(key_type, wallet)
+
+
+class LightKeyChecker(BaseKeyChecker):
```
dni, line 137: i think you should just use 1 keychecker and add another argument and switch the crud getter on it. also cache all keychecks

am i missing something?

lnbits/core/models/wallets.py
```diff
@@ -11,7 +11,7 @@
 from lnbits.settings import settings
 
 
-class BaseWallet(BaseModel):
+class WalletBasicInfo(BaseModel):
```
dni:
```suggestion
class WalletInfo(BaseModel):
```

lnbits/core/models/wallets.py
dni: or `BaseWalletKeyType`

dni: Type is to generic

lnbits/core/models/wallets.py
motorina0: ok

PR #3604, opened 2025-11-28
feat: dynamic login and registering

dni:
- combines public and non-public pages into base.html
- tested with ext public page on extension (lnurlp) logged + public and  not loggedin + public
- tested public error pages


lnbits/templates/base.html
```diff
@@ -72,6 +77,9 @@
     {% if user %}
     <script>
       window.g.user = LNbits.map.user(JSON.parse({{ user | tojson | safe }}));
+      {% if not public %}
```
dni, line 80: `isPagePublic` default to `true` so only render `false` if public is None or False aka, `not public`

lnbits/static/js/pages/home.js
```diff
@@ -136,7 +146,7 @@ window.PageHome = {
   },
   created() {
     if (this.g.isUserAuthorized) {
-      window.location.href = '/wallet'
+      return this.auth()
```
motorina0: why auth if authorized?

lnbits/static/js/pages/home.js
```diff
@@ -98,32 +98,42 @@ window.PageHome = {
...
         await LNbits.api.loginUsr(this.usr)
-        this.usr = ''
-        window.location.href = '/wallet'
+        this.auth()
+      } catch (e) {
+        console.warn(e)
+        LNbits.utils.notifyApiError(e)
+      }
+    },
+    async auth() {
```
motorina0: this should be called `refreshUser()` or something like that

lnbits/static/js/pages/home.js
dni: getting the full `user` from the auth endpoint, `g.isUserAuth...` is just taken from the cookie

PR #3605, opened 2025-11-28
feat: dynamic extension loading via `routes.json`

dni:
- extensions can now provide a `/extname/static/routes.json` with defined VueRouter routes.
- if no vue route is found frontend will try to fetch this file, if this file does not exist it wil reload the page just as it was for extensions before.

example of a dynamic extension refactor: https://github.com/lnbits/bitcoinswitch_extension/pull/49


lnbits/templates/base.html
```diff
@@ -84,6 +84,14 @@
     {% endif %}
     <!-- scripts from extensions -->
     {% block scripts %}{% endblock %}
+    <!-- dynamic extensions routes-->
+    <script>
+      {% if dynamic_routes %}
+        window.dynamicRoutes = {{ dynamic_routes | tojson }}
```
dni: rendering all known routes upfront would solve the issue that public pages for extensions doent load because the route does not exist yet. now the `<lnbits-manage-extension-list>` is responible for loading, and that does not exist on a public page for non-logged in users.

with this you also dont need to redirect when adding dynamic routes from `<lnbits-manage-extensions-list>` after a reload

lnbits/templates/base.html
```diff
@@ -60,6 +63,7 @@
       </q-layout>
     </div>
     <script type="text/javascript">
+      const RENDERED_ROUTE = '{{ request.path }}'
```
motorina0, line 70: Why does this need to be jinja?
Is this different than `window.location.pathname`?

lnbits/static/js/components/lnbits-manage-extension-list.js
```diff
@@ -47,7 +40,7 @@ window.app.component('lnbits-manage-extension-list', {
         })
     }
   },
-  async created() {
-    await this.loadExtensions()
+  created() {
+    this.loadExtensions()
```
motorina0: why not `await`?

lnbits/templates/base.html
dni: yes, we need to somehow now which, extension was rendered is is in `<div>{page template}</div>`

lnbits/static/js/components/lnbits-manage-extension-list.js
dni: i think it is blocking the rendering, and its cleaner to have it without async

dni: i reverted it, its blocking we are loosing 100 ms with this on page loaded event.

lnbits/templates/base.html
dni: https://github.com/lnbits/lnbits/pull/3605/files#diff-f8dbf787f464944da8d9282c32012baaaaf30a30b41be2a63df08a37ed1adbd5R43

PR #3606, opened 2025-11-29
feat: customizable Apple touch icon

btcforplebs:
Add the ability for a user to configure the icon used for the iOS homescreen from the admin site customization field

### DIFF
--- a/lnbits/templates/base.html
+++ b/lnbits/templates/base.html
@@ -25,6 +25,10 @@
     />
     <meta name="mobile-web-app-capable" content="yes" />
     <meta name="apple-mobile-web-app-capable" content="yes" />
+    <link
+      rel="apple-touch-icon"
+      href="{{ LNBITS_APPLE_TOUCH_ICON if LNBITS_APPLE_TOUCH_ICON else static_url_for('static', 'images/maskable_icon.png') }}"
+    />
     {% if web_manifest %}
     <link async="async" rel="manifest" href="{{ web_manifest }}" />
     {% endif %} {% block head_scripts %}{% endblock %}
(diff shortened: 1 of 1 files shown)


btcforplebs:
Your welcome!

motorina0:
The image size is 1024x1024, which is too large.
https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html

Can we use the existing `lnbits/static/images/logos/lnbits.png` instead?

btcforplebs:
Thanks for commenting. I can confirm, `lnbits/static/images/maskable_icon.png` can be used with no issues as the default option. I updated the PR to reflect this change

lnbits/templates/base.html
btcforplebs, line 1: Sorry, got the spacing wrong on this and It updated the whole file

motorina0: > Sorry, got the spacing wrong on this and It updated the whole file

it was very hard to rebase. I have fetched the original `base.html` file from `dev`, can you please update it again.

btcforplebs: Finished

btcforplebs:
Cleaned up, thanks!

lnbits/templates/base.html
```diff
@@ -25,6 +25,10 @@
     />
     <meta name="mobile-web-app-capable" content="yes" />
     <meta name="apple-mobile-web-app-capable" content="yes" />
+    <link
+      rel="apple-touch-icon"
+      href="{{ LNBITS_APPLE_TOUCH_ICON if LNBITS_APPLE_TOUCH_ICON else static_url_for('static', 'images/maskable_icon.png') }}"
```
motorina0, line 30: cc: @dni `static_url`

btcforplebs:
Thank you @dni and @motorina0 for the assistance 👊

PR #3607, opened 2025-11-30
feat: change default table pagination

prusnak:
from

[5, 7, 10, 15, 20, 25, 50, All]

to

[5, 10, 20, 50, 100, 200, 500, All]

Original seems to have too many options in lower range which are not very useful.


prusnak:
Please test I was not able to run LNbits on my current setup :(

dni (changes requested):
hm, it has no effect on <lnbits-payment-list /> seems like it does not see the defaults... even though its not set on payment-list

dni:
maybe we are running into this issue... https://github.com/quasarframework/quasar/discussions/15817
because even when i set i manually, it does not work.
 for quasar v2 you need to use `config:table`

prusnak:
> because even when i set i manually, it does not work.

you did not regenerate the bundle, so maybe that's the reason?

dni:
pushed the bundle and tried it again, but no, this seems like an actual quasar issue :o i am a bit stumped

dni (approved):

PR #3615, opened 2025-12-04
chore: cleanup base.html, minor issue and `g.user` initialisation

dni:
- minor issues using global `user` instead of `g.user`
- add comments to rendering in base.html
- move user rendering from macro into base.html

required in. https://github.com/lnbits/lnbits/pull/3604


lnbits/static/js/components.js
```diff
@@ -360,7 +360,7 @@ window.app.component('lnbits-update-balance', {
   props: ['wallet_id', 'small_btn'],
   computed: {
     admin() {
-      return user.super_user
+      return this.g.user && this.g.user.super_user
```
motorina0:
```suggestion
      return this.g.user?.super_user === true
```

motorina0 (approved):

PR #3617, opened 2025-12-04
fix: extension stayed active in frontend extension list

dni:
watch $route instead of mapping an `isActive` property


lnbits/templates/components/lnbits-manage-extension-list.vue
```diff
@@ -19,7 +19,7 @@
       v-for="extension in userExtensions"
       :key="extension.code"
       clickable
-      :active="extension.isActive"
+      :active="$route.path == extension.url"
```
motorina0: will this work for `my_extension/some_stuff`?
isn't `startsWith` better?

dni: i think startswith is better

PR #3622, opened 2025-12-05
feat: add vue router navigation to `/admin`

dni:
:)


motorina0:
In this case `/foo#tab` is the correct way because it indicates that it is a location in the page.
The `/foo/tab` form indicates a new page.

NACK

dni:
> In this case `/foo#tab` is the correct way because it indicates that it is a location in the page. The `/foo/tab` form indicates a new page.
> 
> NACK

fixed

motorina0 (approved):

PR #3624, opened 2025-12-05
[perf] reuse connection

motorina0:
### Summary
Reconnecting to the DB is a time consuming task. This PR tries to re-use the connection for the most used API calls.

Add indexes for the most common search columns.


lnbits/core/views/payment_api.py
```diff
@@ -186,18 +184,24 @@ async def api_payments_paginated(
...
+            filters=filters,
+            conn=conn,
+        )
+        if not recheck_pending:
+            return page
 
-    for payment in page.data:
-        if payment.pending:
-            await update_pending_payment(payment)
+        payments = []
```
motorina0, line 196: this is in fact a bug fix

PR #3627, opened 2025-12-05
fix: websocket only listen to filtered wallets

dni:
- only open websocket connection to the filterWallets (max 10) currently


lnbits/static/js/components/lnbits-manage-wallet-list.js
```diff
@@ -14,5 +19,69 @@ window.app.component('lnbits-manage-wallet-list', {
...
+      if (data.payment.amount > 0) {
+        eventReaction(data.wallet_balance * 1000)
+      }
+    },
+    paymentEvents() {
+      if (!this.g.user) return
+      let timeout
+      let count = 0
+      this.g.user.wallets.forEach(wallet => {
+        if (count >= this.maxWallets) return
```
motorina0: `return` does not stop the `forEach`
better do a `slice()`, no point in iterating over 1000+ wallets

dni: good point

dni: pushed

motorina0 (approved):

PR #3635, opened 2025-12-06
feat: improve on create wallet frontend and api. (BREAKING CHANGE)

dni:
- use `newWallet` reactive variable for opening the dialog, (just like qrscanner now)
-  BREAKING CHANGE:  `post` route to use check_account_id_exists instead of adminkey of the first wallet.
- dynamically add wallets from frontend
- api exit first if no valid `WalletType`


lnbits/static/js/components/lnbits-wallet-new.js
```diff
@@ -46,36 +68,36 @@ window.app.component('lnbits-wallet-new', {
         LNbits.utils.notifyApiError(err)
       }
     },
-    async submitAddWallet() {
-      const data = this.newWallet
-      if (this.g.newWalletType === 'lightning' && !data.name) {
+    submitAddWallet() {
+      const data = this.wallet
+      if (this.g.newWallet === 'lightning' && !data.name) {
```
motorina0: `newWalletType` is the more suggestive variable name.
Why `newWallet`?

dni: reverted.

PR #3642, opened 2025-12-08
feat: sort payments in wallet

motorina0:
Sort payment in the wallet (payment list)

image


lnbits/core/migrations.py
```diff
@@ -794,9 +794,11 @@ async def m039_index_payments(db: Connection):
         "checking_id",
         "payment_hash",
         "amount",
+        "fee",
```
dni, line 797: you think we should just add it like that because there is no release yet? it will break my current db i guess right?

motorina0: > you think we should just add it like that because there is no release yet? 
- yes
- it will break my current db i guess right?
- no, that is just an index

PR #3644, opened 2025-12-08
refactor: `g.settings` with a `PublicSettings` class for mapping, for reactive `WINDOW_SETTINGS`

dni:
this is required if we want to make the settings from `/admin` reactive without reloading.

from now on use `g.settings` or `g` in general for reactive vars, no more jinja


lnbits/templates/base.html
```diff
@@ -68,10 +68,7 @@
     </div>
     <script type="text/javascript">
       const RENDERED_ROUTE = '{{ request.path }}'
-      const WINDOW_SETTINGS = {{ WINDOW_SETTINGS | tojson }}
-      Object.keys(WINDOW_SETTINGS).forEach(key => {
```
dni, line 74: needs to be tested if thats required for extensions, else now the core frontend does not need this anymore.

motorina0:
I find this a bit risky...

motorina0:
Concept OK.

lnbits/static/js/globals.js
```diff
@@ -9,53 +9,48 @@ const localStore = (key, defaultValue) => {
 }
 
 window.g = Vue.reactive({
+  // vars from server
+  settings: SETTINGS,
+  currencies: CURRENCIES,
+  extensions: SETTINGS.extensions,
+  allowedCurrencies: SETTINGS.allowedCurrencies,
+  denomination: SETTINGS.denomination,
```
motorina0, line 17: why duplicate these fields and not use settings.xxx?

lnbits/helpers.py
```diff
@@ -134,6 +88,9 @@ def template_renderer(additional_folders: list | None = None) -> Jinja2Templates
             t.env.globals["INCLUDED_CSS"] = vendor_files["css"]
             t.env.globals["INCLUDED_COMPONENTS"] = vendor_files["components"]
 
+    # backwards compatibility for extensions (tpos)
+    t.env.globals["LNBITS_DENOMINATION"] = settings.lnbits_denomination
```
motorina0, line 92: why not put it up with the rest?

lnbits/settings.py
```diff
@@ -1126,6 +1126,104 @@ def is_ready_to_install_extension_id(self, ext_id: str) -> bool:
             and ext_id in self.lnbits_all_extensions_ids
         )
 
+    def to_public(self) -> PublicSettings:
+        return PublicSettings.from_settings(self)
+
+
+class PublicSettings(BaseModel):
```
motorina0, line 1138: add a comment that all fields here will be visible to the public and no fields info should be included

lnbits/static/js/globals.js
dni: backwards compat for extensions. i used those in 1 or 2 in the future its better to just use settings.

lnbits/helpers.py
dni: just for visibility, this will be removed those after 1.5.0 updates for those 3-5 ext that use it.

lnbits/settings.py
dni: done

PR #3645, opened 2025-12-08
fix: reload loop in `/upgrades/` routes + error status_code

dni:
2 in 1

### DIFF
--- a/lnbits/templates/base.html
+++ b/lnbits/templates/base.html
@@ -52,9 +52,7 @@
               v-if="g.user && !g.isPublicPage"
             ></lnbits-header-wallets>
             <!-- block page content from static extensions -->
-            <div
-              v-if="$route.path === '{{ request.path }}'.replace(/^\/upgrades\/[A-Za-z0-9_-]+/, '')"
-            >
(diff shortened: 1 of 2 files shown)


lnbits/templates/base.html
```diff
@@ -67,7 +69,7 @@
       </q-layout>
     </div>
     <script type="text/javascript">
-      const RENDERED_ROUTE = '{{ request.path }}'
+      const RENDERED_ROUTE = '{{ request.path }}'.replace(/^\/upgrades\/[A-Za-z0-9_-]+/, '')
```
motorina0:
```suggestion
      const RENDERED_ROUTE = '{{ normalize_path(request.path) }}'
```

we already have a function for this

motorina0: add `normalize_path` in `helpers.py`:
```py
    t.env.globals["static_url_for"] = static_url_for
    t.env.globals["normalize_path"] = normalize_path
```

dni: done

motorina0 (approved):

PR #3654, opened 2025-12-10
fix: restore the blur on drawer

talvasconcelos:
Restore the blurring on transparent drawers and toolbar

Closes lnbits/tpos#185


dni:
i removed the blur because it causes this annoying flickering bug. have you seen that one now?

dni (approved):

PR #3658, opened 2025-12-11
fix: account dropdown

talvasconcelos:
- if no image is set for avatar fallback to icon, and drop it inside menu
- remove user role inside menu
- hide the account dropdown in public pages

### DIFF
--- a/lnbits/templates/components/lnbits-header.vue
+++ b/lnbits/templates/components/lnbits-header.vue
@@ -84,12 +84,15 @@
           <q-avatar v-else icon="account_circle" size="18px"></q-avatar>
         </template>
         <q-list style="max-width: 200px">
-          <q-item
-            v-if="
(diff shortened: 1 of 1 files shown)


lnbits/templates/components/lnbits-header.vue
```diff
@@ -73,32 +73,36 @@
...
-        flat
-        rounded
-        size="sm"
-        class="q-pl-sm"
-      >
+      <q-btn-dropdown v-if="g.user" flat rounded size="sm" class="q-pl-sm">
         <template v-slot:label>
-          <q-avatar v-if="g.user?.extra?.picture !== ''" size="18px">
+          <q-avatar
+            v-if="g.user?.extra?.picture && g.user?.extra?.picture !== ''"
```
dni:
```suggestion
            v-if="g.user?.extra?.picture? !== ''"
```

would this work?

dni (approved):
tested

Issue #2581, opened 2024-07-12
Support sending to BOLT12 offer


prusnak:
Title says it all

Prerequisite of #2582

bitkarrot:
Hooray

chrisguida:
PLEASE DO THIS

21M4TW:
I started to add support for Bolt12 (https://github.com/lnbits/lnbits/pull/3092), but currently the code I have is to receive to offers, and there is no UI yet. I will keep working on it...

21M4TW:
Ok so supposedly using LDK for this purpose would use a wrapped library and would not require a service, but it is not usable yet. An issue (https://github.com/lightningdevkit/ldk-node/issues/504) was opened recently and is being worked on...

21M4TW:
Ok so I just looked in plugins/offers_offer.c, and it seems the lack of option_onion_messages bit in my peers' feature bits is what triggers the error message about my peers not being public. All my peers have the following feature bits:
800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000888252a1

Not sure why and how to interpret this...

ZenenTreadwell:
I am working through the basics of how this would work from the user's perspective - I am assuming use of Core Lightning, just for the sake of writing what I know.

On the mobile view of LNBits, there are three options for interaction - 'send', 'scan', and 'receive'. BOLT12 is mostly an upgrade for offers, so I am more concerned with sending for now. I want to send value to someone, so I either scan their bolt12 code from a QR or paste it directly (or resolve it via human-readable name like in BIP353). 

In all cases, the client receives a 'lno' string. This string prefix acts as positive confirmation that we are dealing with a bolt12 offer and prompts a UI box to appear for the user to ask for the amount_msat value (optionally a memo or other fields, but let's keep it simple for now). When the UI box value is sent, We can send a signal to the lightning node via rpc to run 'fetchinvoice' on the `lno` code with an amount param, which (if successful) returns a `lni` invoice code.

[...]

PR #3663, opened 2025-12-15
feat: Bolt12 Support (Offers Creation and Decoding, Invoice Fetching and Payment support at API level with CLN RPC)

21M4TW:
-Adding Bolt12 support for the CLN RPC (offer creation, decoding, invoice fetching, payment, offer enabling and disabling)
-Using a new apioffers table to keep track of offers associated to LNBits wallets
-Offers decoding through the CLN RPC.

This PR is a rebased and cleaned up version of https://github.com/lnbits/lnbits/pull/3092 .


21M4TW:
Did some updates after running black, ruff and mypy. Would it be possible to approve the workflow again? Thanks!

lnbits/core/crud/offers.py
```diff
@@ -0,0 +1,254 @@
...
+
+async def get_offers(
+    *,
+    wallet_id: str | None = None,
+    active: bool | None = None,
+    single_use: bool | None = None,
+    used: bool | None = None,
+    since: int | None = None,
+    filters: Filters[OfferFilters] | None = None,
+    conn: Connection | None = None,
```
dni: can we put `Connection` last? justfor consistency

lnbits/core/crud/offers.py
```diff
@@ -0,0 +1,254 @@
...
+
+async def delete_expired_offers(
+    conn: Connection | None = None,
+) -> None:
+    # We delete all offers whose expiry date is in the past
+    await (conn or db).execute(
+        f"""
+        DELETE FROM apioffers
+        WHERE expiry < {db.timestamp_placeholder("now")}
+        """,  # noqa: S608
```
dni, line 156: why noqa? it shouldnt need it

lnbits/core/crud/offers.py
```diff
@@ -0,0 +1,254 @@
...
+        f"""
+        UPDATE apioffers SET active = :active,
+        updated_at = {db.timestamp_placeholder("now")}
+        WHERE offer_id = :offer_id
+        """,  # noqa: S608
+        {"now": time(), "offer_id": offer_id, "active": False},
+    )
+
+
+async def mark_webhook_offer_sent(offer_id: str, status: int) -> None:
```
dni, line 249: pass `Connection` here aswell on the last 2 fns.

lnbits/core/crud/payments.py
```diff
@@ -268,13 +269,15 @@ async def delete_expired_invoices(
 async def create_payment(
     checking_id: str,
     data: CreatePayment,
+    created_at: int | None = None,
+    updated_at: int | None = None,
```
dni, line 273: those args should not be passed, those should be created automatically when the model is initialized.

lnbits/core/crud/payments.py
```diff
@@ -285,12 +288,16 @@ async def create_payment(
...
+        offer_id=data.offer_id,
         memo=data.memo,
+        payer_note=data.payer_note,
         preimage=data.preimage,
         expiry=data.expiry,
         webhook=data.webhook,
         fee=-abs(data.fee),
         tag=extra.get("tag", None),
+        created_at=datetime.now(timezone.utc) if not created_at else created_at,
+        updated_at=datetime.now(timezone.utc) if not updated_at else updated_at,
```
dni: they should be initialized on the model.

lnbits/core/models/offers.py
```diff
@@ -0,0 +1,113 @@
...
+    webhook_status: str | None = None
+    tag: str | None = None
+    extension: str | None = None
+    created_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
+    updated_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
+    extra: dict = {}
+
+    @property
+    def is_active(self) -> bool:
+        return not self.active == False  # noqa E712
```
dni:
```suggestion
        return not self.active is False
```

remove noqa's

lnbits/core/models/offers.py
```diff
@@ -0,0 +1,113 @@
...
+        return int(self.amount / 1000) if self.amount else None
+
+    @property
+    def is_expired(self) -> bool:
+        return self.expiry < datetime.now(timezone.utc) if self.expiry else False
+
+    async def check_status(self) -> OfferStatus:
+        funding_source = get_funding_source()
+        status = await funding_source.get_offer_status(self.offer_id)
+        return status
```
dni: can we remove this `check_status`? i know its takes from `payment.get_status` but i really want to get rid of it on payment also.

can we do in `services/payments.py` maybe add a `check_offer_status()` there and pass the offer_id.

lnbits/core/models/offers.py
```diff
@@ -0,0 +1,113 @@
...
+    extra: dict | None = {}
+    expiry: datetime | None = None
+    webhook: str | None = None
+
+
+class Offer(BaseModel):
+    offer_id: str
+    wallet_id: str
+    amount: int | None = None
+    active: bool
```
dni, line 22:
```suggestion
    active: bool = Field(alias="is_active")
```
you could do this instead of those properties checking for a boolean.

lnbits/core/models/offers.py
```diff
@@ -0,0 +1,113 @@
...
+    bolt12: str
+    memo: str | None = None
+    expiry: datetime | None = None
+    webhook: str | None = None
+    webhook_status: str | None = None
+    tag: str | None = None
+    extension: str | None = None
+    created_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
+    updated_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
+    extra: dict = {}
```
dni: what is the extra here?

lnbits/core/services/payments.py
```diff
@@ -34,25 +36,238 @@
...
+from ..crud import disable_offer as crud_disable_offer
+from ..crud import enable_offer as crud_enable_offer
+from ..models import CreateOffer, CreatePayment, Offer, Payment, PaymentState, Wallet
 from .notifications import send_payment_notification_in_background
 
 payment_lock = asyncio.Lock()
 wallets_payments_lock: dict[str, asyncio.Lock] = {}
 
 
+async def create_offer(
```
dni: lets move those to `services/offers.py`

lnbits/core/views/offer_api.py
```diff
@@ -0,0 +1,190 @@
...
+    name="Offer List",
+    summary="get list of offers",
+    response_description="list of offers",
+    response_model=list[Offer],
+    openapi_extra=generate_filter_params_openapi(OfferFilters),
+)
+async def api_offers(
+    key_info: WalletTypeInfo = Depends(require_invoice_key),
+    filters: Filters = Depends(parse_filters(OfferFilters)),
+):
```
dni:
```suggestion
) -> list[Offer]:
```

lnbits/core/views/offer_api.py
```diff
@@ -0,0 +1,190 @@
...
+
+offer_router = APIRouter(prefix="/api/v1/offers", tags=["Offers"])
+
+
+@offer_router.get(
+    "",
+    name="Offer List",
+    summary="get list of offers",
+    response_description="list of offers",
+    response_model=list[Offer],
```
dni:
```suggestion
```

lnbits/core/views/offer_api.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+    return offer
+
+
+@offer_router.post(
+    "/{offer_id}/enable",
+    summary="Enable an offer",
+    description="""
+        This endpoint can be used to enable an offer.
+    """,
```
dni:
```suggestion
    description="This endpoint can be used to enable an offer.",
```

lnbits/core/views/payment_api.py
```diff
@@ -75,6 +77,33 @@
...
+    """,
+    status_code=HTTPStatus.CREATED,
+    responses={
+        520: {"description": "Fetch invoice error."},
+    },
+)
+async def api_payments_fetch_invoice(
+    fetchinvoice: FetchInvoice,
+    key_info: WalletTypeInfo = Depends(require_base_invoice_key),
+) -> str | None:
```
dni: this should return a proper pydantic model, and also thor HttpException when an invoice could no be fetched.

lnbits/core/views/payment_api.py
```diff
@@ -75,6 +77,33 @@
 payment_router = APIRouter(prefix="/api/v1/payments", tags=["Payments"])
 
 
+@payment_router.post(
```
dni: i think this should be in views/offer_api

dni:
hi, i am not fully done with my review yet but i think there is already enough to do. its a big pr :). it already looks pretty good but i still see some things to do.

lnbits/core/crud/offers.py
21M4TW: I was getting a warning about SQL injection without it, it did not like the `{db.timestamp_placeholder("now")}` ( which I got from lnbits/core/crud/payments.py)

lnbits/core/crud/payments.py
21M4TW: With Bolt12 the dynamic is a bit different, invoices are created from fetch invoices commands that are not seen by lnbits at the time they are actually created, c.f. lnbits/tasks.py line 273 (because offers are created through lnbits, but not the related invoices that can be created at an unknown time in the future, and these invoices don't get announced by CLN). This is why I had to add these arguments. So the invoices have to be created after the fact, once a payment is received, because only then lnbits becomes aware of them.

lnbits/core/crud/payments.py
21M4TW: C.f. my comment above

21M4TW:
> hi, i am not fully done with my review yet but i think there is already enough to do. its a big pr :). it already looks pretty good but i still see some things to do.

Ok thanks for the feedback! I left a few comments where I had some applicable feedback right away, but I will get working on the code w.r.t. your other comments.

lnbits/core/models/offers.py
21M4TW: I just tried this but then I can no longer create an instantiate an instance of Offer using `active`

lnbits/core/crud/offers.py
21M4TW: Done

lnbits/core/views/offer_api.py
21M4TW: Done

lnbits/core/services/payments.py
21M4TW: Done

21M4TW:
@dni I just did a push to address your feedback (except for the couple of places I could,not, see my comments).

lnbits/core/crud/offers.py
21M4TW: Done

lnbits/core/views/payment_api.py
21M4TW: Done

lnbits/core/views/payment_api.py
21M4TW: Done

lnbits/core/views/offer_api.py
21M4TW: Done

lnbits/core/views/offer_api.py
21M4TW: Done

lnbits/core/models/offers.py
21M4TW: Just removed it

21M4TW:
I now have a new extension that leverages this PR. It is called [CLNREST](https://github.com/21M4TW/clnrest) and is available in [my repository](https://raw.githubusercontent.com/21M4TW/lnbits-extensions/main/extensions.json). It gives LNbits a CLN REST interface that can be used instead of LNDHub with Zeus. It supports all common Bolt11 and Bolt12 wallet operations, including offer management.

btcforplebs:
This might be the wrong place for this, but I'm just gonna throw it here. 

Does this mean that even if I have LND as my funding source, I can use bolt 12?

21M4TW:
> This might be the wrong place for this, but I'm just gonna throw it here.
> 
> Does this mean that even if I have LND as my funding source, I can use bolt 12?

The code for this PR does not have CLN-specific code, except for the funding source. Currently in the PR only the corelightning funding source has been modified. However if there is demand for it I can add support for other funding sources when possible (please @dni let me know if I should keep it in a separate branch / outside the current PR).

@btcforplebs, I would need some additional info from you as I am not experienced with LND:
- I assume you are using LNDK for Bolt12 support with LND
- Which LND funding source are you using with LNbits?
- Do you happen to know if LNDK would be accessible through the protocol used by the LND funding source you are using (RPC or REST)?

motorina0:
See https://github.com/lnbits/lnbits/issues/2581#issuecomment-5775805213

PR #3664, opened 2025-12-16
feat: add the ability to print qrcodes to `<lnbits-qrcode>` component

dni:
this print is used in extensions like lnurlp. were now there is a `print_qr.html` page which renders the qrcode component specifically for this purpose. this will replace that page entirely and also fix an issue where only the bech32 version was printable.

- add print icon
- dynamic qr code print page

screenshot-1765875180

https://github.com/user-attachments/assets/f80459e7-fd08-49ed-b071-e6e1df13b77f


lnbits/templates/components.vue
```diff
@@ -534,7 +534,7 @@ include('components/lnbits-error.vue') %}
       <q-tab name="bech32" icon="qr_code" label="bech32"></q-tab>
       <q-tab name="lud17" icon="link" label="url (lud17)"></q-tab>
     </q-tabs>
-    <lnbits-qrcode :value="lnurl" nfc="true"></lnbits-qrcode>
+    <lnbits-qrcode :value="lnurl" nfc="true" print="true"></lnbits-qrcode>
```
dni, line 537: default use the print for the qrcode-lnurl component

talvasconcelos (approved):

PR #3668, opened 2025-12-16
CI: create github prerelease and appimage

dni:
having a built appimage ready for testing on the release is very convinient


motorina0 (approved):
LGTM

PR #3676, opened 2025-12-17
Make funding source fields more explicit

arcbtc:
Makes Boltz more explicit, but also adds an advanced expansion element that other funding sources could use.

image
image

### DIFF
--- a/lnbits/templates/components/admin/funding_sources.vue
+++ b/lnbits/templates/components/admin/funding_sources.vue
@@ -50,6 +50,24 @@
               :hint="prop.hint"
               :value="prop.value"
             >
+              <q-btn
+                v-if="prop.copy"
+                @click="utils.copyText(formData[key])"
+                icon="content_copy"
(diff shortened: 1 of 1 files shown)


lnbits/static/js/components/admin/lnbits-admin-funding-sources.js
```diff
@@ -160,12 +166,32 @@ window.app.component('lnbits-admin-funding-sources', {
...
+            },
+            boltz_client_password: {
+              label: 'Wallet Password (optional)',
+              advanced: true
+            },
             boltz_mnemonic: {
-              label: 'Liquid mnemonic (copy into greenwallet)',
+              label: 'Liquid mnemonic (optional)',
+              hint: 'Boltz will fetch once connected, but you can change later (can be opened in a liquid wallet) ',
+              advanced: true,
```
dni: seed should be visible immedialy

PR #3699, opened 2026-01-03
Fix: Update apipayments updated_at field when a payment is updated.

21M4TW:
apipayments table updated_at field was not updated when updating a payment

### DIFF
--- a/tests/unit/test_crud_payments.py
+++ b/tests/unit/test_crud_payments.py
@@ -28,7 +28,7 @@ async def update_payments(payments):
 
 
 @pytest.mark.anyio
-async def test_crud_get_payments():
+async def test_crud_get_payments(app):
 
     user = await create_user_account()
     wallet = await create_wallet(user_id=user.id)


lnbits/core/crud/payments.py
```diff
@@ -314,6 +314,7 @@ async def update_payment(
     new_checking_id: str | None = None,
     conn: Connection | None = None,
 ) -> None:
+    payment.update_timestamp()
```
dni:
```suggestion
    payment.updated_at = datetime.now(timezone.utc)
```

dni: i would prefer it like that without property, we have it like that in other places.

motorina0:
There are 2 functions where `updated_at` must be updated (`update_payment`, `update_payment_checking_id`).
These are the only 2 places. No need for a dedicated function (which will not be called outside of `crud`).
Is sufficient a simple:
```py
    payment.updated_at = datetime.now(timezone.utc)
```
(or the equivalent SQL update)


Note: its ok not to update the value for `mark_webhook_sent`

motorina0:
> There are 2 functions where `updated_at` must be updated (`update_payment`, `update_payment_checking_id`). These are the only 2 places. No need for a dedicated function (which will not be called outside of `crud`). Is sufficient a simple:

Done. Pushed changes.

tests/unit/test_crud_payments.py
```diff
@@ -39,6 +42,15 @@ async def test_crud_get_payments(app):
     filters = Filters(limit=100)
     payments = await get_payments(wallet_id=wallet.id, filters=filters)
     assert len(payments) == 22, "should return 22 successful payments"
+    first_payment = payments[0]
+    first_payment_updated_at = first_payment.updated_at.replace()
```
dni, line 46: i dont quite understand this, why are you replace it first? maybe i am done i dont now what `Date.replace()` does its not obvious for me

lnbits/core/crud/payments.py
```diff
@@ -304,8 +305,16 @@ async def update_payment_checking_id(
...
+        f"""
+            UPDATE apipayments
+            SET checking_id = :new_id, updated_at = {db.timestamp_placeholder('now')}
+            WHERE checking_id = :old_id
+        """,  # noqa: S608
+        {
+            "new_id": new_checking_id,
+            "old_id": checking_id,
+            "now": int(time()),
+        },
```
dni, line 317: what if we refactor this whole thing to use `update_payment()`?

21M4TW:
`update_payment_checking_id` is only called by `update_payment`, and `updated_at` gets updated twice if `checking_id` gets changed? Why not getting rid of `update_payment_checking_id` and put the code within `update_payment`, with `updated_at` changed only once?

PR #3701, opened 2026-01-04
fix(lndrest): improve error handling for better debugging

PatMulligan:
Related to #3700 

## Summary
- Adds specific `httpx.HTTPStatusError` handling to extract LND error messages
- Adds specific `httpx.ConnectError/ConnectTimeout` handling with context
- Improves generic exception handling to include actual error messages

## Problem
Previously, all exceptions in `status()`, `create_invoice()`, and `pay_invoice()` returned a misleading "Unable to connect" message, even for HTTP 400/401/etc errors. This made
debugging difficult as the actual error from LND was hidden.

Example: A 400 Bad Request with LND's helpful error message `"proto: invalid value for bool type: 1"` would be shown to users as just `"Unable to connect to https://..."`.

## Solution
Now errors display the HTTP status code and LND's error message:
- `HTTP 400: proto: invalid value for bool type: 1`
- `Unable to connect to https://...: Connection refused`
- `Payment error: <actual exception>`

## Test plan
- [ ] Test with valid LND connection
- [ ] Test with invalid macaroon (should show 401/403)
- [ ] Test with LND offline (should show connection error)

### DIFF
--- a/lnbits/wallets/lndrest.py
+++ b/lnbits/wallets/lndrest.py
@@ -26,6 +26,42 @@ from .base import (
 from .macaroon import load_macaroon
 
 
+def _parse_http_error(exc: httpx.HTTPStatusError) -> str:
+    """Extract error message from HTTPStatusError response."""
+    error_message = f"HTTP {exc.response.status_code}"
+    try:
+        error_body = exc.response.json()
+        if "message" in error_body:
+            error_message = f"{error_message}: {error_body['message']}"
+        else:
+            error_message = f"{error_message}: {exc.response.text[:200]}"
+    except Exception:
+        error_message = f"{error_message}: {exc.response.text[:200]}"
+    return error_message
+
+
+def _handle_payment_status(payment: dict, checking_id: str) -> PaymentResponse:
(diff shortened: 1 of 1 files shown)


motorina0:
Lint is failing.

PatMulligan:
hey i'm gonna switch this over to a draft until I can get a moment to take a deeper look at it and re-submit :ok_hand:

lnbits/wallets/lndrest.py
```diff
@@ -26,6 +26,42 @@
...
+            fee_msat=abs(int(payment.get("fee_msat", 0))),
+            preimage=payment.get("payment_preimage"),
+        )
+    if status == "FAILED":
+        reason = payment.get("failure_reason", "unknown reason")
+        return PaymentResponse(ok=False, checking_id=checking_id, error_message=reason)
+    if status == "IN_FLIGHT":
+        return PaymentResponse(ok=None, checking_id=checking_id)
+    return PaymentResponse(
+        ok=False,
```
dni, line 59: just to be safe, lets return ok=None for pending here. and maybe log an error with the status.

Issue #3694, opened 2025-12-30
[BUG] Extension search button not working on Mobile


Michilis:
**Describe the bug**
When going to the extensions page clicking the search button doesn't work on mobile. it does nothing.
(On Installed/All/featured tabs)

- LNbits version: 1.4.0

PR #3705, opened 2026-01-07
fix: search input on extension page

talvasconcelos:
prevent search input from collapsing in mobile

- remove input on mobile
- add input bellow on icon click

image

image

### DIFF
--- a/lnbits/templates/pages/extensions.vue
+++ b/lnbits/templates/pages/extensions.vue
@@ -83,6 +83,7 @@
                 dense
                 class=""
                 v-model="searchTerm"
+                autofocus
               >
                 <template v-slot:append>
                   <q-icon


motorina0:
Is this PR related to: https://github.com/lnbits/lnbits/issues/3694 ?

lnbits/templates/pages/extensions.vue
```diff
@@ -19,6 +19,7 @@
                 :dense="dense"
                 class="float-right q-pr-xl"
                 v-model="searchTerm"
+                :style="$q.screen.lt.sm ? 'min-width: 50%' : ''"
```
dni: hm is there something better? 

Image

at least its visible, i guess

talvasconcelos: yes. can't actually find a "better" way, as it shrinks the input on mobile!

motorina0: How about just keeping the magnifying glass icon and on tap expand the card and show the `q-input` (below tab names.

dni (approved):

motorina0:
Please auto-focus on the search input when the user taps the icon. Maybe give a border to the input in this case (to be more visible).

Other than that it looks good.

image

PR #3725, opened 2026-01-16
[feat] Update extension builder for 1.4.1

motorina0:


lnbits/core/services/extensions_builder.py
```diff
@@ -195,21 +195,38 @@ def _copy_ext_stub_to_build_dir(
...
+            ".git",
+            ".env",
+            ".venv",
+            "*.env",
+            "*.log",
+            "node_modules",
+            ".mypy_cache",
+            ".pytest_cache",
+            ".ruff_cache",
+            ".github",
```
dni: there are the workflow inside this dir right? we should not ignore thjem

motorina0: `".github"` added back

Issue #3728, opened 2026-01-19
[BUG] Missing timezone information in responses when using PostgreSQL


dsbaars:
**Describe the bug**
When using PostgreSQL as backend, timestamps are stored without timezone in the database. This also results in the API responses not including timezone information and therefore show UTC times as local in the UI. 

**To Reproduce**

1. Be in a non UTC native timezone
2. Install LNbits with PostgreSQL (tested with v1.4.0 and dev branch)
3. Credit to wallet 
4. Notice that the relative time showing is immediately off (e.g. 1 hours ago in my case)
5. Notice that in the API responses the timestamps are different than when using an SQLite backend.

Example timestamp with Postgres: `2026-01-19T14:26:03.720394`
Example timestamp with SQLite: `2026-01-19T10:00:25.174844+00:00`

**Expected behavior**
A correct representation of the time in both the Wallet UI and API responses.

**Desktop (please complete the following information):**

- LNbits version: 1.4.0 and dev
- Database postgres

motorina0:
The dates are always saved in UTC date. This ensures that they are stored consistently and can be compared.

It is the job of the client-side app to show the correct time based on the user's timezone.

Example:
```js
moment.utc(expireDate).local().fromNow()
```

motorina0:
See on https://demo.lnbits.com
The date on the UI is correctly displayed based on your timezone.

dsbaars:
@motorina0 you misunderstand the problem. The problem is not that they are stored incorrectly, but that when retrieving the UTC date from a Postgres backend the timezone information `+00:00` is missing so moment does interpret the UTC timestamp as local time. 

My PR #3729 makes sure it's shown the same format as when a SQLite backend used, also when consuming API responses it provides more clarity for the consuming app that the timestamp needs to be interpreted as UTC.

motorina0:
PR merged.

dni:
closes #3728

next time add this to the pr will auto close this issues

PR #3729, opened 2026-01-19
[fix] Fix missing timezone information in responses when using PostgreSQL

dsbaars:
See #3728


dni:
@dsbaars can you maybe try to write a test showing the issue and the fix.

dsbaars:
@dni added test to `test_db.py`, tests for all environments if a timezone is return in UTC. Fails outside the feature branch for postgres, succeeds in it.

PR #3736, opened 2026-01-21
fix: better validation error messages

motorina0:
Before:
image

After:
image

### DIFF
--- a/tests/api/test_api.py
+++ b/tests/api/test_api.py
@@ -245,7 +245,7 @@ async def test_create_invoice_validates_used_currency(
     )
     assert response.status_code == 400
     res_data = response.json()
-    assert "The provided unit is not supported" in res_data["detail"]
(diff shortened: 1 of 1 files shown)


lnbits/exceptions.py
```diff
@@ -107,9 +108,10 @@ async def validation_exception_handler(
         request: Request, exc: RequestValidationError
     ):
         logger.error(f"RequestValidationError: {exc!s}")
+        errors = [dict(e) for e in exc.errors()]
         return render_html_error(request, exc) or JSONResponse(
             status_code=HTTPStatus.BAD_REQUEST,
-            content={"detail": str(exc)},
+            content={"detail": json.dumps(errors)},
```
dni: lets not send the json string down the line, either 'str' as error message or an array of errors

PR #3740, opened 2026-01-26
feat: wallet featured button

motorina0:
Allow admins to show an optional on the user's wallet page.

**Settings:**
image

**Contact Paid Support**
image

image

**Loop To Onchain**
image

**Admin view:**
image


lnbits/settings.py
```diff
@@ -252,6 +252,11 @@ class ThemesSettings(LNbitsSettings):
     )
     lnbits_show_home_page_elements: bool = Field(default=True)
     lnbits_default_wallet_name: str = Field(default="LNbits wallet")
+
+    lnbits_wallet_featured_button_label: str = Field(default="")
+    lnbits_wallet_featured_button_url: str = Field(default="")
```
dni: can we use `str | None` i dont like the use of empty string. you would also need to check that empty string in the frontend then

motorina0: done

arcbtc (approved):

PR #3741, opened 2026-01-26
[feat] admin impersonate user

motorina0:
Allow the admins to impersonate other users in order to inspect or configure their account.

https://github.com/user-attachments/assets/d926e6f0-e65e-4d8e-a8a3-64ff363f5dea


arcbtc:
Oh wow, this is cool

Issue #3416, opened 2025-10-16
[BUG] Changing the theme is cached in the browser after switching accounts


EdiWeeks:
**Description**
Changing the theme is cached in the browser after switching accounts

**To Reproduce**

1. Log in to LNbits 
2. Change theme
3. Log out and log in with another account
4. The theme is was set previously for the previous account shows for current account

**Expected behavior**
Always show correct theme for each account

**Desktop**

- LNbits version: 1.2.1

motorina0:
Theme should also be saved on the back-end so it is persisted from a device/browser to another.

motorina0:
https://github.com/lnbits/lnbits/pull/3743

PR #3743, opened 2026-01-27
[feat] persist user ui customization

motorina0:
### Summary
Persist changes made for the UI.
When the user logs in in a different browser/machine he should have the same UI config (theme, language, bg image, etc)

Closes https://github.com/lnbits/lnbits/issues/3416

image


lnbits/core/models/users.py
```diff
@@ -288,6 +289,7 @@ class User(BaseModel):
     fiat_providers: list[str] = []
     has_password: bool = False
     extra: UserExtra = UserExtra()
+    ui_customization: dict = Field(default_factory=dict)
```
dni, line 292:
```suggestion
    ui_customization: UiCustomization = UiCustomization()
```

i would reprefer having a type here.

motorina0: 1. backeend should not know too much about the UI
2. different UIs might be used for the same backed (dict gives more freedom)

PR #3747, opened 2026-01-29
refactor: `payment.check_status()` into  `check_payment_status(payment)`

### LINKED ISSUE #3284: refactor: `payment.get_status()` into a payments/service function
get_status is the only exception we have were we call fundingsource actions from the models not like usually from a service. this is a bit confusing and inconsistent

dni:
closes #3284

grepping extensions shows only 2 use it.

```
dni@arch:~/repos/lnbits/lnbits-extensions/extensions (main) $ rg check_status                                                                                               1
nostrnip5/services.py
453:    status = await payment.check_status()

satspay/helpers.py
93:            status = await payment.check_status()
```

### DIFF
--- a/lnbits/core/models/payments.py
+++ b/lnbits/core/models/payments.py
@@ -6,23 +6,16 @@ from typing import Literal
 
 from fastapi import Query
 from lnurl import LnurlWithdrawResponse
+from loguru import logger
 from pydantic import BaseModel, Field, validator
 
 from lnbits.db import FilterModel
-from lnbits.fiat import get_fiat_provider
 from lnbits.fiat.base import (
-    FiatPaymentFailedStatus,
-    FiatPaymentPendingStatus,
     FiatPaymentStatus,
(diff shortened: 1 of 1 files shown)


lnbits/core/models/payments.py
```diff
@@ -130,6 +130,7 @@ def is_internal(self) -> bool:
             "fiat_"
         )
 
+    # DEPRECATED: in v1.5.0, use service check_payment_status instead
     async def check_status(
         self, skip_internal_payment_notifications: bool | None = False
     ) -> PaymentStatus:
```
motorina0, line 129:
```suggestion
async def check_status(
self, skip_internal_payment_notifications: bool | None = False
) -> PaymentStatus:
       logger.warning("This method will be deprecated, please use...")
       from lnbits.core.services.payments import check_payment_status

       return await check_payment_status(
        payment, skip_internal_payment_notifications=True
    )
```

motorina0 (approved):

PR #3748, opened 2026-01-29
feat: WIP adding root path functionality

### LINKED ISSUE #1941: [BUG]Lnbits only works on root path
**Describe the bug**
Upon attempting to deploy LNbits via docker behind an nginx reverse proxy on a subpath ("test.kukks.org/lnbits"), LNBits does not function as it assume it is always deployed on the root of the domain `test.kukks.org`

**To Reproduce**
Deploy lnbits and nginx and make nginx load lnbits under a path 
```
location /lnbits/ {
  proxy_pass http://lnbits:5000;
}
```

**Expected behavior**
Lnbits should detect the subpath and use that to load the static files from the right path, along with using it internally when calling APIs and generating links.

**Additional context**
I've managed to do some nginx hacks to make the static files load
```
proxy_set_header Accept-Encoding "";
sub_filter_types *;
sub_filter_once off;
sub_filter '"/core/static'  '"/lnbits/core/static';
sub_filter '"/static'  '"/lnbits/static';
proxy_pass http://lnbits:5000/;
```

This will make the initial site load but will fail upon creating a wallet as the link generated is hardcoded to the root of the domain.

See this code:
```
    href: {
        createWallet: function(t, e) {
            window.location.href = "/wallet?" + (e ? "usr=" + e + "&" : "") + "nme=" + t
        },
        updateWallet: function(t, e, n) {
            window.location.href = `/wallet?usr=${e}&wal=${n}&nme=${t}`
        },
        deleteWallet: function(t, e) {
            window.location.href = "/deletewallet?usr=" + e + "&wal=" + t
        }
    },
```

I think you should be setting the base href and making links use that instead of "/xyz"

Example
```
<base href="/lnbits/">

...

<button onclick="window.location.href = 'deletewallet'"> lolz</button>
```

This will take you to `test.kukks.org/lnbits/deletewallet`

I can use nginx rewriting magic to add the base href myself, but I think you should still include a standard "/" value if you cant make it work automagically

nginx magic i used now to add the base href (super hacky)
```
		sub_filter '<title>LNbits</title>'  '<title>LNbits</title><base href="/lnbits/"/>';
```

dni:
closes #1941

- add --root-path to lnbits application start to use in uvicorn
- utils.urlFor is aware of root-path
- api.request is aware of root-path

TODO: dynamic component loading kinda works, but older extension loading is curently broken because of path issues on `normalize_path`

### DIFF
--- a/lnbits/core/views/generic.py
+++ b/lnbits/core/views/generic.py
@@ -32,7 +32,7 @@ generic_router = APIRouter(
 
 @generic_router.get("/favicon.ico", response_class=FileResponse)
 async def favicon():
-    return RedirectResponse(settings.lnbits_qr_logo)
+    return RedirectResponse(settings.root_path + settings.lnbits_qr_logo)
 
 
 @generic_router.get("/robots.txt", response_class=HTMLResponse)
--- a/lnbits/decorators.py
+++ b/lnbits/decorators.py
@@ -517,6 +517,7 @@ async def _check_account_api_access(
         raise HTTPException(HTTPStatus.FORBIDDEN, "Method not allowed.")
 
 
+# TODO: this messes up my extension urls
 def url_for_interceptor(original_method):
     def normalize_url(self, *args, **kwargs):
         url = original_method(self, *args, **kwargs)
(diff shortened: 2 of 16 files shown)


lnbits/static/js/components/lnbits-theme.js
```diff
@@ -127,7 +127,7 @@ window.app.component('lnbits-theme', {
     if (this.g.mobileSimple === true) {
       document.body.classList.add('mobile-simple')
     }
-    Object.entries(this.g.user.uiCustomization || {}).forEach(
+    Object.entries(this.g.user?.uiCustomization || {}).forEach(
```
dni, line 174: bug on dev

PR #3749, opened 2026-01-29
[feat] User activation

motorina0:
Allow the admins to manually activate/deactivate users.

By default, on register, an user is `activated` (this will be customizable in future PRs).
If the user is `deactivated` by the admin then it won't be able to do any actions (pay, receive, create wallet, install ext, etc)

Activate/Deactivate toggle on the users list page
image

Admin toggle has been moved to the details page since it isn't used very often
image

### DIFF
--- a/lnbits/core/crud/users.py
+++ b/lnbits/core/crud/users.py
@@ -98,19 +98,17 @@ async def get_accounts(
 
 
 async def get_account(
-    user_id: str, activated: bool | None = True, conn: Connection | None = None
+    user_id: str, active_only: bool = True, conn: Connection | None = None
 ) -> Account | None:
     if len(user_id) == 0:
         return None
 
-    activate_clause = "" if activated is None else "AND activated = :activated"
-
     return await (conn or db).fetchone(
-        f"""
+        """
             SELECT * FROM accounts
-            WHERE id = :id {activate_clause}
-        """,  # noqa: S608
-        {"id": user_id, "activated": activated},
+            WHERE id = :id AND (activated = true OR activated = :activated)
+        """,
+        {"id": user_id, "activated": active_only},
         Account,
     )
 
@@ -136,7 +134,7 @@ async def delete_accounts_no_wallets(
 
 
 async def get_account_by_username(
-    username: str, activated: bool = True, conn: Connection | None = None
+    username: str, active_only: bool = True, conn: Connection | None = None
 ) -> Account | None:
     if len(username) == 0:
         return None
@@ -144,28 +142,32 @@ async def get_account_by_username(
     return await (conn or db).fetchone(
         """
             SELECT * FROM accounts
-            WHERE LOWER(username) = :username AND activated = :activated
+            WHERE
+                LOWER(username) = :username
+                AND (activated = true OR activated = :activated)
         """,
-        {"username": username.lower(), "activated": activated},
+        {"username": username.lower(), "activated": active_only},
         Account,
     )
 
 
 async def get_account_by_pubkey(
-    pubkey: str, activated: bool | None = True, conn: Connection | None = None
+    pubkey: str, active_only: bool = True, conn: Connection | None = None
 ) -> Account | None:
     return await (conn or db).fetchone(
         """
             SELECT * FROM accounts
-            WHERE LOWER(pubkey) = :pubkey AND activated = :activated
+            WHERE
+                LOWER(pubkey) = :pubkey
+                AND (activated = true OR activated = :activated)
         """,
-        {"pubkey": pubkey.lower(), "activated": activated},
+        {"pubkey": pubkey.lower(), "activated": active_only},
         Account,
     )
 
 
 async def get_account_by_email(
-    email: str, activated: bool = True, conn: Connection | None = None
+    email: str, active_only: bool = True, conn: Connection | None = None
 ) -> Account | None:
     if len(email) == 0:
         return None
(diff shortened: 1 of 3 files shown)


lnbits/core/crud/users.py
```diff
@@ -93,16 +115,33 @@ async def get_accounts(
...
+) -> Account | None:
     if len(user_id) == 0:
         return None
+
     return await (conn or db).fetchone(
-        "SELECT * FROM accounts WHERE id = :id",
-        {"id": user_id},
+        f"""
+            SELECT * FROM accounts
+            WHERE id = :id {_activated_clause(activated)}
```
dni: why doing this? and having to take a linting issue. this should be a plain sql string as much as possible.

dni:
```
async def update_account(account: Account, conn: Connection | None = None) -> Account:
    account.updated_at = datetime.now(timezone.utc)
    await (conn or db).update("accounts", account)
    return account
```  

this pr adds a db column to accounts but does not add it to the model `Account` so the above wont work.

lnbits/core/crud/users.py
```diff
@@ -36,11 +38,30 @@ async def update_account(account: Account, conn: Connection | None = None) -> Ac
     return account
 
 
+async def update_account_activation(
```
dni: cant we use `update_account()` usually the caller that has a user_id has the Account aswell

lnbits/core/views/auth_api.py
```diff
@@ -83,9 +84,13 @@ async def login(data: LoginUsernamePassword) -> JSONResponse:
         raise HTTPException(
             HTTPStatus.FORBIDDEN, "Login by 'Username and Password' not allowed."
         )
-    account = await get_account_by_username_or_email(data.username)
+    account = await get_account_by_username_or_email(data.username, activated=None)
     if not account or not account.verify_password(data.password):
         raise HTTPException(HTTPStatus.UNAUTHORIZED, "Invalid credentials.")
+
+    is_activated = await is_account_activated(account.id)
```
dni: use `account.activated` instead of making another crud call

lnbits/core/views/user_api.py
```diff
@@ -220,6 +222,42 @@ async def api_users_toggle_admin(user_id: str) -> SimpleStatus:
...
+        raise HTTPException(
+            status_code=HTTPStatus.BAD_REQUEST,
+            detail="Cannot deactivate super user.",
+        )
+    if user_id == account.id:
+        raise HTTPException(
+            status_code=HTTPStatus.BAD_REQUEST,
+            detail="You cannot deactivate yourself.",
+        )
+
```
dni, line 241: just `get_account` and `update_account` here no need for more crud functions

dni: its 2 sql calls in both ways

lnbits/core/crud/users.py
motorina0: activated:

`True` = activated
`False` = deactivated
`None` = either

motorina0:
> this pr adds a db column to accounts but does not add it to the model `Account` so the above wont work.

It does work.
The user should not be able to update the `activated` flag via a simple update.
The `activated` flag should not be present in the Account/User class (it is not used by the user itself)

lnbits/core/crud/users.py
```diff
@@ -216,12 +264,31 @@ async def update_user_access_control_list(
...
     return user_acls or UserAcls(id=user_id)
+
+
+async def clear_user_id_cache(user_id: str):
+    user = await get_user(user_id, activated=None)
+    if user:
+        clear_user_cache(user)
+
+
+def clear_user_cache(user: User):
```
dni, line 274: i dont think, the belong into the `crud.py`, it has nothing to do with crud operations

lnbits/core/crud/users.py
```diff
@@ -124,55 +154,72 @@ async def delete_accounts_no_wallets(
...
         """,
-        {"value": username_or_email.lower()},
+        {"value": username_or_email.lower(), "activated": activated},
         Account,
     )
 
 
-async def get_user(user_id: str, conn: Connection | None = None) -> User | None:
+async def get_user(
+    user_id: str, activated: bool | None = True, conn: Connection | None = None
```
dni: i don't like that all this `get_user` and `get_account` are using this keyword argument activated now. 

```
user = get_user(user_id, activated = True)
assert user, "activated user does not exist"
```
```
user = get_user(user_id)
assert user.activated, "user is not activated"
```
i think the last one is cleaner. on the first example you don't know if the user does not exist or is just not activated. it also adds complexity into this single sql queries.

dni: i can live with it, but if we keep adding kwargs like the to our `get_` crud function it will get messy pretty quickly.

on querying many users, this makes more sense to only get activated or not.

talvasconcelos (approved):
tested

lnbits/core/crud/users.py
```diff
@@ -93,12 +97,18 @@ async def get_accounts(
     )
 
 
-async def get_account(user_id: str, conn: Connection | None = None) -> Account | None:
+async def get_account(
+    user_id: str, active_only: bool = True, conn: Connection | None = None
+) -> Account | None:
     if len(user_id) == 0:
```
dni, line 103: how can a user_id be empty string? or is it defending against user error, i would consider to raise an Exception if its an empty string or just let the sql query raise it. i don't see the use of masking this error and silently returning `None`. although we know we provided a wrong id i think raising in this case is the right thing.

motorina0: unrelated to this PR

PR #3750, opened 2026-01-29
feat: route whitelist/blacklist

arcbtc:
Security feature middleware to block/allow routes, useful for stack folks. 
* Excludes "static" files for ease of use.
* Routes are all available locally.
* Whitelist has priority over blacklist.

(Based on an example use case by @francismars)

image

image

image


francismars:
Might I suggest a toggle "Block all admin API endpoints" so the user doesn't have to choose (and know which ones) manually, and just leave public the ones that are required for regular use?

motorina0:
What is a real world usecase for this?

motorina0:
Isn't what we really want a `lnbits_user_deactivate_ui` flag that only allows admins to access the UI and blocks regular users?

Only the login page is accessible. Admins can operate as before, but regular users will be redirected to a `UI Disabled` page after login (similar to the error page)

arcbtc:
> Isn't what we really want a `lnbits_user_deactivate_ui` flag that only allows admins to access the UI and blocks regular users?
> 
> Only the login page is accessible. Admins can operate as before, but regular users will be redirected to a `UI Disabled` page after login (similar to the error page)

Some users want to lock down everything, just to air on the side of caution. Also not sure how you would lock down all endpoints using the flag you suggested without breaking public pages (without a huge refactor)

motorina0:
concept NACK

arcbtc:
> concept NACK

Its from feedback from francis. He runs a server where every route is locked down apart from a few public pages.

francismars:
I don't want any LNBits page accessible from the internet, just the API routs that required by LNURL for deposits and withdrawals

PR #3751, opened 2026-01-29
feat: FIRST_INSTALL_TOKEN to help services like Umbrel secure the first_install endpoint

arcbtc:
Optional `poetry run lnbits FIRST_INSTALL_TOKEN="sometoken"`
(Based on @francismars Alby issue: a user might enable LNbits, Umbrel will add the funding source, but the user forgets to complete the first_install page.)

### DIFF
--- a/lnbits/templates/pages/first-install.vue
+++ b/lnbits/templates/pages/first-install.vue
@@ -55,8 +55,15 @@
                 ></q-input>
                 <q-input
                   filled
-                  v-model="loginData.firstInstallToken"
+                  v-model.trim="loginData.firstInstallToken"
+                  :type="loginData.isPwd ? 'password' : 'text'"
                   :label="$t('first_install_token')"
+                  ><template v-slot:append>
+                    <q-icon
+                      :name="loginData.isPwd ? 'visibility_off' : 'visibility'"
+                      class="cursor-pointer"
+                      @click="loginData.isPwd = !loginData.isPwd"
+                    /> </template
                 ></q-input>
                 <q-btn
                   @click="setPassword()"


lnbits/static/js/pages/first-install.js
```diff
@@ -19,18 +20,28 @@ window.PageFirstInstall = {
...
-          username: this.loginData.username,
-          password: this.loginData.password,
-          password_repeat: this.loginData.passwordRepeat
-        })
+        const tokenQuery = this.setupToken
+          ? `?token=${encodeURIComponent(this.setupToken)}`
+          : ''
+        await LNbits.api.request(
+          'PUT',
+          `/api/v1/auth/first_install${tokenQuery}`,
```
motorina0: `tokenQuery` should be in the body

lnbits/static/js/pages/first-install.js
```diff
@@ -19,18 +20,28 @@ window.PageFirstInstall = {
...
+          null,
+          {
+            username: this.loginData.username,
+            password: this.loginData.password,
+            password_repeat: this.loginData.passwordRepeat
```
motorina0:
```suggestion
            username: this.loginData.username,
            password: this.loginData.password,
            password_repeat: this.loginData.passwordRepeat,
            setup_token: this.setupToken
```

lnbits/settings.py
```diff
@@ -983,6 +983,7 @@ class EnvSettings(LNbitsSettings):
     enable_log_to_file: bool = Field(default=True)
     log_rotation: str = Field(default="100 MB")
     log_retention: str = Field(default="3 months")
+    lnbits_setup_token: str = Field(default="")
```
motorina0:
```suggestion
    lnbits_setup_token: str | None = Field(default="", description="xxx")
```

tests/conftest.py
```diff
@@ -66,12 +67,22 @@ async def app(settings: Settings):
     app = create_app()
     async with LifespanManager(app) as manager:
         settings.first_install = True
+        request = Request(
```
motorina0: this change should not be needed if `SETUP_TOKEN` is empty

lnbits/middleware.py
```diff
@@ -238,11 +238,20 @@ async def block_allow_ip_middleware(request: Request, call_next):
...
     async def first_install_middleware(request: Request, call_next):
-        if (
-            settings.first_install
-            and request.url.path != "/api/v1/auth/first_install"
-            and request.url.path != "/first_install"
-            and not request.url.path.startswith("/static")
-        ):
-            return RedirectResponse("/first_install")
+        if settings.first_install:
+            if settings.lnbits_setup_token:
```
motorina0: `lnbits_setup_token` should only be checked once, then `settings.first_install` is `True` forever

lnbits/decorators.py
```diff
@@ -546,9 +546,21 @@ async def check_extension_builder(
...
+        )
+
+
+async def check_first_install(request: Request) -> None:
     if not settings.first_install:
         raise HTTPException(
             status_code=HTTPStatus.BAD_REQUEST,
             detail="Super user account has already been configured.",
         )
+    validate_setup_token(request)
```
motorina0: No need to have this check.
`settings.first_install` is set to `True` only if a valid token was provided on the `/first_install` page

lnbits/static/js/pages/first-install.js
```diff
@@ -2,6 +2,7 @@ window.PageFirstInstall = {
   template: '#page-first-install',
   data() {
     return {
+      setupToken: '',
```
motorina0: nice to have a text input for this after the password fields

lnbits/core/views/auth_api.py
```diff
@@ -32,6 +32,7 @@
     check_account_exists,
     check_admin,
     check_user_exists,
+    validate_setup_token,
```
motorina0: This function can be a private function here.
It should not be in decorators.

lnbits/settings.py
motorina0: `lnbits_first_install_token` or `lnbits_first_install_secret` would be more suggestive and easier to understand

lnbits/core/models/users.py
```diff
@@ -358,6 +358,7 @@ class UpdateSuperuserPassword(BaseModel):
     username: str = Query(default=..., min_length=2, max_length=20)
     password: str = Query(default=..., min_length=8, max_length=50)
     password_repeat: str = Query(default=..., min_length=8, max_length=50)
+    token: str | None = Query(None)
```
motorina0:
```suggestion
    first_install_token: str | None = Query(None)
```

lnbits/core/views/auth_api.py
```diff
@@ -503,6 +503,12 @@ async def update_ui_customization(
 async def first_install(data: UpdateSuperuserPassword) -> JSONResponse:
     if not settings.first_install:
         raise HTTPException(HTTPStatus.FORBIDDEN, "This is not your first install")
+    if settings.first_install_token:
+        if not data.token:
```
motorina0: This should not be mandatory.
Only if `settings.first_install_token` is set, then check the `data.token`

lnbits/static/js/pages/first-install.js
```diff
@@ -7,7 +7,8 @@ window.PageFirstInstall = {
         isPwdRepeat: true,
         username: '',
         password: '',
-        passwordRepeat: ''
+        passwordRepeat: '',
+        token: ''
```
motorina0:
```suggestion
        first_install_token: ''
```

dni (approved):
very nice feature, i simplified it a bit and now you can set an enviroment variable `FIRST_INSTALL_TOKEN` when starting the lnbits server where the token has to be provided on the first_install page to be able to set the password, you can pass it as ?token= to prefill the formfield.

lnbits/static/js/pages/first-install.js
```diff
@@ -17,20 +18,23 @@ window.PageFirstInstall = {
...
-          password_repeat: this.loginData.passwordRepeat
+          password_repeat: this.loginData.passwordRepeat,
+          token: this.loginData.token
         })
-        window.location.href = '/admin'
-      } catch (e) {
-        LNbits.utils.notifyApiError(e)
-      }
+        .then(() => {
+          window.location.href = '/admin'
```
motorina0: prepare for context path (use utils)

lnbits/static/js/pages/first-install.js
```diff
@@ -17,20 +18,23 @@ window.PageFirstInstall = {
...
+        })
+        .catch(this.utils.notifyApiError)
     }
   },
   created() {
+    const params = new URLSearchParams(window.location.search)
+    this.loginData.token = params.get('token') || ''
```
motorina0:
```suggestion
    this.loginData.firstInstallToken = params.get('token') || ''
```

lnbits/static/js/pages/first-install.js
```diff
@@ -17,20 +18,25 @@ window.PageFirstInstall = {
     }
   },
   methods: {
-    async setPassword() {
-      try {
-        await LNbits.api.request('PUT', '/api/v1/auth/first_install', null, {
+    setPassword() {
```
motorina0, line 21: why change from async to Promise?

dni:
screenshot-1769766206
added it to startup log. and also to `.env.example`. 
```console
# === First Install Token ===
# if set the user is required to enter this token on the /first_install page
# FIRST_INSTALL_TOKEN="myaccesstoken"
```

lnbits/static/js/pages/first-install.js
dni: i prefer it that way, having awaits internally adds small overhead and could make functions blocking in some contexts, although i dont think in this context when called from Vue template.

lnbits/core/views/auth_api.py
dni: it is not mandatory. it only check when enviroment variable is set.

francismars:
You guys are awesome

PR #3754, opened 2026-02-01
fix: use absolute paths for dynamic extension routes

BuilderFred:
Fixes a regression where dynamic extension routes (e.g., LNURL-pay links) fail to load assets when accessed via deep links. Prepending the extension ID ensures that template and component paths are absolute and resolve correctly regardless of the URL nesting or proxy configuration.

### DIFF
--- a/lnbits/static/js/init-app.js
+++ b/lnbits/static/js/init-app.js
@@ -26,8 +26,14 @@ const DynamicComponent = {
             path: r.path,
             name: r.name,
             component: async () => {
-              await LNbits.utils.loadTemplate(r.template)
-              await LNbits.utils.loadScript(r.component)
+              const templatePath = r.template.startsWith('/')
+                ? r.template
+                : `/${name}/${r.template}`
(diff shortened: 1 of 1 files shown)


dni:
i do not really understand that change. how does this help, the routes inside the `routes.json` are correct.

BuilderFred:
@dni The issue is that when serving LNbits behind a reverse proxy (specifically on a subdomain) and accessing the deep link `domain.com/lnurlp/link/<id>`, the browser attempts to resolve relative paths like `static/routes.json` against the current base path (`/lnurlp/link/`), resulting in 404s.

By prepending the extension ID (`/lnurlp/static/...`), we ensure the paths are treated as absolute from the root context of the LNbits instance, which allows Nginx/reverse proxies to correctly route the request to the extension's static file handler regardless of the nesting level of the current page.

This mirrors how `manifest.json` and other core assets are referenced to ensure deep-link compatibility.

dni:
i am currently working on a PR where you can add a root-path application wide, that should take care of this issue

PR #3766, opened 2026-02-05
feat: modern theme stuff like rounded corners and card gradient

arcbtc:
image
image


lnbits/static/js/globals.js
```diff
@@ -31,7 +40,7 @@ window.g = Vue.reactive({
   mobileSimple: localStore('lnbits.mobileSimple', true),
   walletFlip: localStore('lnbits.walletFlip', false),
   lastActiveWallet: localStore('lnbits.lastActiveWallet', null),
-  darkChoice: localStore('lnbits.darkMode', true),
+  darkChoice: localStore('lnbits.darkMode', SETTINGS.defaultDark),
```
motorina0, line 43: The var `SETTINGS` should not be present in front-end anymore.
cc: @dni

lnbits/static/js/pages/account.js
```diff
@@ -40,13 +40,7 @@ window.PageAccount = {
         }
       ],
       defaultSiteCustomisation: {
```
motorina0, line 42: This field is no longer used. Should be removed.
Reset to defaults will not reset the language. Is that on purpose?

lnbits/static/js/globals.js
dni: no this is still correct, for localstorage default settings.

lnbits/static/js/pages/account.js
arcbtc: For now, we can tweak later. Reset to defaults works better than it did

PR #3768, opened 2026-02-06
[feat] spark L2 internal side car

motorina0:
The admin has the option to configure a sidecar version. When so, the side car will be downloaded and started as an internal process by LNbits.

If no sidecar version provided then the wallet will look for an eterna sidecar.

Note: `npm` and `node` must be present on the lnbits machine in order for the internal sidecar to run.

**Internal sidecar**:
image

**External sidecar**:
image


blackcoffeexbt:
Works well from my testing.

A couple of UI notes - Some related to "sidecar" term.

- Only the mnemonic field should be a "secret" field. Network and version can be plain text
- Add some conditional validation that makes the mnemonic field a required field if a version is specified
- "If specified then the sidecar will be downloaded. Alternatively you can specify an External Sidecar endpoint in the Advanced section." change to "If specified then this Spark funding source will be downloaded. Alternatively you can specify an "External Spark endpoint" in the Advanced section."
- "Internal Sidecar version (eg: 0.0.1)" Can this change to 'Spark wallet version to use e.g. "0.0.1"'
- "Only required if Interna Sidecar version is specified." Can this change to "Only required if a spark wallet version is specified above"

PR #3779, opened 2026-02-10
fix: SSO disallow insecure HTTP by default

AlnBits:
This changes SSO provider initialization to only allow insecure HTTP redirects in debug/dev mode (settings.debug), instead of always allowing it.

Rationale: safer defaults for production deployments; avoids accidental OAuth over HTTP.

Dev setups can still use HTTP by running with debug enabled.


lnbits/core/views/auth_api.py
```diff
@@ -616,7 +616,7 @@ def _new_sso(provider: str) -> SSOBase | None:
 
         sso_provider_class = _find_auth_provider_class(provider)
         sso_provider = sso_provider_class(
-            client_id, client_secret, None, allow_insecure_http=True
+            client_id, client_secret, None, allow_insecure_http=settings.dev_mode
```
dni, line 625: i would rather do a `settings.sso_allow_insecure`

motorina0:
Closed in favor of: https://github.com/lnbits/lnbits/pull/3801

PR #3791, opened 2026-02-16
Fix: Open QR in new window/tab

talvasconcelos:
If `lnbits-qrcode` component has an `href` prop open in a new tab/window by default. If desired function is open in same window, pass `:newTab=false` as prop.


lnbits/static/js/components/lnbits-qrcode.js
```diff
@@ -24,6 +24,10 @@ window.app.component('lnbits-qrcode', {
       type: String,
       default: ''
     },
+    newTab: {
+      type: Boolean,
+      default: true
```
motorina0, line 29: where is the value of this field changed?

talvasconcelos: `<lnbits-qrcode :newTab="false"></lnbits-qrcode>`

lnbits/static/js/components/lnbits-qrcode.js
```diff
@@ -24,6 +24,10 @@ window.app.component('lnbits-qrcode', {
       type: String,
       default: ''
     },
+    newTab: {
```
dni, line 27:
```suggestion
    openTab: {
```

arcbtc:
nok
this is a horrible idea
whats the point?

PR #3796, opened 2026-02-18
feat: for QRs that are not case sensitive make uppercase

arcbtc:
Only applies to qrs that are not case sensitive.
Makes a huge difference for being able to scan on cheaper phones.

<table>
  <tr>
    <td>
Before:
Screenshot from 2026-02-18 16-05-43
</td>
    <td>
After:
Screenshot from 2026-02-18 16-11-03
</td>
  </tr>
</table>


motorina0:
Closed in favor of: https://github.com/lnbits/lnbits/pull/3798

PR #3800, opened 2026-02-19
refactor: move payment code from `tasks.py` to `service/payments.py`

dni:
precursor for task manager.

move payment code from tasks.py to service/payments.py

on other places in the code we usually inline import from `lnbits.tasks` because of circular import issues, but here it was different, it was importing from payments. that refactor fixes that

### DIFF
--- a/lnbits/core/services/payments.py
+++ b/lnbits/core/services/payments.py
@@ -1076,9 +1076,9 @@ async def _send_payment_notification_in_background(
 
 async def update_invoice_callback(checking_id: str) -> Payment | None:
     """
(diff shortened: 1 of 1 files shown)


lnbits/core/services/payments.py
```diff
@@ -509,6 +509,8 @@ async def update_wallet_balance(
         )
         payment.status = PaymentState.SUCCESS
         await update_payment(payment, conn=conn)
+        from lnbits.tasks import internal_invoice_queue_put
```
dni, line 512: this imports will be removed when taskmanager arrives

PR #3808, opened 2026-02-20
feat: Add a dialog component

talvasconcelos:
Adds am `<lnbits-dialog>` component to standardize UI/UX across LNbits. 

- cancel/close button on the left
- action button on the right
- optional dialog title

It should cover most dialog uses across LNbits!

Tested on TPOS invoice dialog!


dni:
i am really on the fence with this one, it kinda adds friction in creating more custom dialog and its almost the same amount of code. the only advantage is to have them more consistent across lnbits, that can be good and bad when you want to customize it more. i am 50/50 on that one.

PR #3812, opened 2026-02-24
fix: remove OpenSats

arcbtc:
Apparently our application “wasn’t compelling enough,” so we’re no longer supported by an OpenSats grant.

Despite countless sleepless nights, working through weekends, 1.2k GitHub stars, 100k+ users, and countless projects that depend on LNbits in their stack - plus continuous innovation and R&D in the Bitcoin/Nostr space.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
   </picture>
 </a>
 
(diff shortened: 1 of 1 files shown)


motorina0:
Just not yet.

PR #3814, opened 2026-02-25
feat: support for wasm extension bridge

arcbtc:
Facilitates https://github.com/lnbits/wasm parent extension
Once WASM extension is installed extensions like https://github.com/lnbits/paidtasks can be vibe coded, not vetted and shared safely.

In two minds over this pr/concept.
* Extensions with permissions is very useful
* Giving permission for spending can cause a lot of damage and having as a wasm makes it impossible to review
* The wasm not having access to system files, envs and having its processing restricted adds a lot of value.
* Building an extension using the extension builder and python in MUCH easier (maybe the wasm framework just needs a little more work to be made easier)


arcbtc:
This should be merged, its a minimal change, completely backwards compatible, and enables a load more usecases

talvasconcelos:
## Manual test report (WASM extension / PaidTasks)

### What I tested

- Enabled the WASM host extension (`lnbits/wasm`)
- Installed `paidtasks` by cloning it into `lnbits/extensions/`
- Saved the required WASM permissions/payment tag
- Enabled and opened the `paidtasks` extension
- Created/used the extension from the UI
- Tried the public invoice creation flow

### What worked

- Once the WASM host extension is enabled, installing/using a WASM extension is fairly straightforward if you already know your way around LNbits.
- The warning about needing the WASM host extension is useful, even if the extra dependency feels a bit clunky.
- The permissions screen is informative: it shows what the extension is asking for and what enabling those permissions means.

### Issues / concerns

- I don’t really see the point of asking the user to manually approve required permissions one by one. If the extension needs them to function, it should clearly state that these are required capabilities.
- If the user does not allow one of the required permissions, does the extension become half-functional? broken? unclear.
- The useful part is the disclosure: “this extension needs X/Y/Z and here’s what that means”. The checkbox-style permission flow feels less useful/confusing from a user perspective.
- WASM extensions requiring a separate `wasm` host extension is a bit clunky UX, but acceptable if clearly documented/signposted.

### Blocker found

From a user/manual-test perspective, `paidtasks` failed when trying to create an invoice from the public flow.

```
2026-05-25 15:19:39.87 | INFO | 127.0.0.1:45768 - "POST /paidtasks/api/v1/public/call/public_create_invoice HTTP/1.1" 500
```

Expected:

- public task page creates an invoice
- invoice can be paid
- task updates after payment

Actual:

- invoice creation failed with HTTP 500
- could not complete the full paid task flow

The WASM extension flow seems usable for someone already familiar with LNbits, but the permission UX needs clarification and the example extension should work end-to-end.

- clarify the WASM host dependency in docs/UI either on the wasm extensions or filter wasm extensions if the host extension is not available
- present required permissions as required capabilities, not confusing optional choices
- make permission descriptions more user-facing
- fix/investigate the `paidtasks` invoice creation failure

PR #3817, opened 2026-02-28
fix: skip funding source status check for internal payments

### LINKED ISSUE #3816: [BUG] Internal payments zero out receiver wallet balance (CLNRestWallet)
**Describe the bug**

Internal (wallet-to-wallet) payments within the same LNbits instance result in the receiving wallet's balance showing 0 instead of the paid amount. The payment shows as "success" in the transaction list but the balance is not credited.

The root cause is in `invoice_callback_dispatcher` (tasks.py). When an internal payment settles, the dispatcher calls `check_payment_status` on the receiver's payment record. Because the receiver's `checking_id` does not have the `internal_` prefix, `is_internal` returns `False`, so `check_payment_status` falls through to `get_invoice_status` on the funding source (CLNRestWallet).

Since the invoice was paid internally and never settled on CLN, CLN reports `amount_received_msat = 0` while `amount_msat` equals the invoice amount. The fee calculation in `clnrest.py:get_invoice_status`:

```python
fee_msat = data["invoices"][0].get("amount_received_msat", 0) - data["invoices"][0].get("amount_msat", 0)
```

produces `fee_msat = 0 - 1000000 = -1000000` (negative the full invoice amount).

Back in `invoice_callback_dispatcher`:

```python
```

This truthy `-1000000` overwrites the receiver's fee. Since LNbits computes balance as `SUM(amount - ABS(fee))`, the fee cancels the amount exactly: `1000000 - 1000000 = 0`.

**To Reproduce**

1. Set up LNbits with CLNRestWallet backend (CLN via clnrest)
2. Create two wallets (Wallet A, Wallet B)
3. Generate an invoice from Wallet B
4. Pay the invoice from Wallet A
5. Wallet B shows the payment as "success" in the transaction list but balance remains 0

**Expected behavior**

Wallet B's balance should reflect the received amount (e.g. 1000 sats).

- LNbits version: 1.5.0 (also confirmed in v1.4.2 source — same code path)
- Database: SQLite
- Funding source: CLNRestWallet (CLN v25.12.1)

**Additional context**

The bug is in two locations:

1. **`lnbits/wallets/clnrest.py` → `get_invoice_status`**: The fee calculation `amount_received_msat - amount_msat` returns a large negative number for invoices that were never settled on CLN (internal payments). This likely also affects LndRestWallet and other backends with similar fee derivation.

2. **`lnbits/tasks.py` → `invoice_callback_dispatcher`**: This function receives `is_internal=True` as a parameter but never uses it. It relies on `payment.is_internal` which checks the `checking_id` prefix — but only the sender's record has the `internal_` prefix, not the receiver's.

A minimal fix would be to use the `is_internal` parameter that is already passed in `invoice_callback_dispatcher`:

```python
async def invoice_callback_dispatcher(checking_id: str, is_internal: bool = False):
    ...
    if not is_internal:
        status = await check_payment_status(
        )
        payment.preimage = status.preimage or payment.preimage
    await update_payment(payment)
```

We have verified this patch fixes the issue.

macterra:
## Summary

- Fixes internal payments zeroing out receiver wallet balance when using CLNRestWallet
- Uses the existing `is_internal` parameter in `invoice_callback_dispatcher` to skip the funding source status check for internal payments

## Root cause

When an internal payment settles, `invoice_callback_dispatcher` queries the funding source (`get_invoice_status`) for the receiver's invoice. Since the invoice was paid internally and never settled on CLN, `amount_received_msat` is 0 while `amount_msat` equals the invoice amount. The fee calculation:

```python
fee_msat = amount_received_msat - amount_msat  # 0 - 1000000 = -1000000
```

This negative fee overwrites the receiver's payment record. Since balance is computed as `SUM(amount - ABS(fee))`, the fee cancels the amount exactly, resulting in a zero balance.

LND is unaffected because `LndRestWallet.get_invoice_status` returns `PaymentSuccessStatus()` with `fee_msat=None`.

## Fix

The `is_internal` parameter was already passed to `invoice_callback_dispatcher` but never used. This change skips the funding source status check when `is_internal=True`, since the payment has already been verified internally by LNbits.

## Test plan

- [x] Create two wallets on LNbits with CLNRestWallet backend
- [x] Generate invoice from Wallet B, pay from Wallet A
- [x] Verify Wallet B balance reflects the received amount
- [x] Verify external (non-internal) payments still update status correctly

Fixes #3816

### DIFF
--- a/lnbits/tasks.py
+++ b/lnbits/tasks.py
@@ -178,14 +178,15 @@ async def invoice_callback_dispatcher(checking_id: str, is_internal: bool = Fals
         logger.warning(f"Payment '{checking_id}' is not incoming, skipping.")
         return
 
-    from lnbits.core.services.payments import check_payment_status
+    if not is_internal:
+        from lnbits.core.services.payments import check_payment_status
 
-    status = await check_payment_status(
-        payment, skip_internal_payment_notifications=True
-    )
-    payment.fee = status.fee_msat or payment.fee
-    # only overwrite preimage if status.preimage provides it
-    payment.preimage = status.preimage or payment.preimage
+        status = await check_payment_status(
+            payment, skip_internal_payment_notifications=True
+        )
+        payment.fee = status.fee_msat or payment.fee
+        # only overwrite preimage if status.preimage provides it
+        payment.preimage = status.preimage or payment.preimage
     payment.status = PaymentState.SUCCESS
     await update_payment(payment)
     if payment.fiat_provider:


motorina0:
The internal status is the first thing that it is checked in `check_payment_status`.
Fiat payments (Stripe or Paypal) are also internal payments.

```py
async def check_payment_status(
    payment: Payment, skip_internal_payment_notifications: bool | None = False
) -> PaymentStatus:
    if payment.is_internal:
        if payment.success:
            return PaymentSuccessStatus()
        if payment.failed:
            return PaymentFailedStatus()
        if payment.is_in and payment.fiat_provider:
            fiat_status = await check_fiat_status(
                payment, skip_internal_payment_notifications
            )
            return PaymentStatus(paid=fiat_status.paid)
        return PaymentPendingStatus()
     ... funding_source checks
```

I do not see how moving the `if payment.is_internal:` outside he function call solves any issue.

PR #3849, opened 2026-03-04
feat: currency helper

arcbtc:
currently there is no utility just to get a currency symbol from core


lnbits/static/js/utils.js
```diff
@@ -131,6 +131,22 @@ window._lnbitsUtils = {
...
+    }
+    try {
+      const parts = new Intl.NumberFormat(window.i18n.global.locale, {
+        style: 'currency',
+        currency: code
+      }).formatToParts(0)
+      const symbolPart = parts.find(part => part.type === 'currency')
+      return symbolPart?.value || code || '¤'
+    } catch (e) {
+      return code || '¤'
```
dni, line 147: '¤' what is this symbol?

lnbits/static/js/utils.js
```diff
@@ -131,6 +131,22 @@ window._lnbitsUtils = {
       currency: currency || 'sat'
     }).format(value)
   },
+  getCurrencySymbol(currency) {
```
dni, line 134: i thought amount and symbols are already covered by `formatCurrency`, can you provide maybe an example where this is useful outside of displaying amounts? not sure why we add it if its not used anywhere.

lnbits/static/js/utils.js
motorina0: default value in case nothing is found

lnbits/static/js/utils.js
motorina0: Image

PR #3857, opened 2026-03-05
feat: Improvements to settings > routing fee reserve settings UX 

blackcoffeexbt:
Improved UX of the settings > routing fee reserve settings to help users understand how these settings work.

+ Added intro to admin > funding > Routing fee reserve calcs
+ Added better field labels and tooltips

I'll run the i18n on the non en languages when this PR is approved.


motorina0:
Lint fails.

lnbits/static/scss/base.scss
```diff
@@ -207,3 +207,7 @@ video {
   font-weight: 300;
   opacity: 0.4;
 }
+
+.q-tooltip {
```
dni: do not override default quasar classes, there most likely is a variable to change for this. and it can have unexpected consequences in other unintended places.

blackcoffeexbt: The tooltip font size is defined here in Quasar - https://github.com/quasarframework/quasar/blob/20201fb14e9e2dcbd352c6e055ba45f172d4a0f6/ui/src/components/tooltip/QTooltip.sass#L1

This can't be set as a var from LNbits so has to be set in `base.scss`. The tooltip size defined in Quasar is 10px which is far too small. A 14px minimum font size is recommended for interaction-heavy pages.

Should this tooltip text size increase be a standalone PR?

https://www.learnui.design/blog/mobile-desktop-website-font-size-guidelines.html#text-vs-interaction

dni (approved):

PR #3865, opened 2026-03-11
Migrate from pydantic v1 to v2 API

ritchelakwali:
Automated migration using pydantic-migrate tool.

## Changes (S1 Regex Engine)
- 119x .json() to .model_dump_json()
- - 20x .dict() to .model_dump()
- - 1x @root_validator to @model_validator(mode='before')
## Stats
- 39 files changed, 142 insertions, 141 deletions
- - 129/129 Python files pass AST syntax validation
- - Cost: 0.00 EUR (pure regex, no LLM)
- - Runtime: less than 1 second
Tested on jupyter-scheduler first: 126/126 tests passing (100%).

### DIFF
--- a/lnbits/commands.py
+++ b/lnbits/commands.py
@@ -711,7 +711,7 @@ async def _call_install_extension(
         user_id = user_id or get_super_user()
         async with httpx.AsyncClient() as client:
             resp = await client.post(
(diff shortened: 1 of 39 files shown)


dni:
hey! this migration will happen at some point, but currently it is on a future roadmap. so thanks for the efforts. but i will close it for now. 

and it takes a lot more then just replacing the model_dump functions.

PR #3790, opened 2026-02-15
Fix: expired invoices remaining pending indefinitely

karans-repo:
Fixes #3490

I Added scheduled task to automatically delete expired invoices that have passed their expiration time. 

Please give the feedback as Iam a beginner

PR #3867, opened 2026-03-12
fix: mark expired incoming invoices as failed and add periodic cleanup

### LINKED ISSUE #3490: [BUG] Expired receiving invoices remain “pending” indefinitely
**Describe the bug**
Receiving invoices that pass their expiration time remain in a “pending” state indefinitely in the LNbits UI/API. They are not cleared/marked expired automatically.

**To Reproduce**

1. Go to the wallet and create a receiving invoice with a short expiry.

2. Do not pay it and wait past the expiry.

3. Refresh the UI/API.

4. See the invoice still listed as “pending” in UI and database.

**Expected behavior**
Once expired, the invoice should transition to “expired” and no longer appear as payable/pending.

1. LNbits version: v1.3.1

2. Database: PostgreSQL

**Additional context**
If expiration handling depends on a background task, documenting its cadence or adding an admin control to force cleanup would help operators.

emtll:
## Summary

Fixes #3490

Expired incoming invoices remain in `pending` status indefinitely because:

1. **`update_pending_payment()` always queries the funding source**, which returns `PaymentPendingStatus` for invoices that no longer exist on the backend (e.g. LND garbage-collected them after expiry).

2. **`delete_expired_invoices()` exists but is never called** — it's dead code, exported in `crud/__init__.py` but not invoked by any background task.

### Changes

- **`update_pending_payment()`** (`core/services/payments.py`): Added an early check using the existing `Payment.is_expired` property. If an incoming invoice has expired, it is immediately marked as `PaymentState.FAILED` without querying the funding source. This is safe because an expired invoice can never be paid.

- **`register_async_tasks()`** (`app.py`): Registered `delete_expired_invoices` as a periodic task running every hour (same pattern as `purge_audit_data`). This cleans up pending invoices older than 30 days and pending invoices whose `expiry` date has passed.

### Why this approach

- Uses the existing `Payment.is_expired` property and `delete_expired_invoices()` function — no new code needed beyond wiring them up.
- Avoids modifying wallet backends (LND/CLN/etc), which return `PaymentPendingStatus` for missing invoices as a safety measure (backend could have been temporarily offline).
- Follows maintainer feedback from #3790 discussion: marks invoices as failed rather than silently deleting recent ones.

## Test plan

- [ ] Verify expired pending incoming invoices are marked as `failed` during `check_pending_payments` cycle
- [ ] Verify non-expired pending invoices still query the funding source normally
- [ ] Verify outgoing payments are not affected by the expiry check
- [ ] Verify `delete_expired_invoices` runs hourly and cleans stale records
- [ ] Run existing test suite (`make test`)


lnbits/app.py
```diff
@@ -489,6 +490,7 @@ def register_async_tasks() -> None:
 
     create_permanent_task(run_by_the_minute_tasks)
     create_permanent_task(purge_audit_data)
+    create_permanent_task(run_interval(60 * 60, delete_expired_invoices))
```
motorina0: cc: @dni

dni: lets have them deleted inside the check_pending_invoices, make the deletion configurable, default to 1 week after expiry to make sure there is no held htlc left.

motorina0 (approved):

blackcoffeexbt (approved):

Issue #440, opened 2021-11-29
Payment amount + fee can exceed the wallet balance. Not just a few sats, but e.g. 0.01 BTC, repeatedly. Together more than 0.1 was stolen.


gorrdy:
I've been attacked.

An attacker just used LNbits to sweep my channels' balance. He used a trick with a malicious routing node with an enormous routing fee that resulted in an ln payment with the amount of 891,000 sats and a fee ~890,000 sats.

The attack vector is quite simple. From the attacker's perspective:
1. Open a channel with the target node, e.g. 2M sats.
2. Send payment to an LNbits wallet e.g. 900k sats
3. Send a payment out of the LNbits wallet to another destination node with an amount e.g. 891,000 sats.
4. The only path to a destination node is through a malicious node that charges an enormous fee (999,000 ppm).
5. LNbits will ask LND to send the 891,000 sats to the destination node, but LNbits is not aware of the fee (actually it was in my case because it was a private channel from the malicious so the fees WERE in the invoice. If a public channel is used, there is no information about the fee in the invoice, I get that). LND is aware of the fee but has no problem with that.
6. LND will pay the invoice because it has no information (e.g. fee_limit) what is the maximum fee it can "spend" from the LNbits wallet. (E.g. wallet balance is 900,000 sats, the attacker tries to send 891,000, then LNbits should send to LND fee_limit=9,000 sats.)
7. You just lost your sats.

Hope it helps with resolving the issue. **Currently, every LNbits site that is publicly accessible is in danger.**

Transaction detail:
![image](https://user-images.githubusercontent.com/80678558/143843322-b85a363e-36a7-422c-a050-014d2b508293.png)

Transactions in Thunderhub:
![image](https://user-images.githubusercontent.com/80678558/143843472-c31427d0-7276-4f8f-9cce-f8b30b634cbc.png)

callebtc:
Looking into this right now.

callebtc:
Fix for the expensive routing fee attack (#442) for `LndRestWallet` merged into master. Please update your lnbits with git pull. I recommend changing to `LndRestWallet` if you're using `LndWallet`. 

Note: For `LndRestWallet` you need to enter the hex string of your macaroon in the `.env` file. You can get the hex string via `xxd -ps -u -c 1000  /path/to/admin.macaroon`

arcbtc:
Closing as resolved

Kixunil:
[...]

PR #3868, opened 2026-03-12
feat: decouple max routing fee from fee reserve

### LINKED ISSUE #3493: [Feature request] Add explicit “Max Fee (sats)” and “Max Fee Rate (ppm)” fields (decouple from Fee Reserve)
Setting Fee Reserve and Fee Reserve Percent to zero appears to effectively pass max fee = 0 to LND payments, causing routed payments to fail. Admins set reserve=0 to avoid holding funds, not to block routing fees entirely.

- Add explicit admin fields for Max Fee (sats) and Max Fee Rate (ppm).

- Use these fields to populate LND’s max-fee parameters when paying.

- Keep Fee Reserve/Percent strictly for reserve/UX (not for routing fee limits).

Overloading Fee Reserve as a proxy for max fee, but this is confusing and leads to unintended payment failures when set to 0.

**Additional context**
Decoupling reserve from routing fee limit, should eliminate several UX pitfalls.

emtll:
## Summary

Fixes #3493

The `fee_reserve` setting is currently used for two distinct purposes:

1. **Balance check** — how much to hold back from the wallet balance (UX)
2. **Fee limit** — the `fee_limit_msat` sent to the funding source (LND/CLN)

This coupling means setting fee reserve to 0 (to avoid holding user funds) inadvertently tells the backend `max_fee=0`, **blocking all routed payments**. This was noted as far back as PR #447 by @callebtc: *"In my mind, max_fee and fee_reserve are two separate things"*.

### Changes

Adds two new **optional** admin settings to `FeeSettings`:
- `lnbits_fee_max_msat` — explicit max routing fee in msats
- `lnbits_fee_max_percent` — explicit max routing fee as percentage

A new `fee_limit()` method computes the actual fee limit sent to the funding source:
- If the new fields are configured → uses them (decoupled from reserve)
- If left empty (default) → falls back to `fee_reserve()` (full backward compatibility)

| File | Change |
|------|--------|
| `settings.py` | New fields + `fee_limit()` method on `FeeSettings` |
| `core/services/payments.py` | `_pay_external_invoice()` uses `fee_limit()` for funding source; `fee_reserve()` still used for balance check |
| `core/views/payment_api.py` | `/fee-reserve` endpoint also returns `fee_limit` |
| `templates/components/admin/funding.vue` | Two new fields in admin Funding tab |
| `static/i18n/en.js` + bundle | i18n labels |

### Example configuration

| Setting | Value | Purpose |
|---------|-------|---------|
| Fee Reserve (msats) | `0` | Don't hold back any balance |
| Fee Reserve (%) | `0` | Same |
| **Max Routing Fee (msats)** | `5000` | LND accepts up to 5 sats routing fee |
| **Max Routing Fee (%)** | `1.0` | Or 1% of amount, whichever is higher |

Result: user can spend 100% of balance, but LND still accepts routing fees up to `max(5 sats, 1%)`.

## Test plan

- [x] Empty fields (default) → behavior identical to current (backward compatible)
- [x] All fees zeroed → routed payments fail as expected
- [x] Reserve zeroed + Max Routing Fee set → payments succeed with routing
- [x] Unit tests pass (no regressions)
- [x] Admin UI displays new fields correctly with i18n labels

## Security note

The original coupling was introduced after a security incident (#440) where missing `fee_limit` allowed an attacker to steal funds via high routing fees. This PR preserves the safe default (fee_limit = fee_reserve when unconfigured). When an admin explicitly sets Max Routing Fee, they are consciously taking control of the fee limit independently.


motorina0:
The owner of the wallet is the one that has to pay the routing fees.
If `Balance check = 0`, but `Fee limit = 1000` and the actual node (routing/LSP) fee is `900 sats` then the node operator would actually pay the fee.
If the user sends all the funds in the wallet then the wallet would have a negative balance (the fee is part of the payment).

PR #3884, opened 2026-03-20
chore: fix security audit from `uv audit`

dni:
- updates lnurl to not uyse ECDSA

screenshot-1774012919

only starlette left. that takes another deeper look, we breaking some fastapi compatibility, this may need to wait for the pydantic v2 update.


dni:
depends on https://github.com/lnbits/lnbits/pull/3885

PR #3889, opened 2026-03-23
feat: do automatic bundling `make bundle` on the CI

dni:
removes the need to manually bundle and just let the ci do it. 

has a side effect of not running ci, ff rebased from dev and the last commit is the [skip ci] commit from alan. a workaround would be merging from dev instead of rebasing. the prs will be squashed anyways. or removing alan's pr by either removing alan pr, or force pushing if you have not pulled yet.


lnbits/static/js/components.js
```diff
@@ -1,4 +1,5 @@
 window.app.component(QrcodeVue)
+console.log('testing')
```
motorina0: remove this

PR #3894, opened 2026-03-24
[feat] reset the first install token

motorina0:
Reseting the first install token will allow the reseting of the super user from the UI


dni (approved):
tested

PR #3900, opened 2026-03-25
feat: optimize QR code value for bach32

motorina0:
Bech32 values should always be uppercase (including the prefix).

URLs and other non-bech32 values stay as they are


dni (approved):
😷 i dont like

PR #3905, opened 2026-03-26
EXPERIMENTAL: working pydantic v2 with latest fastapi

dni:
break all extension compatibility probably should be included in a major release, maybe `v2.0.0`.

pydantic v2 block updating to latest fastapi/starlette.


dni:
please, lets keep this PoC open for my sake or if some1 else attempts to upgrade to pydantic v2

PR #3908, opened 2026-03-28
fix: prevent incoming payments from having fees set

### LINKED ISSUE #3907: Internal transfers: incoming payment fee set to full amount (100% fee), zeroing balance
## Bug

Internal transfers between wallets on the same LNbits instance result in the **incoming** payment having `fee = -amount`, effectively zeroing out the received amount.

### Environment
- LNbits v1.5.3 (commit 7e0fadad)
- Backend: CLNRestWallet (Core Lightning)
- `lnbits_service_fee: 0.0`
- `lnbits_service_fee_ignore_internal: true`

### Steps to reproduce
1. Have two wallets (A and B) on the same LNbits instance
2. Wallet B has an LNURL-pay link
3. Send from wallet A to wallet B's Lightning Address (LNURL-pay)
4. Check wallet B's incoming payment

### Expected
Incoming payment has `fee: 0` (internal transfers should have no fees)

### Actual
Incoming payment has `fee: -<amount>` (fee equals full payment amount)

Example from DB:
```
-- Outgoing (sender side) - correct, fee=0
internal_80bea0c7...|wallet_a|-102000|0|success

-- Incoming (receiver side) - BUG: fee=-102000 (equals amount)
80bea0c7...|wallet_b|102000|-102000|success
```

### Root cause
In `update_invoice_callback()` (`services/payments.py` ~line 1091):

```python
status = await check_payment_status(payment, ...)
```

For incoming payments from internal transfers:
- `payment.is_internal` is `False` (checking_id doesn't have `internal_` prefix — that's only on the outgoing side)
- So `check_payment_status()` falls through to `funding_source.get_invoice_status()`
- CLN returns `fee_msat` equal to the payment amount
- This gets written to `payment.fee`

### Workaround
Incoming payments should never have fees — fees are a sender-side concept:

```python
# In update_invoice_callback():
if payment.is_in:
    payment.fee = 0
else:
```

Also need to fix existing corrupted payments:
```sql
UPDATE apipayments SET fee = 0 WHERE amount > 0 AND fee < 0 AND abs(fee) = amount;
```

ralyodio:
## Summary

Incoming payments should never have fees — fees are a sender-side concept. Without this guard, internal transfers between wallets on the same LNbits instance get `fee_msat` from the funding source equal to the full payment amount, effectively zeroing out the received balance.

## Root cause

In `update_invoice_callback()`, `check_payment_status()` is called for incoming payments. For internal transfers, the incoming payment's `checking_id` doesn't have the `internal_` prefix (only the outgoing side does), so `payment.is_internal` is `False`. This causes the code to fall through to `funding_source.get_invoice_status()`, which (at least with CLNRest) returns `fee_msat` equal to the payment amount. This value then gets written to `payment.fee`, zeroing out the balance.

## Fix

Incoming payments (`payment.is_in`) always get `fee = 0`. Only outgoing payments use `status.fee_msat`.

## Evidence

From a production instance (v1.5.3, CLNRestWallet backend):

```
-- Outgoing (sender side) - correct, fee=0
internal_80bea0c7...|wallet_a|-102000|0|success

-- Incoming (receiver side) - BUG: fee=-102000 (equals amount)  
80bea0c7...|wallet_b|102000|-102000|success
```

61 payments were affected across multiple wallets on our instance.

Fixes #3907

### DIFF
--- a/lnbits/core/services/payments.py
+++ b/lnbits/core/services/payments.py
@@ -1091,7 +1091,13 @@ async def update_invoice_callback(checking_id: str) -> Payment | None:
     status = await check_payment_status(
         payment, skip_internal_payment_notifications=True
     )
-    payment.fee = status.fee_msat or payment.fee
+    # Incoming payments should never have fees — fees are a sender-side concept.
+    # Without this guard, internal transfers (same instance) get fee_msat from the
+    # funding source equal to the full payment amount, zeroing out the balance.
+    if payment.is_in:
+        payment.fee = 0
+    else:
+        payment.fee = status.fee_msat or payment.fee
     # only overwrite preimage if status.preimage provides it
     payment.preimage = status.preimage or payment.preimage
     payment.status = PaymentState.SUCCESS


motorina0:
> Incoming payments should never have fees 

If the funding source is a LSP like Phoenix then the incoming payments will have fees too

PR #3914, opened 2026-04-01
feat: add extension profiles

talvasconcelos:
Adds filters for extensions by profile

Needs lnbits/lnbits-extensions#785

image


lnbits/core/models/extensions.py
```diff
@@ -58,6 +58,7 @@ class Manifest(BaseModel):
     featured: list[str] = []
```
dni: i think we can remove featured? it should be included in the profile.

lnbits/core/models/extensions.py
```diff
@@ -680,6 +681,7 @@ async def _get_installable_extensions(
 
                     meta = ext.meta or ExtensionMeta()
                     meta.featured = ext.id in manifest.featured
```
dni: remove here also

lnbits/templates/pages/extensions.vue
```diff
@@ -7,7 +7,25 @@
...
               <q-tab name="all" :label="$t('all')"></q-tab>
-              <q-tab name="featured" :label="$t('featured')"></q-tab>
+              <q-btn-dropdown auto-close stretch flat :label="$t('profiles')">
+                <q-list>
+                  <q-item clickable v-close-popup @click="tab = 'featured'">
+                    <q-item-section>
+                      <q-item-label v-text="$t('featured')"></q-item-label>
+                    </q-item-section>
+                  </q-item>
+                  <q-item clickable v-close-popup @click="tab = 'merchant'">
```
dni: don't hardcode profiles, you should be able to iterate them, right?

dni: then we can add new ones easily to manifest.

lnbits/core/models/extensions.py
talvasconcelos: i think this is used elsewhere, shouldn't we keep it for backwards compatibility?

talvasconcelos:
@dni can you find out why this is failing the tests?

`list[str] = []` seems to be the issue?

motorina0:
Multiple CI jobs fail.

motorina0:
- most CI jobs fail
- please add screenshots

lnbits/static/js/pages/extensions.js
```diff
@@ -106,13 +107,17 @@ window.PageExtensions = {
         }
       }
 
+      const isProfileTab = !['installed', 'all'].includes(tab)
```
motorina0: why not check the tab name?

lnbits/static/js/pages/extensions.js
```diff
@@ -832,6 +837,9 @@ window.PageExtensions = {
     async fetchAllExtensions() {
       try {
         const {data} = await LNbits.api.request('GET', `/api/v1/extension/all`)
+        data.forEach(ext => {
```
motorina0, line 841: isn't this done on server side already?

talvasconcelos: Server side, we attach profiles to each extension, when aplicable. the loop is creating the profile names for the dropdown button, so if we add another profile in the future, it's rendered dynamicaly and there's no hardcode of the profiles.

dni (approved):

dni (approved):

motorina0 (approved):

PR #3916, opened 2026-04-02
fix(lnd): enable MPP via max_parts=16 and time_pref in SendPaymentV2

emtll:
## Problem

When paying a BOLT11 invoice via `SendPaymentV2`, `max_parts` is not set (defaults to `0`), which causes LND to attempt a **single-path payment only**. This results in `FAIL_REASON_NO_ROUTE` for payments where no single channel path has sufficient liquidity, even when multiple smaller paths exist across the network.

## Fix

- Set `max_parts=16` to enable MPP (Multi-Path Payment), allowing LND to split the payment across up to 16 independent paths
- Set `time_pref=0.9` to prefer more reliable routes over cheapest routes

## Impact

Invoices with `basic_mpp: supported` in their feature bits (most modern wallets including Phoenix, Blink, IBEX/Wama) can now be paid via MPP, significantly improving payment success rates on nodes with fragmented outbound liquidity.


motorina0:
Please remove `ghcr.yml`

dni (approved):

motorina0 (approved):

PR #3925, opened 2026-04-12
fix: non-blocking relay publishes in send_nostr_dm

### LINKED ISSUE #3924: [BUG] send_nostr_dm blocks asyncio event loop on unreachable relays
> Take the proposed resolution with a pinch of salt — drafted quickly with AI assistance. The diagnosis (sync websocket call blocking the event loop) is verified against live logs, but the exact fix shape deserves maintainer review.

## Summary

`send_nostr_dm()` in `lnbits/core/services/nostr.py` uses the **synchronous** `websocket.create_connection()` inside an async function and iterates relays sequentially. One unreachable or slow relay blocks the asyncio event loop for up to ~30 seconds per attempt, making NWC Provider and other WebSocket-based extensions unresponsive during that time.

This is a **separate event-loop-starvation bug** from #3917 / #3918 (IN_FLIGHT payment polling). After applying the #3918 fix, this became the next visible bottleneck.

## Offending code

`lnbits/core/services/nostr.py`, line ~30:

```python
ws_connections: list[WebSocket] = []
for relay in relays:
    try:
        ws = create_connection(relay, timeout=2)   # sync call in async fn
        ws.send(notification)
        ws_connections.append(ws)
    except Exception as e:
        logger.warning(f"Error sending notification to relay {relay}: {e}")
await asyncio.sleep(1)
```

Issues:
1. `create_connection()` is synchronous — it blocks the event loop on TCP connect + TLS handshake
2. The `timeout=2` is not always honored (we observe 30s blocks in production against `wss://relay.snort.social` and `wss://relay.nostr.band`)
3. Relays are tried **sequentially** — 3 dead relays = 3 × blocking time added up
4. `ws.send()` is also sync

With 3 dead relays in the user's configured list, a single `send_nostr_dm()` call can block the event loop for 60–90 seconds.

## Symptoms

- NWC `list_transactions`, `get_balance`, `pay_invoice` intermittently time out
- `docker logs` repeatedly shows:
  ```
  WARNING | Error sending notification to relay wss://nostr.wine: Handshake status 403 Forbidden
  WARNING | Error sending notification to relay wss://relay.snort.social: Connection timed out
  WARNING | Error sending notification to relay wss://relay.nostr.band: timed out
  ```
- Web UI works (HTTP served separately), masking the issue
- Only recovery from a pile-up is `docker restart`

## Proposed fix

Two options (either or both):

**A. Offload sync calls via `asyncio.to_thread`** and run concurrently:
```python
async def _publish(relay: str):
    try:
        ws = await asyncio.wait_for(
            asyncio.to_thread(create_connection, relay, timeout=2),
            timeout=3,
        )
        await asyncio.to_thread(ws.send, notification)
        return ws
    except Exception as e:
        logger.warning(f"Error sending notification to relay {relay}: {e}")
        return None

results = await asyncio.gather(*(_publish(r) for r in relays))
ws_connections = [ws for ws in results if ws]
```

**B. Switch to an async websocket library** (`websockets` or `aiohttp`) — larger change but removes the sync/async footgun entirely.

Option A is minimally invasive and fixes the starvation. Happy to open a PR with that approach.

## Related

- #3917 — IN_FLIGHT payment polling starves asyncio event loop (sibling bug)
- #3918 — draft fix for #3917 (currently in production testing)
- lnbits/nostrclient#40 — nostrclient temporarily unresponsive (same class of symptom, different trigger)
- lnbits/nostrclient#69 — initially misfiled there; moved here after tracing to `lnbits/core/services/nostr.py`

## Environment

- LNbits v1.5.3
- LNDRest backend
- Python 3.x, asyncio

BenGWeeks:
> Take the proposed fix with a pinch of salt — drafted quickly with AI assistance. The diagnosis (sync websocket call blocking the event loop) is verified against live production logs, but the exact fix shape deserves maintainer review.

## Summary

Fixes #3924

`send_nostr_dm()` in `lnbits/core/services/nostr.py` uses `websocket.create_connection()` and `ws.send()` — both synchronous — inside an `async def`. With one or more unreachable relays, each call blocks the asyncio event loop for up to ~30 seconds (TCP/TLS handshake, DNS, etc.), and relays are tried sequentially, so 3 dead relays = 60–90s of event-loop starvation per call.

This starves NWC Provider, Nostrclient, and other WebSocket-based extensions — users see `pay_invoice`, `get_balance`, and `list_transactions` time out.

Solution:
Run notifications in the backgrodun

### DIFF
--- a/lnbits/core/services/nostr.py
+++ b/lnbits/core/services/nostr.py
@@ -29,18 +29,32 @@ async def send_nostr_dm(
     dm_event.sign(private_key_hex=from_private_key_hex)
     notification = dm_event.to_message()
 
-    ws_connections: list[WebSocket] = []
-    for relay in relays:
+    async def _publish(relay: str) -> WebSocket | None:
+        # websocket-client's create_connection and send are synchronous and
+        # will block the asyncio event loop on slow TCP/TLS handshakes,
+        # starving NWC and other WebSocket-based extensions. Offload to a
+        # worker thread and cap total time with wait_for so a single dead
+        # relay cannot stall the whole publish batch.
         try:
-            ws = create_connection(relay, timeout=2)
-            ws.send(notification)
-            ws_connections.append(ws)
+            ws = await asyncio.wait_for(
+                asyncio.to_thread(create_connection, relay, timeout=2),
+                timeout=3,
+            )
+            await asyncio.wait_for(
+                asyncio.to_thread(ws.send, notification),
+                timeout=3,
+            )
+            return ws
         except Exception as e:
             logger.warning(f"Error sending notification to relay {relay}: {e}")
+            return None
+
+    results = await asyncio.gather(*(_publish(r) for r in relays))
+    ws_connections: list[WebSocket] = [ws for ws in results if ws]
     await asyncio.sleep(1)
     for ws in ws_connections:
         try:
-            ws.close()
+            await asyncio.to_thread(ws.close)
         except Exception as e:
             logger.debug(f"Failed to close websocket connection: {e}")
 
--- a/lnbits/core/services/nostr.py
+++ b/lnbits/core/services/nostr.py
@@ -29,32 +29,18 @@ async def send_nostr_dm(
     dm_event.sign(private_key_hex=from_private_key_hex)
     notification = dm_event.to_message()
 
-    async def _publish(relay: str) -> WebSocket | None:
-        # websocket-client's create_connection and send are synchronous and
-        # will block the asyncio event loop on slow TCP/TLS handshakes,
-        # starving NWC and other WebSocket-based extensions. Offload to a
-        # worker thread and cap total time with wait_for so a single dead
-        # relay cannot stall the whole publish batch.
+    ws_connections: list[WebSocket] = []
+    for relay in relays:
         try:
-            ws = await asyncio.wait_for(
-                asyncio.to_thread(create_connection, relay, timeout=2),
-                timeout=3,
-            )
-            await asyncio.wait_for(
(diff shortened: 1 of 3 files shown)


lnbits/core/services/nostr.py
```diff
@@ -29,18 +29,32 @@ async def send_nostr_dm(
...
             logger.warning(f"Error sending notification to relay {relay}: {e}")
+            return None
+
+    results = await asyncio.gather(*(_publish(r) for r in relays))
+    ws_connections: list[WebSocket] = [ws for ws in results if ws]
     await asyncio.sleep(1)
     for ws in ws_connections:
         try:
-            ws.close()
+            await asyncio.to_thread(ws.close)
```
motorina0, line 57: the close operation is usually very fast
did you experience high times for it?

lnbits/core/services/nostr.py
```diff
@@ -29,18 +29,32 @@ async def send_nostr_dm(
...
+        # websocket-client's create_connection and send are synchronous and
+        # will block the asyncio event loop on slow TCP/TLS handshakes,
+        # starving NWC and other WebSocket-based extensions. Offload to a
+        # worker thread and cap total time with wait_for so a single dead
+        # relay cannot stall the whole publish batch.
         try:
-            ws = create_connection(relay, timeout=2)
-            ws.send(notification)
-            ws_connections.append(ws)
+            ws = await asyncio.wait_for(
```
motorina0, line 39: why is there a need for `asyncio.wait_for` inside the function that will be used in `asyncio.gather`?

motorina0 (approved):

dni (approved):

PR #3927, opened 2026-04-12
fix: first_install for local

arcbtc:
I get the logic in the initial change, but default behavior should continue as it was

### DIFF
--- a/lnbits/settings.py
+++ b/lnbits/settings.py
@@ -1007,8 +1007,9 @@ class EnvSettings(LNbitsSettings):
     debug_database: bool = Field(default=False)
     bundle_assets: bool = Field(default=True)
     # When enabled, auth cookies require HTTPS and SSO will reject insecure HTTP.
-    # Set to false for local/dev environments that run without TLS.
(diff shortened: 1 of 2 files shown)


motorina0 (approved):

lnbits/settings.py
```diff
@@ -1007,8 +1007,9 @@ class EnvSettings(LNbitsSettings):
     debug_database: bool = Field(default=False)
     bundle_assets: bool = Field(default=True)
     # When enabled, auth cookies require HTTPS and SSO will reject insecure HTTP.
-    # Set to false for local/dev environments that run without TLS.
-    auth_https_only: bool = Field(default=True)
+    # Keep disabled by default so local HTTP installs continue to work unless
+    # operators explicitly opt into HTTPS-only auth cookies.
+    auth_https_only: bool = Field(default=False)
```
dni, line 1012: this will use insecure cookies when running behind a proxy basically on all production instances. we need to add it to the docs on how to  run behind a proxy and what to configure. in my opionen the sane default is `True` and if you want to develop locally, you set it to false.

PR #3929, opened 2026-04-12
feat:  adds an upload for assets on backgroundImage in account and site customisation

arcbtc:
image
image


motorina0 (approved):

dni (approved):
lgtm, changed title its a `feat` not a `fix`

idea, possibly a component. `form-image-selector`. could also have thumbnails and a image selector showing a dialog with already uploaded assets and from there you could also upload and then select.

dni:
you dont need to commit the bundle anymore, thats done via CI

PR #3931, opened 2026-04-14
feat: Add phoenixd mnemonic display

talvasconcelos:
Add a display of the phoenixd funding source mnemonic (seed phrase) for backup purposes! 

It needs the correct `data-dir` path of phoenixd to read the `seed.dat` file!

If the `data-dir` is not reacheable by LNbits, the mnemonic will not be displayed. This doesn't mean that phoenixd is not running, just that LNbits can't reach the file.

For better results, you can launch phoenixd with the `--data-dir` flag and use `lnbits/data/` folder as target. 

image


lnbits/templates/components/admin/funding_sources.vue
```diff
@@ -36,6 +36,14 @@
           fund === formData.lnbits_backend_wallet_class
         "
       >
+        <!-- <div class="row">
```
motorina0: dead code

lnbits/settings.py
```diff
@@ -324,7 +325,7 @@ class AssetSettings(LNbitsSettings):
             "heif",
             "heics",
             "text/plain",
-            "text/json" "text/xml",
+            "text/jsontext/xml",
```
dni: this is wrong,  was wrong before, needs an `,`

lnbits/settings.py
```diff
@@ -588,6 +589,38 @@ class ZBDFundingSource(LNbitsSettings):
 class PhoenixdFundingSource(LNbitsSettings):
     phoenixd_api_endpoint: str | None = Field(default="http://localhost:9740/")
     phoenixd_api_password: str | None = Field(default=None)
+    phoenixd_data_dir: str | None = Field(default="data/")
+    phoenixd_mnemonic: str | None = Field(default=None)
+
+    @validator("phoenixd_mnemonic", always=True)
+    @classmethod
```
motorina0: why `classmethod` (all other functions are not like this)?

lnbits/settings.py
```diff
@@ -588,6 +589,38 @@ class ZBDFundingSource(LNbitsSettings):
...
+    def validate_phoenixd_mnemonic(cls, val, values):
```
motorina0: this is not a validate function, it fetches data

lnbits/settings.py
motorina0: why is this?

lnbits/settings.py
motorina0: `values` means nothing, why do you pass a full object?

talvasconcelos: moved the check and get mnemonic to the wallet file, like sparkl2 does!

lnbits/settings.py
motorina0: @talvasconcelos how is this resolved?

lnbits/wallets/phoenixd.py
```diff
@@ -319,3 +321,31 @@ async def paid_invoices_stream(self) -> AsyncGenerator[str, None]:
...
+            return
+
+        if not mnemonic_is_valid(mnemonic):
+            logger.warning(
+                f"Phoenixd seed file '{seed_path}' does not contain a valid "
+                "BIP39 mnemonic."
+            )
+            return
+
+        settings.phoenixd_mnemonic = mnemonic
```
motorina0, line 355: this assignment only will not persist the value

talvasconcelos: i see... it would stay in memory only and not save to settings. fixed

lnbits/settings.py
talvasconcelos: wtf!! i fixed it! maybe some undo screwed me, sorry

motorina0 (approved):

lnbits/wallets/phoenixd.py
```diff
@@ -319,3 +323,46 @@ async def paid_invoices_stream(self) -> AsyncGenerator[str, None]:
                     "retrying in 5 seconds"
                 )
                 await asyncio.sleep(5)
+
+    def _load_mnemonic_from_seed_file(self):
+        data_dir = settings.phoenixd_data_dir or "data/"
```
dni: defaults should not be defined here. else it makes not much sense to have the settings as `optional + default`. you either can set the setting it to `str` and have a default with `data/` so not optional. or you have it `str | None` and let people set it to `None` and not override it in the fundingsource.

talvasconcelos: makes sense... was thinking about  the saas and how we discussed it yesterday!

dni (approved):

PR #3944, opened 2026-04-28
feat: Add fork migrations namespace

PatMulligan:
## Summary

- Adds an empty lnbits/core/migrations_fork.py and wires it into migrate_databases() under a separate core_fork key in the dbversions table.
- Lets downstream forks add custom schema changes (m001_*, m002_*, …) without conflicting with upstream's core migration numbering on every git pull upstream.
- Upstream behavior unchanged: the file is empty, so the new run_migration call is a no-op.

## Motivation

Forks that need extra columns/tables today have to patch lnbits/core/migrations.py directly, which produces conflicts on every upstream sync. This gives forks a stable, conflict-free migrations :hugs:

### DIFF
--- a/lnbits/core/helpers.py
+++ b/lnbits/core/helpers.py
@@ -7,6 +7,7 @@ from uuid import UUID
 from loguru import logger
 
 from lnbits.core import migrations as core_migrations
+from lnbits.core import migrations_fork as core_migrations_fork
 from lnbits.core.crud import (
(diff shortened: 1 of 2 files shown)


motorina0:
Concept NACK.
Forks should git rebase upstream.

PatMulligan:
hehe, i figured it might not be totally legit, but figured I'd PR just in case :)

PR #3946, opened 2026-04-29
Fix: “Show all” table pagination

### LINKED ISSUE #3945: [BUG] Show all" in Users & Tx Table does not display all records
**Description:**
In the Core → Users table, the "Records per page" option is not working as expected.

When selecting **"Show all"**, the table still defaults to displaying only 10 entries instead of all available records.

**Current behavior:**

* "Records per page" set to "All"
* Table still shows: `1–10 of 72`

**Expected behavior:**

* Selecting "All" should display all records without pagination

**Additional notes:**
I also noticed the same behavior in the wallet transaction tables, so this may affect multiple tables using the same pagination component.

Image

**Version: 1.5.3.**

talvasconcelos:
This fixes table pagination when selecting “Show all” from the records-per-page dropdown.

Quasar sends `rowsPerPage: 0` to mean “all rows”, but LNbits treated `limit=0` as a missing limit and replaced it with the default of 10. As a result, tables like Users and wallet transactions still showed only 10 records.

The fix preserves `limit=0` as “no SQL limit” and keeps the existing default of 10 only when no limit is provided.

Also added a regression test for `limit=0` pagination.

Closes #3945

### DIFF
--- a/lnbits/db.py
+++ b/lnbits/db.py
@@ -560,7 +560,9 @@ class Filters(BaseModel, Generic[TFilterModel]):
 
     def pagination(self) -> str:
         stmt = ""
-        self.limit = self.limit or 10
+        if self.limit == 0:
+            return stmt
+        self.limit = 10 if self.limit is None else self.limit
         stmt += f"LIMIT {min(1000, self.limit)} "
         if self.offset:
             stmt += f"OFFSET {self.offset}"
--- a/lnbits/db.py
+++ b/lnbits/db.py
@@ -35,7 +35,7 @@ if settings.lnbits_database_url:
     else:
         if not database_uri.startswith("postgres://"):
             raise ValueError(
-                "Please use the 'postgres://...' " "format for the database URL."
+                "Please use the 'postgres://...' format for the database URL."
             )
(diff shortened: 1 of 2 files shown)


DoktorShift (approved):

motorina0:
NACK
what if there are 1 million payments?

dni:
i am with vlad here, we cannot show all, that will in some cases most likely crash the server. i would go with removing the show all. from q-table

talvasconcelos:
Can we cap it to 200 for example?

I see the concern! maybe "All" caps it to 200/500 ?

motorina0:
> Can we cap it to 200 for example?
> 
> I see the concern! maybe "All" caps it to 200/500 ?

```py
        stmt += f"LIMIT {min(1000, self.limit)} "
```

see: https://github.com/lnbits/lnbits/pull/3946/changes#diff-5855a9aec8f89dbcb613cd599d551e090dc55f68b23b20b7c2473261db814dadR566

motorina0 (approved):

motorina0 (approved):

motorina0 (approved):

PR #3948, opened 2026-04-30
Fix: Restore "instawallet" on LNURLw claim 

talvasconcelos:
Fixes a bug that wasn't showing the claim LNURLw claim button for users with no LN wallet!

This won't show if the LNbits instance doesn't allow register of new users, or register users with `user-id-only`!

image

The `lnbits-qrcode-lnurl` component changed ALL lnurl qr's to show `lightning:LNURL1......` which broke the instawallet feature (scanning the withdraw link with a regular scanner would point the user to LNbits and show a claim button. create an account -> redeem the LNURLw)

The solution is to use the fallback scheme, when dealing with LNURLw (as was used before, and it's used on the printed LNURLw QR codes). This is a **DELIBERATE** decision and intended for **ALL LNURLw created on LNbits**!

While investigating the issue, I also found that there was a bug on the computed `showClaimLnurl` that would always be false. The 2 changes can't be dissociated and must be done together in order for the intended purpose of restoring the "instawallet" feature! 

If it's absolutely necessary, I can create 2 separate PR's for this, but in my opinion it doesn't make sense.

As is, and to reduce the scope of the change, it will only affect the QR display for LNURLw, when/if the `prefix` prop 'lnurlw' is passed to the `lnbits-qrcode-lnurl`, as the withdraw extension already does. All other QR codes remain the same.

To reproduce the issue:

- create an LNURLw on the Withdraw Links extension
- scan with a regular scanner, with no LN wallet
- instead of opening LNbits and allow a new user to claim sats, you just get the text decoded ('lightning: LNURL1...')

Even if a user can construct the URL (assuming an LNURLw on demo.lnbits.com) like `https://demo.lnbits.com/?lightning=LNURL1....` it won't work because of the error in the `showClaimLnurl`

The goal of this PR is to restore the LNbits Instawallet feature. 

Why it's an important feature?
Regular people can scan a QR code (a faucet, voucher, giveaway) and instantly get an LNbits account and automatically redeem an LNURLw into his new LN wallet!


lnbits/static/js/pages/home.js
```diff
@@ -21,7 +21,7 @@ window.PageHome = {
       return (
         this.lnurl !== '' &&
         this.g.settings.allowRegister &&
-        'user-id-only' in this.g.settings.authMethods
+        this.g.settings.authMethods.includes('user-id-only')
```
motorina0, line 24: why change this?

lnbits/static/js/components/lnbits-qrcode-lnurl.js
```diff
@@ -21,7 +21,7 @@ window.app.component('lnbits-qrcode-lnurl', {
       if (this.tab == 'bech32') {
         const bytes = new TextEncoder().encode(this.url)
         const bech32 = NostrTools.nip19.encodeBytes('lnurl', bytes)
-        this.lnurl = `lightning:${bech32.toUpperCase()}`
+        this.lnurl = `${new URL(this.url).origin}/?lightning=${bech32.toUpperCase()}`
```
motorina0: will this affect LNURLp?

lnbits/static/js/pages/home.js
talvasconcelos: `g.settings.authMethods` is an array, the `in` would always return `false`!

lnbits/static/js/components/lnbits-qrcode-lnurl.js
talvasconcelos: I tested scanning the resulting QR! works as expected.

https://github.com/lnurl/luds/blob/luds/01.md#fallback-scheme

lnbits/static/js/pages/home.js
dni: that was a bug.

lnbits/static/js/components/lnbits-qrcode-lnurl.js
dni: seems scary :D but i think its correct.

dni (approved):

motorina0 (approved):

lnbits/static/js/components/lnbits-qrcode-lnurl.js
motorina0: this PR modifies a reusable component
the PR name and description must be adjusted

lnbits/static/js/components/lnbits-qrcode-lnurl.js
```diff
@@ -21,7 +21,10 @@ window.app.component('lnbits-qrcode-lnurl', {
       if (this.tab == 'bech32') {
         const bytes = new TextEncoder().encode(this.url)
         const bech32 = NostrTools.nip19.encodeBytes('lnurl', bytes)
-        this.lnurl = `lightning:${bech32.toUpperCase()}`
+        this.lnurl =
```
motorina0, line 24: please make another PR for this

talvasconcelos: Please check #3960 and #3961

PR #3949, opened 2026-05-01
feat: admin_only super_user_only flags for extension configs/manifests

arcbtc:
admin_only or super_user_only flags can be used in the config or manifest to restrict access to extensions. Includes api guard for authenticated routes, just in case is_admin auth method is omitted.

```
{
  "repos": [
    {
      "id": "tunnel_me_out",
      "organisation": "lnbits",
      "repository": "tunnel_me_out",
      "super_user_only": true
    }
  ]
}
```

image


dni:
how does this fit into the  setting.  `lnbits_admin_extensions: list[str] = Field(default=[])`, would your feature be an alternative to this?

dni (commented):
wouldnt it be easier to just add the superuser/admin decorator in the extensions views.py generic routes? not sure if we want to add this complexity into the extension config... when then endpoints inside the extension need to be checked anyways, else you can access them

PR #3950, opened 2026-05-01
feat: translation support for extensions

arcbtc:
Keeps same framework.
Example extension here https://github.com/lnbits/tunnel_me_out/pull/2
Extension stub would need updating.
image


dni:
let me give it a spin, i think it should be dynamically loaded like our dynamic components of the newer extensions.

dni (changes requested):
should be dynamically loaded not rendered inside the template

motorina0:
Superseded by https://github.com/lnbits/lnbits/pull/4018

PR #3953, opened 2026-05-02
fix: populate unit selector when opening send/pay dialog

AxelHamburch:
## Problem

When opening the **Send** dialog and reading a lightning address or LNURL-pay
request, the **Unit** dropdown only ever showed `sat`. Selecting a fiat
currency (EUR, USD, etc.) was only possible after having opened the Receive
dialog first.

The root cause: the `<q-select>` in the LNURL-pay view binds to
`receive.units`, but that array was only populated inside `showReceiveDialog()`.
When the Send dialog was opened directly, the list remained empty.

## Fix

Populate `receive.units` inside `lnurlScan()` at the point where a
`payRequest` response is received — the earliest and only moment the unit
selector actually becomes visible.

This works regardless of how the dialog was opened (`showParseDialog()`,
`handleSendLnurl()`, QR scan, or URL parameter).

## Changed files

- `lnbits/static/js/pages/wallet.js` — 6 lines added in `lnurlScan()`

No backend changes required. The backend (`/api/v1/payments/lnurl`) already
handles the `unit` field and converts fiat amounts to millisatoshi via
`fiat_amount_as_satoshis()`.

## How to test

1. Open a wallet → **Send** → paste a lightning address → READ
2. The **Unit** dropdown now shows `sat`, `EUR`, `USD`, …
   without having opened the Receive dialog first
3. Select a fiat currency, enter an amount, confirm — payment goes through
   with correct conversion


talvasconcelos (approved):
tested

motorina0 (approved):

lnbits/static/js/pages/wallet.js
```diff
@@ -223,6 +223,12 @@ window.PageWallet = {
           if (data.tag === 'payRequest') {
             this.parse.lnurlpay = Object.freeze(data)
             this.parse.data.amount = data.minSendable / 1000
+            this.receive.units = [
+              'sat',
```
dni:
```suggestion
              'sats',
```

i think we use sats in other places.

PR #3955, opened 2026-05-06
test: add pyinstrument profiler

dni:
just add `?profiler=true` to the url to get a profiler html output

- add EnvSettings `PROFILER`
- add profiler middleware
Screenshot_2026-05-06_08-01-46


talvasconcelos (approved):
tested

lnbits/settings.py
```diff
@@ -1009,6 +1009,7 @@ class EnvSettings(LNbitsSettings):
     debug: bool = Field(default=False)
     debug_database: bool = Field(default=False)
     bundle_assets: bool = Field(default=True)
+    profiler: bool = Field(default=False)
```
motorina0, line 1012: nitpick: `enable_profiler`

so it is not confused with `profiler name` or other type of property

motorina0 (approved):
LGTM

see comment

lnbits/settings.py
dni: i'll say we keep it like that, we don't have enable_debug or enable_bundle_assets either.

PR #3961, opened 2026-05-07
Fix: Allow the LNURL spec fallback scheme for LNURLw

talvasconcelos:
The `lnbits-qrcode-lnurl` component changed ALL lnurl qr's to show `lightning:LNURL1......` which broke the instawallet feature (scanning the withdraw link with a regular scanner would point the user to LNbits and show a claim button. create an account -> redeem the LNURLw)

The solution is to use the fallback scheme, when dealing with LNURLw (as was used before, and it's used on the printed LNURLw QR codes). This is a DELIBERATE decision and intended for ALL LNURLw created on LNbits!

As is, and to reduce the scope of the change, it will only affect the QR display for LNURLw, when/if the `prefix` prop 'lnurlw' is passed to the `lnbits-qrcode-lnurl`, as the withdraw extension already does. All other QR codes remain the same.

The goal of this PR is to restore the LNbits Instawallet feature. 

Why it's an important feature?
Regular people can scan a QR code (a faucet, voucher, giveaway) and instantly get an LNbits account and automatically redeem an LNURLw into his new LN wallet!

It needs #3960 for it to work as expected.
Splitted from #3948


motorina0:
It works.
I'm not convinced its the best default though.

motorina0:
Better approach:
 - add a `href` property to this component
 - if that property is set then show the `bech32` as an url
 - update the `withdraw` extension to pass the `href=`

motorina0:
> Better approach:
> 
>     * add a `href` property to this component
> 
>     * if that property is set then show the `bech32` as an url
> 
>     * update the `withdraw` extension to pass the `href=`

current implementation does not match above specs

motorina0 (approved):

dni (approved):

PR #3637, opened 2025-12-08
feat: Adds paypal as a fiat choice

arcbtc:
image
image
image

PR #3184, opened 2025-05-30
[feat] Add stripe payments

motorina0:
## Allow stripe payments
This PR allows the admin to configure `stripe` as a fiat payment provider.
This feature allows all (or a selected subset) of users to **RECEIVE** payments via stripe.

https://github.com/user-attachments/assets/6eb13a9a-89ec-4550-a41b-e94236d61601

### User Flow
- When the Fiat Provider is configured (`Stripe` in this case) the end user is allowed to select how it wants to receive the funds (lightning or stripe):
   - image

- If the user selects `Stripe` then a Stripe Payment Request is generated instead of a LN bolt11.
- The buyer/payer can scan the QR code or click on it and it will be redirected to a Stripe checkout page like this one:
   - image

- After the Stripe payment is completed, Stripe will send a notification to the LNbits instance and the LNbits payment will be marked as successful.
   -  image

For a user that only has fiat funds, but wants to **PAY** a bolt11 Lightning Invoice (received from someone) then all it has to do is:
 - create a payment request in his LNbits wallet for the desired amount with Stripe as Fiat Payment Provider
 - pay the Stripe payment request -> the LNbits wallet is toped-up with that amount
 - pay the external Lightning Invoice (his wallet now has the funds)

### Admin Configs
Admins now have a `Fiat Providers` tab where they can configure multiple fiat providers (for now only Stripe is supported):
image

These configs should be self explanatory (they have hints and descriptions).

Todos:
   - [x] check stripe callback signature

PR #3962, opened 2026-05-11
feat: square integration

motorina0:
Add a new fiat provider. Previous ones: https://github.com/lnbits/lnbits/pull/3184 , https://github.com/lnbits/lnbits/pull/3637

This PR also adds `external_id` field for payments (and filtering on it).

Sandox URL: https://connect.squareupsandbox.com
Add a test app here: https://developer.squareup.com/console/en/apps
Configure test items: https://app.squareupsandbox.com/dashboard/

### Configure Square fiat provider
 - Sandbox Access Token
 - Location ID
 - Webhook Signature Key

image

### Request fiat payment with Square:
image

### Pay with square (sandbox page is different than prod):
image

### Check your payments here: https://app.squareupsandbox.com/dashboard/sales/transactions
image


lnbits/core/crud/payments.py
```diff
@@ -107,6 +107,35 @@ async def get_latest_payments_by_extension(
     )
 
 
+async def get_latest_payment_by_extra_key_value(
```
dni: i am really not a fan of that function. its not a blocker. but if we could get around using that, i would prefer that. just seems wrong making a LIKE query over key and value of some json.

lnbits/core/views/callback_api.py
```diff
@@ -280,3 +303,196 @@ def _deserialize_paypal_metadata(custom_id: str) -> FiatSubscriptionPaymentOptio
...
+        return FiatSubscriptionPaymentOptions(
+            wallet_id=wallet_id,
+            tag=tag,
+            subscription_request_id=subscription_request_id,
+            extra=extra,
+            memo=memo,
+        )
+    except (json.JSONDecodeError, IndexError, TypeError) as e:
+        logger.debug(f"Failed to deserialize Square metadata: {e}")
+        return FiatSubscriptionPaymentOptions()
```
dni, line 713: why is there so much business logic specific to square in `lnbits/core/views/callback_api.py`? i dont think it belongs here parse square stuff should be in the square Fiatprovider.

lnbits/core/crud/payments.py
motorina0: removed

talvasconcelos:
Brushed through the code... but I need to take some time to make a square account to test!

arcbtc (approved):

PR #3969, opened 2026-05-15
feat: hide burger bg toggle

arcbtc:
image
image


talvasconcelos (approved):
works as expected! i like it...

i wouldn't call it "burger background", effectively is the drawer/left drawer background

motorina0:
I don't see/understand the change

PR #3974, opened 2026-05-18
feat: allow extra appending

### LINKED ISSUE #3237: [Feature request] Ex-post transaction metadata endpoint
Sometimes it could be very useful to add extra_json metadata after a transaction was completed (e.g. compliance tracking or linking a splitpayments mother to the daughters or simpy adding additional notes)

Either a dedicated field or an appendix in the extra_json. Could be implemented so that the other fields are still immutable. Ideally, the field would have the option to add data multiple times, but only append new data, not overwrite existing metadata.

-

talvasconcelos:
Allow for post payment updates to the extra dict

Closes #3237


lnbits/core/crud/payments.py
```diff
@@ -20,8 +21,22 @@
 )
 
 
-def update_payment_extra():
-    pass
+async def update_payment_extra(
```
dni: this is equivalent to the `update_payment`. please remove again and use update_payment

lnbits/core/views/payment_api.py
```diff
@@ -296,6 +298,42 @@ async def api_update_payment_labels(
...
+        data.payment_hash, wallet_id=key_type.wallet.id
+    )
+    if payment is None:
+        raise HTTPException(HTTPStatus.NOT_FOUND, "Payment does not exist.")
+    if not payment.success:
+        raise HTTPException(
+            HTTPStatus.BAD_REQUEST, "Payment extra can only be updated after success."
+        )
+    if not data.extra:
+        raise HTTPException(
```
dni: why only allowing updating extra when it exists? i think that is a wrong check. create it if it did not exist before.

lnbits/core/views/payment_api.py
```diff
@@ -296,6 +298,42 @@ async def api_update_payment_labels(
...
+
+    duplicate_keys = sorted(set(payment.extra).intersection(data.extra))
+    if duplicate_keys:
+        raise HTTPException(
+            HTTPStatus.BAD_REQUEST,
+            f"Extra keys already exist: {', '.join(duplicate_keys)}.",
+        )
+
+    payment.extra.update(data.extra)
+    await update_payment_extra(payment)
```
dni: use `update_payment`

lnbits/core/views/payment_api.py
```diff
@@ -296,6 +298,42 @@ async def api_update_payment_labels(
...
 
+@payment_router.patch(
+    "/extra",
+    name="Update payment extra",
+    description="Append new extra metadata to a payment.",
+    response_model=Payment,
+)
+async def api_update_payment_extra(
+    data: UpdatePaymentExtra,
+    key_type: WalletTypeInfo = Depends(require_admin_key),
```
dni, line 309: sidenote: imho, we should think about not using wallet admin key here and keep the scope of those keys strictly to simple wallet functions. maybe in this case updating the payment it is ok.

lnbits/core/views/payment_api.py
```diff
@@ -296,6 +297,40 @@ async def api_update_payment_labels(
...
+    payment = await get_standalone_payment(
+        data.payment_hash, wallet_id=key_type.wallet.id
+    )
+    if payment is None:
+        raise HTTPException(HTTPStatus.NOT_FOUND, "Payment does not exist.")
+    if not payment.success:
+        raise HTTPException(
+            HTTPStatus.BAD_REQUEST, "Payment extra can only be updated after success."
+        )
+    if not data.extra:
```
dni: it shouldnt be optional. i think you can remove that?

dni (approved):

motorina0 (approved):

PR #3979, opened 2026-05-19
fix: OIDC SSO login issues.  closes #3970

dni:
> Some OIDC providers (e.g. PocketID) return the literal string "null" as the OAuth state parameter when the original state was None/empty

closes #3970


arcbtc (approved):

boyroywax:
greatly appreciated!

PR #3980, opened 2026-05-19
Fix: skip internal invoice staus check

### LINKED ISSUE #3816: [BUG] Internal payments zero out receiver wallet balance (CLNRestWallet)

### LINKED ISSUE #3869: Internal transfers set fee=-amount on receiver's payment record, zeroing balance
### Bug

Internal (same-instance) wallet-to-wallet transfers set `fee = -amount` on the **receiver's** payment record in `apipayments`. The `balances` view calculates `SUM(amount - ABS(fee))`, so received funds show as 0.

### Reproduce

1. Two wallets (A, B) on same LNbits instance (default config, no service fee)
2. Create invoice on wallet B: `POST /api/v1/payments` with `{"out": false, "amount": 100, "memo": "test"}`
3. Pay from wallet A: `POST /api/v1/payments` with `{"out": true, "bolt11": "..."}`
4. Check wallet B balance via API → **shows 0**
5. Check DB:

```sql
SELECT amount, fee, status FROM apipayments
WHERE wallet_id = '<wallet_B_id>' AND memo = 'test';
```

### Expected

| amount | fee | status  |
|--------|-----|---------|
| 100000 | 0   | success |

### Actual

| amount | fee     | status  |
|--------|---------|---------|
| 100000 | -100000 | success |

Balance view: `amount - ABS(fee) = 100000 - 100000 = 0`

### Impact

**Critical** — all internal transfers result in receivers showing 0 balance. Affects any deployment using wallet-to-wallet transfers on the same instance.

### Versions

- Confirmed on **v1.4.1**
- **v1.5.0** has identical code path in `_pay_internal_invoice()` (`lnbits/core/services/payments.py`)

### Workaround

```sql
UPDATE apipayments SET fee = 0
WHERE amount > 0 AND fee < 0 AND status = 'success';
```

### Environment

- Backend: CLNRest (Core Lightning)
- DB: SQLite
- OS: Ubuntu 24.04

### LINKED ISSUE #3907: Internal transfers: incoming payment fee set to full amount (100% fee), zeroing balance

talvasconcelos:
Internal invoices are already marked `SUCCESS`, no need to check for status in funding source (apparently when using CLNRest it send a negative value) that would effectively make the receiver balance as 0!

Closes #3816
Closes #3869
Closes #3907

### DIFF
--- a/lnbits/core/services/payments.py
+++ b/lnbits/core/services/payments.py
@@ -1074,7 +1074,9 @@ async def _send_payment_notification_in_background(
     send_payment_notification_in_background(wallet, payment)
 
 
-async def update_invoice_callback(checking_id: str) -> Payment | None:
+async def update_invoice_callback(
+    checking_id: str, internal: bool = False
+) -> Payment | None:
     """
     Takes a checking_id of an incoming payment, from either paid_invoices_stream()
     or internal_invoice_queue. Checks its status, updates and returns it.
@@ -1088,12 +1090,13 @@ async def update_invoice_callback(checking_id: str) -> Payment | None:
         logger.warning(f"Payment '{checking_id}' is not incoming, skipping.")
         return None
 
-    status = await check_payment_status(
-        payment, skip_internal_payment_notifications=True
-    )
-    payment.fee = status.fee_msat or payment.fee
-    # only overwrite preimage if status.preimage provides it
-    payment.preimage = status.preimage or payment.preimage
+    if not internal:
+        status = await check_payment_status(
+            payment, skip_internal_payment_notifications=True
+        )
+        payment.fee = status.fee_msat or payment.fee
+        # only overwrite preimage if status.preimage provides it
+        payment.preimage = status.preimage or payment.preimage
     payment.status = PaymentState.SUCCESS
     await update_payment(payment)
     if payment.fiat_provider:
--- a/tests/unit/test_services_payments.py
+++ b/tests/unit/test_services_payments.py
@@ -370,6 +370,7 @@ async def test_internal_invoice_callback_does_not_zero_receiver_balance(
 
     payment = await get_payment(checking_id)
     wallet = await get_wallet(wallet.id)
+    assert wallet is not None
 
     funding_source.get_invoice_status.assert_not_awaited()
     assert (payment.fee, wallet.balance_msat) == (0, amount_msat)
--- a/tests/unit/test_services_payments.py
+++ b/tests/unit/test_services_payments.py
@@ -369,11 +369,11 @@ async def test_internal_invoice_callback_does_not_zero_receiver_balance(
     await update_invoice_callback(checking_id, internal=True)
 
     payment = await get_payment(checking_id)
-    wallet = await get_wallet(wallet.id)
-    assert wallet is not None
+    _wallet = await get_wallet(wallet.id)
+    assert _wallet is not None
 
     funding_source.get_invoice_status.assert_not_awaited()
-    assert (payment.fee, wallet.balance_msat) == (0, amount_msat)
+    assert (payment.fee, _wallet.balance_msat) == (0, amount_msat)
 
 
 @pytest.mark.anyio
--- a/tests/unit/test_services_payments.py
+++ b/tests/unit/test_services_payments.py
@@ -11,6 +11,7 @@ from lnbits.core.crud import (
     create_wallet,
     get_payment,
     get_payments,
+    get_wallet,
     update_payment,
 )
 from lnbits.core.models import (
@@ -31,6 +32,7 @@ from lnbits.core.services.payments import (
     create_payment_request,
     get_payments_daily_stats,
     settle_hold_invoice,
+    update_invoice_callback,
     update_pending_payment,
     update_pending_payments,
     update_wallet_balance,
@@ -42,6 +44,7 @@ from lnbits.wallets.base import (
     InvoiceResponse,
     PaymentFailedStatus,
     PaymentPendingStatus,
+    PaymentStatus,
     PaymentSuccessStatus,
 )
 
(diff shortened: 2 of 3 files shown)


lnbits/core/services/payments.py
```diff
@@ -1088,12 +1090,13 @@ async def update_invoice_callback(checking_id: str) -> Payment | None:
         logger.warning(f"Payment '{checking_id}' is not incoming, skipping.")
         return None
 
-    status = await check_payment_status(
-        payment, skip_internal_payment_notifications=True
-    )
-    payment.fee = status.fee_msat or payment.fee
-    # only overwrite preimage if status.preimage provides it
-    payment.preimage = status.preimage or payment.preimage
+    if not internal:
```
dni, line 1093:
```suggestion
    if not payment.is_internal:
```

lnbits/core/services/payments.py
```diff
@@ -1074,7 +1074,9 @@ async def _send_payment_notification_in_background(
     send_payment_notification_in_background(wallet, payment)
 
 
-async def update_invoice_callback(checking_id: str) -> Payment | None:
+async def update_invoice_callback(
+    checking_id: str, internal: bool = False
```
dni, line 1078:
```suggestion
    checking_id: str
```

tests/unit/test_services_payments.py
```diff
@@ -341,6 +344,38 @@ async def test_check_transaction_status_and_payment_status(mocker: MockerFixture
...
+    )
+    mocker.patch(
+        "lnbits.core.services.payments.get_funding_source",
+        return_value=funding_source,
+    )
+
+    await update_invoice_callback(checking_id, internal=True)
+
+    payment = await get_payment(checking_id)
+    _wallet = await get_wallet(wallet.id)
```
motorina0, line 372: why not use the top wallet?

motorina0 (approved):
LGTM

tests/unit/test_services_payments.py
talvasconcelos: it didn't allow the reassigment! tests failed.

dni:
superseded by https://github.com/lnbits/lnbits/pull/3984

PR #3982, opened 2026-05-20
fix: stricter asset upload

motorina0:
## Summary

Only allow image uploads. Older non image uploads will force download (not shown in the browser)


dni (approved):

talvasconcelos (approved):
tested. 

denies non image asset uploads

PR #3775, opened 2026-02-10
feat: create TaskManager singleton class

dni:
closes #2657
closes #3987 

- simple task monitoring
- cleaner interface
- no more circular import issues with tasks
- easier to register invoice_listener for extensions example usage in extensions: https://github.com/lnbits/lnurlp/pull/128
- backwards compatible

PR #3987, opened 2026-05-22
fix: use iterative loop instead of recursion in catch_everything_and_restart

dadofsambonzuki:
The `catch_everything_and_restart` function used recursion to restart failed background tasks, causing the call stack to grow by one frame every 5 seconds on each failure. Over long-running instances (12+ hours), this accumulated ~8,640 frames deep, gradually congesting the asyncio event loop until the service became unresponsive.

Replaced recursion with a `while True` loop to eliminate stack growth.

### DIFF
--- a/lnbits/tasks.py
+++ b/lnbits/tasks.py
@@ -61,7 +61,7 @@ def cancel_all_tasks() -> None:
 async def catch_everything_and_restart(
     func: Callable[[], Coroutine],
     name: str = "unnamed",
-) -> Coroutine:
+) -> None:
     while settings.lnbits_running:
         try:
             return await func()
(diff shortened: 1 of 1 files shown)


mercie-ux:
Tested this locally by simulating repeated task failures the ```while True``` loop keeps stack depth flat no matter how many restarts happen. A good fix.

dni:
love it good catch, ill incorparete your commit into #3775.

motorina0 (approved):

dni:
@dadofsambonzuki added your commit here. https://github.com/lnbits/lnbits/pull/3775/changes/fb415bbfbbe75711a6af7d61d233fc8c152c8e20

PR #4001, opened 2026-06-08
feat: use cached rate from exchange_rate_history_setting

dni:
performance improvement for wallet page, use already fetched rate from background task if available


mercie-ux:
I have tested this locally, when the background task has already stored a rate, `_get_price()` returns it immediately and `btc_price()` never fires. Falls back to the live API correctly when no cached rate exists. Clean improvement.

blackcoffeexbt (approved):

motorina0 (approved):

PR #4004, opened 2026-06-09
fix(payments): propagate per-call fee_limit_msat

Forte11Cuba:
## Summary
 
`POST /api/v1/payments` now accepts an optional `fee_limit_msat`. The service computes `effective_cap = min(caller, operator_cap)` so neither party  can raise the other's ceiling the operator's `LNBITS_RESERVE_FEE_PERCENT` remains the upper bound, but the caller can tighten it per request.

## What changed 

- New optional field `fee_limit_msat` on `CreateInvoice`.
- New capability flag `Wallet.supports_fee_limit_msat` (default `False`). Drivers that already propagate natively opt in: `lndrest`, `lndgrpc`, `clnrest`, `corelightning`, `corelightningrest`, `spark`, `sparkl2`, `lnbits`.
- The `lnbits → lnbits` driver now includes `fee_limit_msat` in its outgoing payload when the caller supplies one.
- Service raises `HTTPException(400)` if a caller passes a cap to a backend that cannot honor it, rather than silently dropping it.
- Abstract `Wallet.pay_invoice` signature widened to `int | None` to match the contract; drivers that arithmetically use the value narrow back via `assert`.                                                                                                                                             

## Test plan
                
- [x] Unit tests for `_resolve_fee_limit_msat`: None / caller<op / caller>op / unsupported backend.
- [x] Unit test for `lnbits → lnbits` payload propagation (present when set, absent when not).
- [x] Regression test asserting 12 custodial drivers keep `supports_fee_limit_msat = False`.
- [x] `ruff check lnbits/wallets/` clean.
- [x] `mypy lnbits/wallets/` clean (28 source files, 0 errors).

### DIFF
--- a/lnbits/core/models/payments.py
+++ b/lnbits/core/models/payments.py
@@ -269,6 +269,14 @@ class CreateInvoice(BaseModel):
     fiat_provider: str | None = None
     labels: list[str] = []
     external_id: str | None = Query(default=None, max_length=256)
+    fee_limit_msat: int | None = Query(
+        default=None,
+        ge=0,
+        description=(
+            "Caller-supplied cap on routing fees in millisatoshis. "
+            "Applied as min(fee_limit_msat, operator cap)."
+        ),
+    )
 
     @validator("payment_hash")
     def check_hex(cls, v):
--- a/lnbits/core/services/payments.py
+++ b/lnbits/core/services/payments.py
@@ -5,6 +5,7 @@ from datetime import datetime, timedelta, timezone
 from bolt11 import Bolt11, MilliSatoshi, Tags
 from bolt11 import decode as bolt11_decode
 from bolt11 import encode as bolt11_encode
+from fastapi import HTTPException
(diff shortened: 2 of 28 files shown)


dni (changes requested):
sorry there is no way to allow the user to set the fee_limit this is a setup to be exploited by a user sending to his own node and draining your funds.

motorina0:
AI Slop.
Please make a smaller PR and use human description.

Forte11Cuba:
> sorry there is no way to allow the user to set the fee_limit this is a setup to be exploited by a user sending to his own node and draining your funds.

Thanks for the review. Quick clarification, the design is min(caller, operator_cap), the caller can only lower the fee ceiling, never raise it. LNBITS_RESERVE_FEE_PERCENT stays the hard cap and the amount reserved from the user's wallet is unchanged, so a self-payment attacker can't make the node pay more than it already can today, only less.

PR #4007, opened 2026-06-14
feat: Bark backend support (using barkd)

prusnak:
Tested sending and receiving sats. Worked fine.

However, I added this to documentation:

> Bark fee estimates can be higher than LNbits' default minimum routing fee reserve. If small outgoing payments fail with an error like `fee of 20000 msat exceeds limit of 5000 msat`, raise the minimum reserve fee under **Settings → Funding** or set:
> 
> ```bash
> LNBITS_RESERVE_FEE_MIN=20000
> ```


blackcoffeexbt:
Thank you for the PR @prusnak !

I've been testing today. Are you running LNbits with the barkd funding source against an existing barkd wallet?

I was getting `ERROR | The backend for BarkWallet isn't working properly: 'Server error: 'No wallet set''` errors on start up with a new barkd daemon running.

In other funding sources e.g. Phoenixd & SparkL2 we allow a user to specify a mnemonic in LNbits > Settings would this be a good pattern for the Bark funding source or do you think that we should always use an already created wallet?

image

prusnak:
> I've been testing today. Are you running LNbits with the barkd funding source against an existing barkd wallet?

Yes, I used the existing barkd wallet.

> In other funding sources e.g. Phoenixd & SparkL2 we allow a user to specify a mnemonic in LNbits > Settings would this be a good pattern for the Bark funding source or do you think that we should always use an already created wallet?

Bark mnemonic only covers the onchain wallet of barkd, not the offchain (Ark, Lightning, etc.) funds. So I prefer to leave the PR as it is, so only people who know what they are doing will use this backend.

prusnak:
I updated the `docs/guide/wallets.md` to contain the following info:


> Initialize the Bark wallet before starting LNbits using `bark create`, then run `barkd`
 separately, keep its data directory persistent, and set `BARK_API_TOKEN` to the auth token from `~/.bark/auth_token`.

.env.example
```diff
@@ -144,6 +144,10 @@ CLICHE_ENDPOINT=ws://127.0.0.1:12000
 SPARK_URL=http://localhost:9737/rpc
 SPARK_TOKEN=myaccesstoken
 
+# BarkWallet
+BARK_API_ENDPOINT=http://localhost:3000
+BARK_API_TOKEN=barkdBearerToken
```
prusnak: Nit:

```suggestion
BARK_API_TOKEN=auth_token
```

blackcoffeexbt:
> I updated the `docs/guide/wallets.md` to contain the following info:
> 
> > Initialize the Bark wallet before starting LNbits using `bark create`, then run `barkd`
> > separately, keep its data directory persistent, and set `BARK_API_TOKEN` to the auth token from `~/.bark/auth_token`.

ty

> > I've been testing today. Are you running LNbits with the barkd funding source against an existing barkd wallet?
> 
> Yes, I used the existing barkd wallet.
> 
> > In other funding sources e.g. Phoenixd & SparkL2 we allow a user to specify a mnemonic in LNbits > Settings would this be a good pattern for the Bark funding source or do you think that we should always use an already created wallet?
> 
> Bark mnemonic only covers the onchain wallet of barkd, not the offchain (Ark, Lightning, etc.) funds. So I prefer to leave the PR as it is, so only people who know what they are doing will use this backend.

Makes sense. I'm not yet 100% up to speed on Bark. The PR looks good to me. I'm doing some more testing on it then will approve if all is good.

bitkarrot:
Arf Arf! I approve, this dog lover approves 🐶

blackcoffeexbt:
I’ve tested this and it works nicely. The main limitation AFAICT is that `bark` does not support `description_hash` invoices, so LNURL-pay, including lightning addresses, are not compatible when bark is used as the funding source.

blackcoffeexbt:
I've added a generic supports_description_hash property to the Bark funding source that allows for nicer error handling 

PR here @prusnak https://github.com/prusnak/lnbits/pull/1

blackcoffeexbt (approved):

blackcoffeexbt (approved):

Psycarlo:
I like this

lnbits/wallets/bark.py
```diff
@@ -0,0 +1,382 @@
...
+        self, bolt11: str, checking_id: str
+    ) -> PaymentResponse | None:
+        try:
+            r = await self.client.post(
+                "/api/v1/lightning/pay",
+                json={"destination": bolt11},
+                timeout=40,
+            )
+            r.raise_for_status()
+            r.json()
```
motorina0: why not return a payment response here?
is there no `preimage` or `status` to be checked?

lnbits/wallets/bark.py
```diff
@@ -0,0 +1,382 @@
...
+                ok=False,
+                checking_id=checking_id,
+                error_message=self._http_error_message(exc.response),
+            )
+        except httpx.RequestError as exc:
+            message = f"Unable to connect to {self.endpoint}."
+            logger.warning(message)
+            logger.warning(exc)
+            return PaymentResponse(
+                ok=False, checking_id=checking_id, error_message=message
```
motorina0: Are we sure that the payment is always failed in this case?
There is a risk of double spent if the payment is marked as failed in lnbits (not subtracted from the wallet balance) but on the funding source it actually goes trough.

(same for the above `httpx.HTTPStatusError`)

motorina0:
I see that `async def paid_invoices_stream(self) -> AsyncGenerator[str, None]:` is not implemented.

This means the incoming payment status will only be updated on scheduled checks or if the user explicitly requests so.
Is that intended?

prusnak:
> I see that `async def paid_invoices_stream(self) -> AsyncGenerator[str, None]:` is not implemented.

Added in 5a0bfd4d3a2d849246d5632440d3b2b92b4511a3

lnbits/wallets/bark.py
prusnak: fixed in 612741f5446512c383dd81fdf529216a4d0db8b4

lnbits/wallets/bark.py
prusnak: fixed in 612741f5446512c383dd81fdf529216a4d0db8b4

motorina0 (approved):

PR #4016, opened 2026-06-19
feat: sign outgoing webhooks with per-wallet secret

### LINKED ISSUE #1192: [Feature request] Webhook post bodies should have a signature to verify authenticity
When receiving webhook events from LNBits, I would like to be able to verify that it was actually my LNBits instance which triggered the webhook.

This seems to be a common problem for payments app developers, and it seems that both stripe and btcpay have something called a "webhook secret" which is created on the payment app, and then the POST body of the webhook is signed using that secret. This allows the receiving app to verify that it was the payment app (LNBits, stripe, btcpay) which triggered the webhook.

Examples:
Stripe:
https://stripe.com/docs/webhooks/signatures
BTCPay
https://docs.btcpayserver.org/VirtueMart/#23-create-a-webhook-on-btcpay-server

dsbaars:
Closes #1192.

When LNbits posts a webhook after an invoice is paid, there's currently no way for the receiver to check that the request actually came from their LNbits instance. This adds an HMAC signature to outgoing webhook requests so the receiving side can verify authenticity, similar to how Stripe and BTCPay do it.

### How it works

Every wallet gets a `webhook_secret`. When an invoice with a webhook is paid, the request body is signed with that secret and the signature is sent in an `LNbits-Signature` header:

```
LNbits-Signature: t=<unix_timestamp>,v1=<hmac_sha256_hex>
```

The signed value is `"{timestamp}.{raw_body}"`, hashed with HMAC-SHA256 using the wallet's `webhook_secret` as the key. This is the same scheme we already use on the inbound side for Stripe (`check_stripe_signature`), so the two directions stay consistent. The timestamp lets receivers reject replayed requests.

The secret is shown in the wallet view under "Node URL, API keys and API docs", with the usual reveal and copy controls and a reset button.

### Breaking change: webhook body encoding

This also fixes a long-standing bug where the webhook body was double encoded. The old code did `client.post(..., json=payment.json())`, and since `payment.json()` already returns a string, httpx wrapped it in another layer of JSON, so receivers got a quoted JSON string instead of a JSON object. Outgoing webhooks now send the payment object as a normal `application/json` body.

If you have a receiver that worked around the old behaviour (for example by JSON decoding twice, or parsing a quoted string), update it to read the body as plain JSON. This applies to all webhooks, signed or not.

### Notes for reviewers

- Signing happens in `dispatch_webhook`, which is the single place core uses to send payment webhooks, so there's one spot to reason about.
- The body is serialized once to bytes, those exact bytes are both hashed and sent over the wire (via `content=`), so the signature always matches what the receiver gets.
- Behaviour is gated by `LNBITS_WEBHOOK_SIGNING_ENABLED` (default on). New wallets get a secret on creation. Wallets created before this change have no secret until it's reset from the wallet view, and their webhooks go out unsigned until then, so nothing breaks for existing setups.
- Shared wallets don't expose the source wallet's secret (same handling as adminkey and inkey).

### Verifying on the receiver side

```python
import hashlib
import hmac
import time


def verify_lnbits_webhook(raw_body, signature_header, secret, tolerance_seconds=300):
    items = dict(i.split("=", 1) for i in signature_header.split(","))
    timestamp = int(items["t"])
    if abs(time.time() - timestamp) > tolerance_seconds:
        return False
    signed_payload = f"{timestamp}.{raw_body.decode()}"
    expected = hmac.new(
        secret.encode(), signed_payload.encode(), hashlib.sha256
    ).hexdigest()
    return hmac.compare_digest(expected, items["v1"])
```

### Testing

- New unit tests in `tests/unit/test_webhook_signatures.py` cover the signature output, a round trip against the existing verifier, a tampered body, a wrong secret, the `Content-Type` header, the disabled and no-secret cases, and the path where the wallet is loaded via `get_wallet` instead of being passed in.
- New API test in `tests/api/test_wallet_api.py` for the reset endpoint (success, persistence, and a 404 when another user tries to reset a wallet they don't own).
- Updated the existing notification test for the new `dispatch_webhook` argument.
- Tried it end to end on a local instance with FakeWallet: created an invoice with a webhook, paid it, and confirmed the receiver got the `LNbits-Signature` header and that it validates against the wallet secret (and fails with a wrong secret).

### Changes

- Migration `m046` adds `webhook_secret` to `wallets`
- `webhook_secret` on the `Wallet` model, generated in `create_wallet`
- `LNBITS_WEBHOOK_SIGNING_ENABLED` setting
- `create_webhook_signature` helper and signing in `dispatch_webhook`
- `PUT /api/v1/wallet/webhook-secret/reset/{wallet_id}` to regenerate the secret
- Wallet UI additions for viewing and resetting the secret
- Docs at `docs/devs/webhooks.md`


lnbits/core/services/notifications.py
```diff
@@ -237,24 +240,56 @@ def _send() -> bool:
...
+
     async with httpx.AsyncClient(headers=headers) as client:
         try:
             check_callback_url(payment.webhook)
         except ValueError as exc:
             await mark_webhook_sent(payment.payment_hash, "-1")
             logger.warning(f"Invalid webhook URL {payment.webhook}: {exc!s}")
         try:
-            r = await client.post(payment.webhook, json=payment.json(), timeout=40)
+            r = await client.post(payment.webhook, content=body, timeout=40)
```
motorina0, line 301: why change `json` to `content`
`body = payment.json().encode()`

dsbaars: It's literally in the PR body...

docs/devs/webhooks.md
```diff
@@ -0,0 +1,62 @@
...
+```python
+import hashlib
+import hmac
+import time
+
+
+def verify_lnbits_webhook(
+    raw_body: bytes,
+    signature_header: str,
+    secret: str,
```
motorina0, line 48: Stripe is not necessarily the best model.
If there are two parties that receive callbacks from the same wallet then both will have access to the `secret`. This means `Party A` can create valid callbacks and send it to `Party B`.

We could instead use `wallet.webhook_secret` as a private key and sign the message with it.
The parties receiving the callbacks will only get the public key (derived from `wallet.webhook_secret`_key)

dsbaars:
Rebased onto current `dev` (`1c25383`), so this is mergeable again.

One thing worth flagging for review: the migration is renumbered from `m046` to **`m051`**. `dev` has since taken 046 through 050 (`m046_add_permissions_to_installed_extensions` through `m050_add_lightning_address_to_wallets`), and since `run_migration` compares `mNNN` against a single stored integer rather than matching by name, leaving this at 046 would have made `dev`'s own 046 unreachable on any database that ran this one first.

The other conflict was in `lnbits/core/services/notifications.py`, where `dev` added `dispatch_payment_notification`. Both are kept; `dispatch_webhook` is `dev`'s body plus the signing block.

Bundles regenerated with `make bundle`. `tests/unit` shows the same failure set before and after the rebase, with 8 more passing (this PR's own tests); ruff, black, prettier and `make checkbundle` are clean.

docs/devs/webhooks.md
motorina0: @dsbaars what do you think about this?

PR #4022, opened 2026-06-23
fix: get rid of url rewrites for extensions `/upgrades/`

dni:
should open the way for custom root-path https://github.com/lnbits/lnbits/pull/3748. based on my PoC. https://github.com/dni/fastapi-dynamic-routes-poc

### DIFF
--- a/lnbits/app.py
+++ b/lnbits/app.py
@@ -2,6 +2,7 @@ import asyncio
 import glob
 import importlib
 import os
+import shutil
 import sys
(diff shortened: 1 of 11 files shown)


motorina0 (approved):

talvasconcelos (approved):
Tested.,with an extension update without restarting. 
Routes keep normal extension paths and no /upgrades/… routes show up, the extension continued working after update.

PR #4026, opened 2026-06-26
feat: use lnbits price aggregator to save loads of request

dni:
more reliable and fast price queries, one request instead of querying all exchanges. will save a lot of traffic for saas and for everyone running many lnbits instances.

- opt-out if you want to run the rates on your lnbits.
- https://price.lnbits.com/ open-source so you can run it yourself.

aggregator


talvasconcelos (approved):
Tested

arcbtc (approved):

PR #4028, opened 2026-06-29
feat: optimise Uvicorn Settings for heavy web socket use

dadofsambonzuki:
Uvicorn's default `ws_ping_interval=20` / `ws_ping_timeout=20` can non-deterministically close local WebSocket connections (e.g. NWC provider ↔ nostrclient relay) under event-loop load. The client side already sends its own pings, so server-side pings are redundant here.

See discussion on the related nostrclient PR: https://github.com/lnbits/nostrclient/pull/71#issuecomment-4803033307

### DIFF
--- a/lnbits/server.py
+++ b/lnbits/server.py
@@ -28,7 +28,7 @@ from lnbits.settings import set_cli_settings, settings
     "--reload", is_flag=True, default=False, help="Enable auto-reload for development"
 )
 @click.option("--ws-max-queue", default=128, help="Websocket max queue size")
-@click.option("--ws-ping-timeout", default=900.0, help="Websocket ping timeout")
+@click.option("--ws-ping-timeout", default=300.0, help="Websocket ping timeout")
 def main(
     port: int,
     host: str,
--- a/lnbits/server.py
+++ b/lnbits/server.py
@@ -37,7 +37,7 @@ def main(
     ssl_certfile: str,
     reload: bool,
     ws_max_queue: int,
-    ws_ping_timeout: float
+    ws_ping_timeout: float,
 ):
     """Launched with `uv run lnbits` at root level"""
 
--- a/lnbits/server.py
+++ b/lnbits/server.py
@@ -58,6 +58,8 @@ def main(
             ssl_keyfile=ssl_keyfile,
             ssl_certfile=ssl_certfile,
             reload=reload or False,
+            ws_ping_interval=None,
+            ws_ping_timeout=None,
         )
 
         server = uvicorn.Server(config=config)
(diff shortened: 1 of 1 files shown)


dni:
how about raising the default of `ws_ping_timeout` from 20.0 seconds to something like a 15 mins or an hour? i don't see why disabling ping altogether.  that may have some sideeffects in browsers.

dni (changes requested):
by disabling it we have no chance to detect zombie sockets, that would be bad.

dni:
https://uvicorn.dev/settings/#implementation

another very interesting setting for me is `--ws-max-queue <int> - Set the maximum length of the WebSocket incoming message queue. Only available with the websockets protocol. Default: 32.`

defaulting to only 32, which i think with a couple of relays and other things we could run into this limit. that maybe also a cause of disconnects.

dadofsambonzuki:
Also, which implementation:

>ws <str> - Set the WebSockets protocol implementation. Either of the websockets and wsproto packages are supported. There are two versions of websockets supported: websockets and websockets-sansio. Use 'none' to ignore all websocket requests. Options: 'auto', 'none', 'websockets', 'websockets-sansio', 'wsproto'. Default: 'auto'.

This PR should prbly be: 'Optimise Uvicorn Settings for heavy web socket use'.

dni:
i think `auto` is fine we use `"websockets~=15.0.1"`

dni (approved):

dni:
maybe 15 mins is too long. what do we think? i think 5 mins maybe fine also.

lnbits/server.py
```diff
@@ -27,13 +27,17 @@
 @click.option(
     "--reload", is_flag=True, default=False, help="Enable auto-reload for development"
 )
+@click.option("--ws-max-queue", default=128, help="Websocket max queue size")
+@click.option("--ws-ping-timeout", default=900.0, help="Websocket ping timeout")
```
dni:
```suggestion
@click.option("--ws-ping-timeout", default=300.0, help="Websocket ping timeout")
```

lnbits/server.py
```diff
@@ -27,13 +27,17 @@
 @click.option(
     "--reload", is_flag=True, default=False, help="Enable auto-reload for development"
 )
+@click.option("--ws-max-queue", default=128, help="Websocket max queue size")
+@click.option("--ws-ping-timeout", default=300.0, help="Websocket ping timeout")
```
dni:
```suggestion
@click.option("--ws-ping-timeout", default=60.0, help="Websocket ping timeout")
```

motorina0 (approved):

PR #4037, opened 2026-07-04
fix(nwc): decode bolt11 before NWC call to prevent payment record loss

dadofsambonzuki:
Some NWC transactions were being silently dropped from the DB despite being paid as the transaction is being rolled back at the database level.

The flow:
1. _pay_external_invoice calls create_payment (INSERT) inside a db.connect() transaction
2. NWC pay_invoice sends the request to the rizful relay — the provider processes it
3. But the relay response is lost/delayed, so self.conn.call("pay_invoice", ...) raises an exception
4. The outer except Exception in nwc.py:637 returns PaymentResponse(error_message=msg) — checking_id is None because payment_hash was never extracted
5. _pay_external_invoice sees checking_id is None, marks the payment FAILED, then raises PaymentError
6. The exception propagates out of db.connect() → SQLAlchemy rolls back the entire transaction — the create_payment INSERT is undone
7. The payment record vanishes from the database entirely
8. NWC provider already sent the sats, but LNbits has zero record

### DIFF
--- a/lnbits/wallets/nwc.py
+++ b/lnbits/wallets/nwc.py
@@ -10,6 +10,7 @@ from collections.abc import AsyncGenerator, Awaitable, Callable
 from typing import Any, cast
 from urllib.parse import parse_qs, unquote, urlparse
 
+from bolt11 import Bolt11Exception
 from bolt11 import decode as bolt11_decode
 from coincurve import PrivateKey, PublicKey
 from Cryptodome.Cipher import ChaCha20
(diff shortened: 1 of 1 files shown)


dadofsambonzuki:
@blackcoffeexbt

motorina0 (approved):

dni (approved):

lnbits/wallets/nwc.py
```diff
@@ -579,11 +579,11 @@ async def status(self) -> StatusResponse:
             return StatusResponse(str(e), 0)
 
     async def pay_invoice(self, bolt11: str, fee_limit_msat: int) -> PaymentResponse:
+        invoice_data = bolt11_decode(bolt11)
```
dni: you should put this in its own try / catch bolt11 decoding can throw an Bolt11Exception. return a failed paytment_repsonse then.

dni (changes requested):

dadofsambonzuki:
Superseded by #4066

PR #4051, opened 2026-07-09
Add fiat webhook dispatcher for extensions

dadofsambonzuki:
## Summary
Adds a core fiat webhook dispatch hook so extensions can handle verified provider webhook events without claiming shared callback URLs.
Also fixes DynamicComponent extension route loading so extension sub-routes can be added even when the extension root route is already registered.
## Changes
- Add register/unregister helpers for fiat webhook handlers.
- Dispatch verified Stripe, PayPal, Square, and Revolut events to matching extension handlers before existing core fallback handling.
- Extract provider metadata/source for handler matching.
- Deduplicate dynamic extension routes by route name instead of skipping route loading when the extension root path already exists.
## Context
This supports extensions such as SatsPay using the core-owned `/api/v1/callback/{provider}` endpoints while preserving existing core fiat webhook behavior as the fallback path.
Related extension PRs:
- https://github.com/lnbits/events/pull/64
- https://github.com/lnbits/satspay/pull/87

### DIFF
--- a/lnbits/core/views/callback_api.py
+++ b/lnbits/core/views/callback_api.py
@@ -1,4 +1,5 @@
 import json
+from collections.abc import Awaitable, Callable
 
 from fastapi import APIRouter, Request
 from loguru import logger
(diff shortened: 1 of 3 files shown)


lnbits/core/views/callback_api.py
```diff
@@ -31,6 +32,82 @@
...
+    metadata = _extract_fiat_webhook_metadata(provider, event)
+    source = metadata.get("source")
+    source = source.lower() if isinstance(source, str) else None
+
+    for handler_provider, handler_source, handler in list(fiat_webhook_handlers):
+        if handler_provider != provider:
+            continue
+        if handler_source and handler_source != source:
+            continue
+        handled = await handler(provider, event, metadata)
```
Angella-Mulikatete, line 103: This call isn't wrapped in a try/except. I tested this locally — registered a handler that raises, sent a Stripe event through it, and the exception propagates straight out of the route instead of being caught. Since this runs before handle_stripe_event() is called at line 130, core's own fallback never executes either. Screenshot of the failing test attached below in the main review comment.

lnbits/core/views/callback_api.py
```diff
@@ -31,6 +32,82 @@
 from lnbits.settings import settings
 
 callback_router = APIRouter(prefix="/api/v1/callback", tags=["callback"])
+FiatWebhookHandler = Callable[[str, dict, dict], Awaitable[bool]]
+fiat_webhook_handlers: list[tuple[str, str | None, FiatWebhookHandler]] = []
+
+
+def register_fiat_webhook_handler(
+    provider: str, source: str | None, handler: FiatWebhookHandler
```
Angella-Mulikatete, line 40: Should source be required instead of optional? Line 101 means a handler registered with source=None intercepts every event for that provider before core gets a chance at it.

Angella-Mulikatete (commented):
To test if the dispatcher works as described, I registered a fake extension handler tagged source="myext". Sent one Stripe event tagged for that extension, and one plain untagged Stripe event. The log confirms it: the first event triggers "Fiat webhook handled by extension handler. Provider: 'stripe'." (line 105 of callback_api.py) — the extension correctly claimed it. No such line appears after the second, untagged event, meaning it correctly fell through toward core's own handling instead. Test passes.

Image

To test what happens if a registered handler is buggy, I registered a handler that does nothing but raise an exception, then sent a normal Stripe event through the same endpoint. The traceback shows the crash happening inside dispatch_fiat_webhook_event at callback_api.py:103 . The exception isn't caught anywhere — it propagates all the way out as an unhandled RuntimeError, and core's own handle_stripe_event never gets a chance to run. This confirms one broken extension handler can block core's webhook processing entirely for that provider.

Image

motorina0:
Why is this needd?
I do not understand the scenario.

dadofsambonzuki:
> Why is this needd? I do not understand the scenario.

This is related to the two linked PRs. It is required in order to process fiat payments outside of core in the satspay extension (v2). This is to support fiat check-out for events in my case.

motorina0:
> This is related to the two linked PRs. It is required in order to process fiat payments outside of core in the satspay extension (v2). This is to support fiat check-out for events in my case.

The referenced PRs are too big (very hard to review). It is better to have many small PRs than a monster one.

I still do not understand why what core offers is not sufficient.

- the extension is notified when a payment is successful
- the extension can check if the payment is a fiat one and take the proper actions

dadofsambonzuki:
Notwithstanding the comment on the large PRs (they are!), you are correct - core already gives me what I need.

This dispatcher was introduced when I was leveraging to fiat code in core, but handling the config of the fiat providers in satpay. I have since simplified my satspay implementation to use both the fiat code and config in core and so no longer need the dispatcher.

PR #4053, opened 2026-07-10
Add lnaddress to host.py

arcbtc:


motorina0:
Maybe its better to add utils for `resolveLnurl(amount)` which returns a payment request.
Then use the normal wallet pay functionality.

arcbtc:
> Maybe its better to add utils for `resolveLnurl(amount)` which returns a payment request. Then use the normal wallet pay functionality.

The payout happens inside `recordCoinflipPayment`, after an invoice is paid, that callback runs as an event, not as a logged-in user request, so the host needs the event wallet context to safely know which wallet it is allowed to pay from.

motorina0:
addressed separately
closing this pr

image

PR #4058, opened 2026-07-13
Fix: compatibility for postgres

talvasconcelos:
There was a timestamp compatibility issue with PostgreSQL. WASM storage uses Unix timestamps, but PostgreSQL expects timestamp values. It now uses LNbits existing timestamp_placeholder helper to convert them into the correct database format. A small _value_placeholder helper was added to apply this conversion only to datetime fields.

### DIFF
--- a/lnbits/core/crud/extensions.py
+++ b/lnbits/core/crud/extensions.py
@@ -253,13 +253,14 @@ async def get_wasm_invocation_stats(
     since: datetime | None = None,
     conn: Connection | None = None,
 ) -> WasmInvocationStats:
+    database = conn or db
     where: list[str] = []
     values: dict = {}
     if extension_id:
         where.append("extension_id = :extension_id")
         values["extension_id"] = extension_id
     if since:
-        where.append("started_at >= :since")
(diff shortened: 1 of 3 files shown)


lnbits/core/crud/extensions.py
```diff
@@ -314,11 +315,13 @@ async def delete_old_wasm_invocations(
...
+    database = conn or db
+    placeholder = database.timestamp_placeholder("cutoff")
+    result = await database.execute(
+        f"""
         DELETE FROM wasm_invocations
-        WHERE status != 'running' AND started_at < :cutoff
-        """,
+        WHERE status != 'running' AND started_at < {placeholder}
+        """,  # noqa: S608
         {"cutoff": cutoff},
```
motorina0, line 325: is this still needed?

dni (approved):
:o

lnbits/core/crud/extensions.py
talvasconcelos: yes! ruff doesn't like it...

PR #4061, opened 2026-07-15
fix: Infinite invoice listener loop with VoidWallet

blackcoffeexbt:
Removes the `VoidWallet.paid_invoices_stream()` override.

**Issue**

VoidWallet yielded an empty checking ID and immediately completed. Since the funding-source invoice producer is registered as a permanent task, it restarted immediately and repeatedly logged:

`No incoming payment found for ''.`

This affects fresh installations because `VoidWallet` is the default backend.

**Why removing should be safe**

The override was added when `paid_invoices_stream()` was abstract and every wallet needed to provide an async generator implementation.

`Wallet` now provides a fallback stream that polls pending invoices and sleeps. `VoidWallet` has no pending invoices, so inheriting it correctly keeps the listener idle without emitting a fake payment notification or causing a restart loop.

### DIFF
--- a/lnbits/wallets/void.py
+++ b/lnbits/wallets/void.py
@@ -1,5 +1,3 @@
-from collections.abc import AsyncGenerator
-
 from loguru import logger
 
 from .base import (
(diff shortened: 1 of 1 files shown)


dni (approved):

dni:
nice!

motorina0 (approved):

PR #4064, opened 2026-07-16
Fix wasm invocation limits check

riccardobl:
`start_wasm_invocation` tracks an invocation only after `create_wasm_invocation` has completed.

If many requests are submitted in quick succession, they can bypass the limit because they will be processed before `create_wasm_invocation` returns (being it async) and, so, before `start_wasm_invocation` has a chance to track them.

This PR fixes the race condition by adding an async lock around invocation creation and tracking.

### DIFF
--- a/lnbits/core/services/extensions.py
+++ b/lnbits/core/services/extensions.py
@@ -66,6 +66,7 @@ class WasmInvocationHandle:
 
 _wasm_invocation_lock = RLock()
 _wasm_invocation_ready_lock = asyncio.Lock()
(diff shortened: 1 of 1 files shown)


motorina0:
It fixes the race, but it means every  WASM invocation start waits behind every other WASM invocation start, across all extensions and users. If create_wasm_invocation() is slow, startup throughput drops.

riccardobl:
I made a new commit that fixes the race condition by using a dictionary of pending invocations, without locking the entire `start_wasm_invocation`

PR #4067, opened 2026-07-17
i18n: add Faroese (fo) localization

krvi:
This PR introduces the initial Faroese (fo) localization for LNbits.


motorina0:
Thank you!

PR #4068, opened 2026-07-17
fix: return preimages for Phoenixd invoices and settled payments

clankwright:
## Summary

Same-node / NWC pays against Phoenixd-backed LNbits can settle successfully while returning an unusable (missing or all-zero) preimage. That breaks L402 and any NIP-47 client that needs `SHA256(preimage) == payment_hash`.

Two gaps:

1. **Phoenixd `createinvoice`** does not include `paymentPreimage`. The preimage is already available from `GET /payments/incoming/{paymentHash}` (even before the invoice is paid). Without storing it, internal settles copy a null preimage into the payer row.
2. **`check_transaction_status`** (and the internal branch of `check_payment_status`) returned `PaymentSuccessStatus` **without** the payment’s stored `preimage`. The NWC provider then substitutes the FakeWallet placeholder (`0000…0000`).

## Changes

- `lnbits/wallets/phoenixd.py`: after `createinvoice`, if `paymentPreimage` is absent, fetch `preimage` from `/payments/incoming/{paymentHash}` (best-effort; warn and continue if the follow-up fails).
- `lnbits/core/services/payments.py`: pass `preimage=payment.preimage` (and fee on the internal path) when returning success from `check_transaction_status` / internal `check_payment_status`.

## Test plan

- [ ] Phoenixd funding source: create invoice via API; confirm response/DB row has a non-empty 64-hex preimage (not only after pay).
- [ ] Same-node internal pay between two LNbits wallets: payer payment row has real preimage; `check_transaction_status` returns that preimage.
- [ ] NWC `pay_invoice` against an invoice from another wallet on the same instance: response `preimage` is 64-hex and not all-zeros; `sha256(preimage) == payment_hash`.
- [ ] Non-Phoenixd funding source: create/pay still behave as before (no regression on status/preimage for backends that already return preimage).

## Context

Validated on a Phoenixd + LNbits + NWC stack used for L402 settle (merchant invoice → NWC `pay_invoice` → retry with `Authorization: L402 macaroon:preimage`).

### DIFF
--- a/lnbits/core/services/payments.py
+++ b/lnbits/core/services/payments.py
@@ -640,9 +640,7 @@ async def check_transaction_status(
 async def check_payment_status(payment: Payment) -> PaymentStatus:
     if payment.is_internal:
         if payment.success:
-            return PaymentSuccessStatus(
-                fee_msat=payment.fee, preimage=payment.preimage
-            )
+            return PaymentSuccessStatus(fee_msat=payment.fee, preimage=payment.preimage)
         if payment.failed:
             return PaymentFailedStatus()
         if payment.is_in and payment.fiat_provider:
--- a/lnbits/core/services/payments.py
+++ b/lnbits/core/services/payments.py
@@ -632,7 +632,7 @@ async def check_transaction_status(
         return PaymentPendingStatus()
 
     if payment.status == PaymentState.SUCCESS.value:
-        return PaymentSuccessStatus(fee_msat=payment.fee)
+        return PaymentSuccessStatus(fee_msat=payment.fee, preimage=payment.preimage)
 
     return await check_payment_status(payment)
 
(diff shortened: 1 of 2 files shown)


motorina0 (approved):

talvasconcelos (approved):

motorina0:
- GET /payments/incoming/{paymentHash} in Api.kt: https://github.com/ACINQ/phoenixd/blob/c7abded8a1b6dbdfc3672a243edbd98dc24f2342/src/commonMain/kotlin/fr/acinq/phoenixd/Api.kt#L262
  - The returned IncomingPayment JSON includes preimage: https://github.com/ACINQ/phoenixd/blob/c7abded8a1b6dbdfc3672a243edbd98dc24f2342/src/commonMain/kotlin/fr/acinq/phoenixd/json/JsonSerializers.kt#L149
  - Phoenixd CLI also calls the same route for getincomingpayment: https://github.com/ACINQ/phoenixd/blob/c7abded8a1b6dbdfc3672a243edbd98dc24f2342/src/commonMain/kotlin/fr/acinq/phoenixd/cli/PhoenixCli.kt#L177

PR #4069, opened 2026-07-18
fix: detect extension zip by real path on startup restore

arbadacarbaYK:
## Summary
- Fix `check_installed_extension_files()` to use `ext.zip_path.is_file()` instead of `f\"./{ext.zip_path}\" not in glob(...)`.
- The old comparison never matched absolute paths, so `download_archive()` deleted a good local zip and re-downloaded whenever extension files were missing from the container FS.

## Impact
Docker hosts that only persist `/data` (e.g. Umbrel) can mass-call `deactivate_extension()` after container recreate when network/DNS is briefly unavailable — even though local zips were present.

## Test plan
- [ ] Remove `extensions/<id>/`, keep `data/zips/<id>.zip`, restart offline → extract from local zip, do not deactivate
- [ ] Zip missing + network OK → still downloads
- [ ] Normal boot with files present → unchanged


arbadacarbaYK:
Closing — opened from a stale fork tip. Replacing with a clean single-commit PR based on current upstream/dev.

PR #4085, opened 2026-07-22
fix: 12 word mnemonic

motorina0:


blackcoffeexbt:
Tested and verified functionality. LGTM

PR #4088, opened 2026-07-23
needed for permssions to trigger from wasm extension json

arcbtc:


motorina0:
Fix works fine. Included commits here: https://github.com/lnbits/lnbits/pull/4089

PR #4095, opened 2026-07-24
fix(admin): add missing service_fee i18n keys to en.js

Bartok9:
English admin showed raw service_fee_* keys; they were in de.js but missing from en.js. Add the three strings to en.js.

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim:** `sera`


motorina0:
> Stop English admin UI from rendering raw missing keys (service_fee_hint, service_fee_max, service_fee_max_hint).

it looks like these labels are only missing from en.js, but are present in other language files (`lnbits/static/i18n/de.js`)
The fix is not to remove them from the UI, but to add them to `lnbits/static/i18n/en.js`

motorina0:
Please keep the description short and concise. PRs with rich AI description will be ignored/closed.

Bartok9:
Good call — reverted the template change and instead added the missing `service_fee_hint`, `service_fee_max`, and `service_fee_max_hint` keys to `en.js` (matching `de.js`). Also trimmed the description. Thanks!

motorina0 (approved):

PR #4097, opened 2026-07-26
fix(wallet): use payment_processing i18n key for NFC pay spinner

Bartok9:
NFC pay spinner used missing `processing_payment` i18n key; switch to existing `payment_processing` (same as payInvoice).

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim:** `sera`


motorina0 (approved):

Bartok9:
CI note: the single-line wallet.js i18n key change is independent of static vendor bundles. bundle/checkbundle fails because regenerating bundle-components.min.js differs from the committed artifact on this runner (not from this PR path). Happy to refresh committed min bundles in a follow-up if maintainers want that separate.

Thanks @motorina0 for the approve.

Agent-Owner: sera

PR #4105, opened 2026-07-27
feat: Amboss Payments funding source

bufo24:
## What

Adds `AmbossWallet`, a new LNbits funding source backed by the [Amboss Payments](https://www.amboss.tech) GraphQL API. Receive and send Lightning payments from an Amboss Payments wallet, with every payment recorded in the Amboss ledger.

## How it works

- **Receive:** `create_receive` → invoice; status polled via `find_one(id)`.
- **Send (non-custodial):** the node admin macaroon is fetched encrypted and decrypted in-process with the team password (Argon2id + NIP-44 v2, mirroring the [`@ambosstech/payments`](https://github.com/AmbossTech/sdk/tree/main/packages/payments) SDK). The node is used **only** to pay the invoice created by `create_send`, never for lookups.
- `pay_invoice` returns `checking_id = payment_hash`. LNbits already keys outgoing payments by the bolt11 payment_hash and only rewrites `checking_id` when a funding source returns a different one, so echoing the hash back keeps the ledger and core aligned and matches how `get_payment_status` resolves sends.
- **Error reporting:** `AmbossApiError` separates "rails answered with an error" (a GraphQL `errors` array, an error status code, an unusable body) from "rails was unreachable" (transport failures), so a mistyped wallet id, a dead API key or a 429 is reported as itself rather than as a connection problem.
- **Node socket:** a send targets the wallet's litd REST socket when there is one, otherwise lnd, matching the SDK's `selectSendNode` order. Only `https` sockets qualify, since the decrypted admin macaroon travels on that connection; a wallet with no https socket fails before `create_send` rather than after.
- `get_payment_status(payment_hash)` resolves sends via `find_one(lightning: {payment_hash})` on the ledger; `get_invoice_status` resolves receives via `find_one(id)`.
- Per-wallet send context (team id, node, decrypted macaroon) is cached so repeat sends skip the Argon2 + node-permission round-trips and settle within the pay-invoice wait window.
- Optional sandbox testing config (`amboss_sandbox` / `amboss_sandbox_auto_complete`): when enabled, invoices and sends carry metadata that tells the Amboss sandbox API to auto-settle the transaction shortly after creation, instead of leaving it pending until natural expiry.
- **LNURL-pay support**: `create_invoice` accepts `description_hash` / `unhashed_description` (standard `Wallet` params) and sends the resulting hash on `create_receive`'s `bolt11.description_hash` field, so invoices generated for LNURL-pay (LUD-06) carry a `description_hash` matching `sha256(metadata)`, the check paying wallets perform before settling.

## Changes

- `lnbits/wallets/amboss.py` (new): the funding source, including a `python -m lnbits.wallets.amboss` self-check that verifies the ported Argon2id/NIP-44 crypto reproduces known-good test vectors. The self-check also runs in pytest.
- `tests/wallets/test_amboss.py` (new): unit tests plus tests that drive the wallet against a local mock server over real httpx, so the GraphQL documents are actually executed.
- `lnbits/settings.py`, `lnbits/wallets/__init__.py`: registration + allowed funding sources list.
- `lnbits/static/js/components/admin/lnbits-admin-funding-sources.js`, `lnbits/templates/components/admin/funding_sources.vue`: admin UI entry; also adds a small `text` flag to the shared field config so non-secret fields (like an endpoint URL) can render unmasked instead of via the password-style toggle used for secrets.

## Test plan

- [x] `ruff`, `black`, `mypy`, and the crypto self-check pass; `make bundle` committed so `make checkbundle` is green.
- [x] Manually verified end-to-end against both a production and a sandbox Amboss environment: receive, send, and status polling for both.
- [x] Verified the ported Argon2id/NIP-44 implementation reproduces fixtures generated by the reference SDK's own crypto libraries.
- [x] Verified LNURL-pay support end-to-end: created an invoice via `unhashed_description` with a realistic LNURL-pay metadata payload and confirmed the resulting BOLT11's `description_hash` matches `sha256(metadata)` exactly.
- [x] `pytest tests/wallets/` passes (316 passed, 20 skipped). The suite drives the wallet against a local mock server over real httpx, so the GraphQL documents are executed rather than mocked away.
- [x] Every guard in `amboss.py` is mutation-tested (18/18): breaking any of the asset checks, the `_gql` error classification, the `payment_request` cross-check, the https socket predicate, `_resolve_node`'s node scan, the password length check, any of `status()`'s connection/API-error/asset/balance branches, the in-band return for an unusable socket, the ledger-miss → PENDING mapping, or the `description_hash` on the wire fails exactly the test written for it.

## Notes

- Supports BTC (BASE_ASSET) wallets only. Taproot Asset wallets are rejected in `status()`, `create_invoice()` and `pay_invoice()`, so a mismatched wallet errors before any transaction is booked rather than only on a live send.


dsbaars (commented):
Reviewed the branch locally and tested the funding source end to end against the Amboss sandbox API.

**Verdict: CHANGES_REQUESTED**, but nothing blocking. Three concrete, closeable items.

## Verified live against the sandbox API

Run with a sandbox service API key against `https://rails.amboss.tech/graphql`, using the PR's own wallet code (`AmbossWallet`), not standalone curl calls.

| What | Result |
|---|---|
| `status()` | `StatusResponse(error_message=None, balance_msat=0)` |
| `create_invoice(1000, memo=...)` | `ok=True`, bolt11 for 1000000 msat, `d` tag set |
| `create_invoice(..., unhashed_description=...)` | the bolt11's `description_hash` is **exactly** `sha256(metadata)`, and the `d` tag is correctly dropped |
| `get_invoice_status(tx_id)` | `paid=None` to `paid=True` within 6s (sandbox auto-complete) |
| `pay_invoice(bolt11)` | `checking_id` equals the bolt11 payment_hash |
| `get_payment_status(payment_hash)` | via `find_one(lightning:{payment_hash})`, resolves to `paid=True` within 8s |

All four testable claims from the PR description hold. LNURL-pay support works exactly as described: the backend drops the `d` tag as soon as `bolt11.description_hash` is set, which is what the code comment says.

### Crypto port

Not testable live (the sandbox has no node and no team password), but verified against the SDK source (`packages/payments/src/crypto/argon2.ts`, `decryptAdminMacaroon.ts`). The functions were extracted and run standalone: the official NIP-44 v2 vector and all Argon2 fixtures pass.

- params `{dkLen:32, t:3, m:64000, p:4}` = `length=32, iterations=3, memory_cost=64000, lanes=4`
- salt `teamId.trim().toLowerCase()` = `team_id.strip().lower()`
- second step salted with `password.trim()`, hex master key as the password = `_argon2id(master_key.encode(), password.strip().encode())`
- envelope masterKey -> symmetricKey -> macaroon, identical

The MAC is verified before decryption using `hmac.compare_digest`, so authenticate-then-decrypt in the correct order.

### Other things verified correct

- **Fee sign is right.** Core normalizes it itself: `-(abs(status.fee_msat or 0) + abs(service_fee_msat))` in `lnbits/core/services/payments.py:905`, and the balances view does `SUM(amount - abs(fee))`. The `abs()` matches `lnbits/wallets/lndrest.py:223`.
- **`checking_id = payment_hash` works, but not for the stated reason.** Core sets checking_id to the payment_hash itself (`payments.py:810`) and rewrites it via `update_payment(new_checking_id=...)` if the funding source returns something else. So "requires the funding source to echo it back" is inaccurate; the behavior is still correct. An empty string is harmless because that update is guarded by `if new_checking_id and ...`.
- **The payment_request cross-check before paying** (`amboss.py:328-343`) is a genuinely good defense, better than what some existing funding sources do.
- **TLS is correct.** `ssl.create_default_context(cadata=tls_cert)` trusts only that certificate; CPython skips `load_default_certs` as soon as cadata is supplied. No `verify=False` anywhere.
- **Status enum mapping is exact:** the live schema has `PaymentsTransactionStatus {COMPLETED, EXPIRED, FAILED, PENDING}`.
- **The shared `prop.text` change in the .vue is backwards compatible**, and `amboss_team_password` correctly stays masked.

## P1: pre-dispatch failures are reported as PENDING instead of FAILED

`amboss.py:281-303`. The `except Exception` returns `PaymentResponse(error_message=...)`, where `ok` defaults to `None`, and LNbits reads `None` as PENDING.

Demonstrated live:

```
pay_invoice("this-is-not-a-bolt11")  ->  ok=None  checking_id=None
get_payment_status("00"*32)          ->  paid=None   ("Payments transaction not found")
```

What falls into this bucket and provably never left the process: an undecodable bolt11, a missing `amboss_team_password` (defaults to None, no startup validation), a wrong password, a team password shorter than 8 characters (verified: `Argon2id: salt must be at least 8 bytes`, because the password is the salt of the second KDF), no LND REST socket, and any create_send GraphQL error.

Traced through core:

- the balances view counts pending outgoing, so the balance is locked
- `get_payment_status` finds nothing in the ledger and stays pending forever
- `_verify_external_payment` raises "Payment is still pending" on a retry, so there is no self-recovery
- `check_pending_payments()` only runs at startup and only looks back 15 days

The codebase models this distinction explicitly: `payment_request_was_rejected()` in `lnbits/wallets/base.py:18` and `_is_pre_dispatch_payment_error()` in `lnbits/wallets/lndrest.py:36`. Everything before the `_gql(_CREATE_SEND, ...)` line is provably pre-dispatch and should return `ok=False`; only from create_send onward is it genuinely ambiguous.

## P1: the BASE_ASSET check only runs on the send path of a non-sandbox wallet

The check at `amboss.py:382` lives in `_resolve_node`, reachable only via `_send_context` (behind `not is_sandbox`), called only from `pay_invoice`. `status()` and `create_invoice()` never check.

Demonstrated live with a real TAPROOT_ASSET wallet from the same account:

```
status()                         -> StatusResponse(error_message=None, balance_msat=0)   no complaint
create_invoice(amount=1000)      -> ok=True                                              no complaint
bolt11 amount_msat               -> 1000000        (LNbits books this as 1000 sats)
actual tx on rails               -> amount_sats "1000", asset {symbol USDTL, type TAPROOT_ASSET}
```

LNbits books 1000 sats where the ledger records 1000 USDTL. The PR description claims "Taproot Asset wallets raise a clear error"; that only holds for live sends. The guard belongs in `__init__`, or in `status()` and `create_invoice()`.

## P1: `bundle-components.min.js` was not rebuilt

The bundle is tracked, contains `lnbits-admin-funding-sources` and `blink_send_without_probe`, and contains zero occurrences of "amboss". `make checkbundle` is part of `make check` (`Makefile:7`, `Makefile:139-147`) and diffs the rebuilt bundle against the committed one. CI fails on this, and admins see no Amboss fields in production because the UI loads the bundle, not the source JS.

Run `make bundle` and commit.

## P2

- **The rails ledger is the only status source for a payment your own node made.** The docstring says the node is "never for lookups". Meanwhile core cancels the pay task after `lnbits_funding_source_pay_invoice_wait_seconds` (default 5) via `asyncio.wait_for`, while `_pay_via_node` waits up to 35s. So for live sends the LND result is usually discarded and recovery depends entirely on rails observing a node-to-node payment. lndrest recovers via TrackPaymentV2 against the node itself. With `_map_tx_status` treating FAILED and EXPIRED as terminal, a rails timeout on a still in-flight HTLC would refund the user for sats that already left the node. Not testable in the sandbox (no node), so this is the main open question.
- **Only `lnd.rest`, never `litd.rest`.** `_GET_NODE_PERMISSIONS` selects both, `_resolve_node` accepts only `lnd`. The SDK's `selectSendNode` tries litd first and only then falls back to lnd. A litd-only node yields "no LND REST endpoint available", which then lands in the pending bug above.
- **Tests touch none of the money paths.** `_self_check()` sits only under `if __name__ == "__main__"` and appears in no Makefile, workflow, or pyproject entry, so the crypto is never proven in CI; a `def test_self_check(): _self_check()` closes that. Both tests mock `_gql`, so no GraphQL document is ever executed. The wallet was not added to `tests/wallets/fixtures/json/fixtures_rest.json` (currently corelightningrest, lndrest, alby, eclair, lnbits, phoenixd), which is the project's harness for REST funding sources. The referenced `scripts/amboss-fixture` is not in the repo.

## P3

- The docstring (`amboss.py:8-10`) contradicts the code: it says checking_id is the Amboss tx id and that status is polled by id. True for receives, false for sends since commit `40b4faf26`.
- `__init__` validates the api key and wallet id but not `amboss_api_endpoint`, while Blink and Alby both do. Clearing the field yields `TypeError: Invalid type for url` instead of a clean ValueError.
- No minimum length on the team password, so anything shorter than 8 characters produces an opaque Argon2 ValueError that disappears into the pending bug.
- `nip44_decrypt` skips the spec's padded-length validation and the 6-byte length path. Not exploitable, since the MAC check comes first.
- Argon2id (64 MiB, t=3, p=4) runs twice synchronously on the event loop on the first send.
- No scheme guard on `rest_host`, so an `http://` socket would put the admin macaroon on the wire in cleartext.
- No lock around populating `_send_cache`, so concurrent first sends duplicate the Argon2 work.

## Bottom line

The integration works. Receive, send, status polling, and LNURL-pay were tested end to end against the sandbox and do exactly what the PR description promises. The crypto port is a faithful port of the SDK.

What still needs to happen before merge:

1. Return `ok=False` for pre-dispatch failures instead of pending.
2. Move the asset guard into `__init__`, or into `status()` and `create_invoice()`.
3. Run `make bundle` and commit the bundle.
4. Wire `_self_check()` into pytest, and ideally add an entry to `fixtures_rest.json`.
5. Answer this: does rails reliably reconcile a node-to-node payment, given that the node itself is never queried?

bufo24:
Thanks for the thorough review — pushed fixes for all three P1s plus most of the P2/P3s in `d3344f82`..`54976d4a`.

1. **Pre-dispatch failures → `ok=False`** (P1): split `pay_invoice`'s error handling so everything before `create_send` (bad bolt11, missing/short team password, no LND/litd endpoint) returns `ok=False` instead of `ok=None`. Only the `create_send` call itself — which may have applied server-side before an error surfaces — stays ambiguous.
2. **BASE_ASSET guard** (P1): now checked in `status()` and `create_invoice()` too, not just on the send path, so a Taproot Asset wallet errors before booking a receive in the wrong unit. Cached after the first check so it's not a round-trip on every call.
3. **Bundle rebuild** (P1): ran `make bundle`; `bundle-components.min.js` now includes the Amboss admin fields and `make checkbundle` passes.
4. **litd fallback** (P2): `_resolve_node`/`_pay_via_node` now accept `litd.rest` sockets too, preferring litd over lnd — checked against the SDK's actual `selectSendNode` (litd first, lnd fallback) rather than guessing at the order.
5. **Crypto self-check in CI** (P2): `_self_check()` is now wired into pytest.
6. P3s: fixed the stale docstring, added `amboss_api_endpoint` validation at init (matches Blink/Alby), a team-password length check (it's the second Argon2 salt), a TLS-scheme guard on the resolved REST socket, moved Argon2id off the event loop via `asyncio.to_thread`, and added a lock around first-fill of the send context so concurrent first sends can't duplicate that work.

Verified: ruff, mypy, and the wallet test suite (297 passed) all green; added a regression test covering the pre-dispatch/create_send split.

Left open, deliberately:
- **`fixtures_rest.json` entry** (P2, "ideally"): that harness assumes one REST endpoint per method; this wallet's single GraphQL endpoint doesn't fit it cleanly without more work than a quick pass, so I skipped it — open to revisiting if you think it's worth it.
- **Reconciliation question** (item 5): confirmed with the author — rails is intentionally the sole source of truth for send outcome, even for payments this instance dispatched itself; the node's own REST response is best-effort only (core's wait window is usually shorter than the node call anyway). Documented that as a comment on `get_payment_status` in `54976d4a` so it doesn't read as an oversight to the next reviewer.
- **nip44 padding validation** (P3): left as-is per your own note that it's not exploitable (MAC check precedes it).

bufo24:
Follow-up on the two test items, pushed in `a4185fc6`.

**On `fixtures_rest.json` — I looked into it properly and I'd argue against adding Amboss there.** Reasons, in case you disagree and want to overrule me:

- `_apply_mock` matches on `uri`, `method`, `headers` and `query_params` only. Request-body matching is dead code: it reads `getattr(mock.dict(), "request_type", None)`, and since `mock.dict()` is a dict that always evaluates to `None` (there's a comment above it noting it doesn't work). `request_body`/`request_type` are only honoured by the NWC and RPC harnesses. Amboss sends every operation to one URI, one method, identical headers — so the harness can't tell `status` from `create_invoice` from `pay_invoice`.
- Consequently, methods that make more than one call (`create_invoice` now does the asset check plus `create_receive`; a sandbox `pay_invoice` does send-context plus `create_send`) can only be expressed with a single merged response containing both subtrees at once — a fixture that doesn't resemble any real rails response. And a non-sandbox `pay_invoice` can't be expressed at all: three rails calls plus a node call to an `https://` host with a real Argon2/NIP-44 macaroon envelope.
- 28 test cases across the five functions; 23 would need a hand-written Amboss blob. `WalletTest.tests_for_funding_source` sets `skip = True` when a test's `mocks` lacks the funding-source key, so a partial entry yields silently skipped tests that read as green.
- Precedent: Blink, the other single-endpoint GraphQL funding source, isn't in the harness either — it has its own `test_blink.py`. Same for Bark. Only 10 of 27 wallet classes use the shared harnesses.

**What I did instead**, since your underlying point (no GraphQL document is ever executed) is fair: five tests that drive the wallet against a local mock server over real httpx — balance mapping, the `errors` array surfacing its message, the Taproot guard short-circuiting *before* an invoice exists, the LNURL-pay `description_hash` actually going out on the wire, and a not-yet-visible send reading pending rather than failed. They use a private server on an ephemeral port, not the session-scoped `httpserver` that `test_rest_wallets.py` pins to 8555.

I mutation-tested them rather than trusting green: my first drafts of two passed *even with the asset guard and the errors handling removed* (an incidental `KeyError` produced the same observable result). Both are now tightened so reverting either fix fails the test.

**One more CI find:** `make check` was also red on `checkblack`, independently of the bundle — the nested ternary in `create_invoice` fails `black --check` on the pristine PR head. Rewritten as plain control flow.

Also dropped `paid_invoices_stream`, which only re-yielded `super()`; its comment moved to the class docstring.

**Unrelated, for a separate issue if you think it's worth one:** the `request_type` bug in `tests/wallets/test_rest_wallets.py:62` silently disables body matching for every REST funding source, not just this one. I left it alone as out of scope here.

bufo24:
Cleared the last loose ends from your review in `6229b9d4`, plus one gap I found while re-checking them.

**Found while re-reading the asset guard:** `pay_invoice` still had no asset check for **sandbox** wallets. The guard lived in `_resolve_node`, which is only reached when `not is_sandbox`, so a sandbox Taproot Asset wallet could send with the wrong unit accounting even after the `status()`/`create_invoice()` fix. `pay_invoice` now runs the check itself (cached, so live sends pay for it once). Test included; it fails if the guard is removed.

**`checking_id` comment** — corrected per your finding. It claimed core "fails the payment unless `pay_invoice` returns the same `checking_id`", which isn't true: core sets `checking_id` from the payment_hash itself at `payments.py:810` and only rewrites it via the `if payment_response.checking_id and ...` guard when a funding source returns something different. Behaviour was already fine, the stated reason wasn't. The same wrong claim was in the PR description, so I fixed that too.

**`scripts/amboss-fixture`** — dropped the dangling reference; the docstring now just names the SDK libraries the fixtures came from (`@noble/hashes` argon2id, nostr-tools `nip44.v2`), which is the part that's actually verifiable.

**Also:** dropped `amount_sats` from `_TX_FIELDS`, since nothing reads it.

One process note: no CI has run on this branch yet — `gh pr checks` reports no checks and the head commit has zero check-runs, presumably pending first-time-contributor approval. Locally `ruff`, `black --check`, `mypy` are clean, `make checkbundle` passes, and `pytest tests/wallets/` is 304 passed / 20 skipped. Happy to chase anything CI turns up once it can run.

dsbaars (commented):
Re-reviewed everything from `ddcb8822` through `6229b9d4`, including a live run against the Amboss sandbox and my own mutation tests against the new suite.

All three P1s from the first review are genuinely fixed, and I verified the two behavioural ones live rather than by reading:

| Check | Result |
|---|---|
| `pay_invoice("not-a-bolt11")` | `ok=False`, `error='Bech32 string is not valid.'` |
| `create_invoice(1000)` on a real TAPROOT_ASSET wallet | `ok=False`, `error='AmbossWallet only supports BASE_ASSET (BTC) wallets'` |
| LNURL receive, end to end | `description_hash` matches `sha256(metadata)`, settles in 12s |
| Bundle | a fresh `npm run bundle` produces no diff, so `checkbundle` is genuinely reproducible |

I also mutation-tested the new suite independently. Removing the asset guard, the `errors` handling in `_gql`, the `bolt11.description_hash` payload, or flipping the ledger-miss mapping from pending to failed each makes exactly one test fail. The tests are load-bearing, and the request-body assertion in the LNURL test is the right shape for this.

Good catch on the sandbox send path. That third entry point was a real gap and I missed it when I asked for the guard in `status()` and `create_invoice()`. Confirmed against `a4185fc6`: `pay_invoice` had no check at all, and the only one lived in `_resolve_node` behind `not is_sandbox`. The new call is in the right place, inside the pre-dispatch block, so a Taproot wallet fails terminally with nothing sent.

Your reading of the harness is right too. `getattr(mock.dict(), "request_type", None)` is `getattr` on a dict, so it is always `None` and both body-matching branches are dead. With that gone the harness can only discriminate on uri, method, headers and query params, and every Amboss operation is identical on all four. Blink and Bark are absent for the same reason. Agreed on leaving it out, and the `request_type` bug is worth its own issue.

The `checkblack` find is also correct, and it predates this PR's later commits: the nested ternary fails `black --check` on the original head too.

## One blocker

**`test_pay_invoice_rejects_payment_request_hash_mismatch` stopped testing what it is named for, as a side effect of `6229b9d4`.**

The new `_ensure_base_asset()` call hits the generic `_gql` mock first. That mock is shaped as a `create_send` response and has no `wallet` key, so it raises `KeyError('wallet')`, which the pre-dispatch handler turns into `ok=False`. That happens to satisfy both assertions, and the second `bolt11_decode` never runs.

```
as committed:                     pay_invoice:351  send error (pre-dispatch): 'wallet'
with _asset_verified pre-set:     pay_invoice:395  send error: payment_request does not
                                                   match submitted invoice (hash/amount mismatch)
```

Both pass, but only the second exercises the guard. That guard is the one your own comment describes as the thing standing between a third-party API response and the node paying an invoice nobody asked for, so it should stay covered. Setting `amboss_wallet._asset_verified = True` before the call restores it, or give the `_gql` mock a `side_effect` that tells the balance query apart from `create_send`.

## Further findings

Three more from this round. None of them can lose or lock funds, since the pre-dispatch split now catches all three as clean terminal failures.

**[P2] Socket selection and TLS validation disagree, and the check sits on the wrong side of `create_send`.**

`_pick_rest_socket` returns litd's `rest` whenever it is truthy, with no scheme check. `_pay_via_node` then rejects anything that is not `https://`. Reproduced:

```
sockets: litd=http://litd-host:8080, lnd=https://lnd-host:8080
_pick_rest_socket picks -> http://litd-host:8080
_pay_via_node accepts   -> False        the usable https lnd socket on the same node is never tried
```

Three consequences. The rejection returns `ok=False` after `create_send` has already succeeded, so rails holds a send that lnbits has marked failed. It returns before the httpx `try`, so the 401/403 cache-bust never runs and the unusable node stays in `_send_cache` for the process lifetime. And a usable socket on the same node, or on a later node in the list, is skipped.

Whether this is reachable at all depends on Amboss only ever handing out `https://` sockets, which the SDK's own fixtures and `sendLndPayment` building `` `${restHost}/v2/router/send` `` both suggest. So this is latent rather than live. The fix is to move the `https://` predicate into `_pick_rest_socket`, so selection and validation are the same function and a total miss raises in `_resolve_node`, landing pre-dispatch before `create_send` is ever called.

**[P3] The team password length guard checks the wrong string.**

`len(self.team_password) < 8`, but the KDF salts the second Argon2id step with `password.strip()`. Reproduced: `"ab      "` is eight characters, passes the guard, and then dies on `salt must be at least 8 bytes`, which is the exact error the guard exists to prevent. A trailing newline out of an env file is the realistic case. `len(self.team_password.strip()) < 8` closes it.

**[P3] `status()` reports the asset error as a connectivity failure.**

The `ValueError` is raised inside the `try` whose `except` returns `StatusResponse(f"Unable to connect to {self.endpoint}.", 0)`. Confirmed live: pointing at a Taproot wallet gives `Unable to connect to https://rails.amboss.tech/graphql.` An operator will go and debug DNS. `create_invoice` gets this right and surfaces the real message.

## Nits

- `_ensure_base_asset()` now also runs on the non sandbox path, which already got the same check for free inside `_resolve_node`. One extra read on a cold cache, and `_resolve_node` still does not set `_asset_verified`.
- The asset guard test asserts `ok is False` but not the error string or the call count, and the pending test does not assert the mock was hit. Both catch the mutation they were written for; an always failing or always pending stub would still pass.
- The module docstring still says lnbits "requires pay_invoice to echo that hash back", which is the claim `6229b9d4` corrected further down in `pay_invoice`.

## On CI

Your process note checks out: the head commit has zero check runs and `gh pr checks` reports none, so nothing here has been verified by CI yet. Noting it so the state is on the record rather than assumed green. Locally I get the same as you, 304 passed and 20 skipped on `tests/wallets/`, with ruff, black and mypy clean and the bundle reproducible.

The vacuous test is the one I would want addressed, since it silently removes coverage that exists today. The socket predicate is worth doing while it is cheap. The remaining two are one-liners and I do not feel strongly about them.

bufo24:
All four addressed in `57d9a62b`. Thanks for catching the vacuum — that one was mine, and it's the more useful kind of find.

**Blocker: the hollowed-out test.** Confirmed exactly as you describe. Adding `_ensure_base_asset()` to `pay_invoice` made it reach the shared `_gql` mock first; that mock is shaped as a `create_send` response, so it died on `KeyError('wallet')` as a pre-dispatch failure and satisfied both asserts without the cross-check ever running. Fixed by pre-setting `_asset_verified` so the guard is reached, and the test now also asserts the specific error message rather than just `ok is False`.

The wider lesson I took: I mutation-tested the tests I *added* and not the one I *changed the behaviour underneath*. So I built a mutation harness over the whole file rather than spot-checking — nine guards, each mutated in turn:

| mutation | caught by |
|---|---|
| asset guard in `_ensure_base_asset` | both taproot tests |
| `pay_invoice` asset guard | `test_pay_invoice_rejects_taproot_asset_wallet` |
| `_gql` errors array | `test_gql_raises_the_graphql_error_message` |
| `payment_request` cross-check | `test_pay_invoice_rejects_payment_request_hash_mismatch` |
| https-only socket predicate | both `_pick_rest_socket` tests |
| `strip()` in the password check | `test_send_context_rejects_password_that_strips_below_argon2_salt` |
| `status()` asset branch | `test_status_reports_wrong_asset_as_itself...` |
| ledger-miss → PENDING | `test_get_payment_status_pending_while_tx_not_yet_in_ledger` |
| `description_hash` on the wire | `test_create_invoice_sends_lnurl_description_hash` |

9/9, each by the test written for it. Two of those tests didn't exist before this round: the `status()` branch survived my first pass, which is how I found it was untested.

**P2, socket selection vs validation.** Agreed, and I took your fix: the https predicate now lives inside `_pick_rest_socket`, so selection and validation are one policy. A plaintext-only node now fails in `_resolve_node` — inside the pre-dispatch block, before `create_send` — instead of after. `_pay_via_node`'s duplicate check is gone; it asserts non-None instead, since `_resolve_node` can no longer hand it an unusable socket. Covers all three consequences you listed, including the cache pollution.

**P3, password check on the wrong string.** Fixed to measure `strip()`ed length, with your trailing-newline case as the test.

**P3, `status()` misreporting.** The asset check now sits outside the connection-error handler and returns its own message. Live-visible improvement: an admin gets the real reason instead of a phantom DNS problem.

**Smaller points:** module docstring corrected (it still carried the echo-back claim that `6229b9d4` fixed further down); the repeated BASE_ASSET string is now one `_BASE_ASSET_ONLY` constant; the taproot and pending tests now assert messages and call counts, so an always-failing or always-pending stub wouldn't slip through.

On `_resolve_node`'s now-redundant asset check: I kept it deliberately. The field is already in that response so it costs nothing, and it's the last checkpoint before a real send. Happy to drop it if you'd rather have the single source of truth.

Local: `ruff`, `black --check`, `mypy` clean; `pytest tests/wallets/` 309 passed / 20 skipped; `make checkbundle` OK. Still zero CI check-runs on the head commit.

dsbaars (commented):
All four are fixed. I re-ran the verification rather than reading the diff, including my own mutations.

**The blocker.** The test now produces `pay_invoice:410 payment_request does not match submitted invoice (hash/amount mismatch)` instead of the `pre-dispatch: 'wallet'` KeyError, so it reaches the cross-check for real. Asserting the specific `error_message` is the right hardening; that is what would have caught it the first time.

Your read on why it happened is the useful part. You mutation-tested the tests you added, not the one whose behaviour you changed underneath. That distinction is worth keeping.

**The socket policy.** Unifying the predicate is exactly right, and I exercised the rewritten function directly:

| Sockets on one node | Picked |
|---|---|
| litd https, lnd https | litd |
| litd http, lnd https | lnd |
| plaintext only | none, node skipped |
| plaintext node followed by an https node | the later node |
| `HTTPS://` uppercase | accepted |
| bare `host:port`, empty string, missing key | none, no crash |

A plaintext-only wallet now fails inside `_resolve_node`, in the pre-dispatch block, before `create_send`. That closes all three consequences, cache pollution included.

**The other two.** The password guard measures the stripped length, and `status()` reports the asset error as itself. Both confirmed.

**Your mutation matrix holds.** I re-ran four of the nine guards independently, each caught by exactly the test written for it: the https predicate (two tests), the `payment_request` cross-check, the `status()` asset branch, and `strip()` in the password check. Full suite 309 passed and 20 skipped, with ruff, black and mypy clean.

## One thing left, P3

`_pay_via_node` now has `assert rest_host is not None` where a `PaymentResponse` used to be. It is unreachable today and provably so: `_pick_rest_socket` is pure and receives the same dict `_resolve_node` selected with. The concern is the failure mode if that invariant ever breaks, because the call sits outside all three try blocks in `pay_invoice`. The `AssertionError` escapes to core, `lnbits/exceptions.py:85` turns it into a 400 whose `detail` is the assert's message, which for a bare assert is the empty string. And since the exception escapes rather than returning a response, the `payment_response.failed` branch never runs, so the row stays PENDING rather than FAILED.

That is a worse outcome than what it replaced, on a path that moves money. Moving the check earlier was the right call; dropping the in-band return was the part I would reconsider. Either give the assert a message, or keep it in band:

```python
if rest_host is None:
    return PaymentResponse(ok=None, checking_id=checking_id, error_message="...")
```

which also survives `python -O`, where the assert compiles out and `rest_host` reaches httpx as `None`.

## Smaller

- `int(balance)` is still inside the connection `try` in `status()`, so a balance that will not parse still surfaces as "Unable to connect". Reading `asset_type` and branching before parsing the balance would close the last of that masking.
- The later-node skip is covered as a unit test on the predicate but not as a composition through `_resolve_node`'s `next(...)`. I exercised it in process and it selects the later node, so this is a coverage note rather than a defect.

## On your open question

I would keep the asset check in `_resolve_node`, for the reason you gave. The field is already in that response, it costs nothing, and it is the last checkpoint before a real send. Redundancy on a funds path is worth more than a single source of truth here.

## CI

Still zero check runs on the head commit and `mergeStateStatus` is BLOCKED, so nothing here has been verified outside our two local runs. That is the part that is not yours to fix.

bufo24:
Fixed in `7a5b2465`. Your P3 was right on the mechanism, and I verified the chain before changing it: `lnbits/exceptions.py:85` does register an `AssertionError` handler returning 400 with `detail: str(exc)`, empty for a bare assert — and because the exception escapes instead of returning, `payment_response.failed` never runs and the row stays PENDING. Worse than what it replaced, on a funds path, exactly as you said. The `python -O` point stands too.

**One deliberate divergence from your snippet:** you wrote `ok=None`, I used `ok=False`. At that point `_pay_via_node` has not built the httpx client, so nothing has been sent to the node — the payment definitively did not happen. `ok=None` would strand it pending, which is the round-one bug in a new place. `ok=False` matches your own round-two reasoning on the pre-`create_send` path: nothing reached the node, so FAILED is truthful and the rails tx simply expires unused. Say the word if you'd rather have the conservative version.

**Balance masking.** Closed. `asset_type` is read and branched on first, then the balance is parsed in its own handler that reports `"wallet response has no usable balance"` instead of a phantom connectivity error.

**The composition gap.** Now covered, and it was worth writing: two tests through `_resolve_node` — a plaintext node followed by an https node selects the later node, and a wallet with only plaintext nodes raises. The mutation that truncates the scan to `nodes[:1]` is caught by the first of those, which is precisely the case the predicate's unit tests could not see.

**Mutation battery, now 12 guards, 12 caught:**

| new mutation | caught by |
|---|---|
| in-band return → bare `AssertionError` | `test_pay_via_node_returns_in_band_when_socket_is_unusable` |
| balance error → "Unable to connect" | `test_status_reports_unusable_balance_as_itself` |
| `next(...)` truncated to the first node | `test_resolve_node_skips_plaintext_node_for_a_later_https_one` |

The nine from last round still hold, and the https predicate now trips three tests instead of two.

**On `_resolve_node`'s asset check:** agreed, keeping it. Redundancy on a funds path beats a single source of truth.

Local: ruff, black, mypy clean; `pytest tests/wallets/` 313 passed / 20 skipped; `make checkbundle` OK. Still zero CI check-runs on the head commit — the PR description now also documents the socket policy and the mutation coverage, so whenever a maintainer can approve the workflow run, everything here is claim-checkable rather than asserted.

dsbaars (commented):
You are right about `ok=False` and my suggestion was wrong. Worth spelling out why, because the reasoning generalises.

I mapped every return in the pay path:

| Line | ok | Situation |
|---|---|---|
| 374 | False | pre-dispatch |
| 384 | None | `create_send` itself, genuinely ambiguous |
| 394 | None | sandbox, rails settles |
| 408 | False | undecodable payment_request, node not contacted |
| 421 | False | hash/amount mismatch, node not contacted |
| 514 | False | no https socket, node not contacted |
| 554 | None | httpx error, node was contacted |

The boundary that matters is the node call, not `create_send`. `create_send` writes a ledger row; it does not move money. Lines 408 and 421 already sit after a successful `create_send` and already return `ok=False`, for exactly the reason you gave. My `ok=None` would have made 514 the odd one out and would have locked the balance for a payment that provably never left, which is the bug I spent two rounds pushing against. Your version is consistent with the file's own rule; mine would have broken it.

The docstring settles the premise too: sending is non-custodial, the node pays, so with no node call nobody pays.

## The rest of the commit

All three points closed. My own mutations:

| Mutation | Result |
|---|---|
| in-band return reverted to a bare `assert` | caught |
| `ok=False` changed to `ok=None` | caught, so the test pins the semantics and not just the shape |
| node scan truncated to `nodes[:1]` | caught |

Full suite 313 passed and 20 skipped, ruff, black and mypy clean.

## Two leftovers, both P3

**The masking is not fully gone.** `asset_type = wallet["asset"]["type"]` is still inside the connection `try`, so a missing or malformed `asset` key still reports "Unable to connect". That is the same shape you just fixed for `balance`, left in place one line above it. Low impact, since it needs schema drift rather than a plausible data variation like a decimal balance string, but it is the last of that pattern.

**A blind spot in the mutation battery, worth knowing because you rely on it.** The two new `_resolve_node` tests distinguish the nodes on two axes at once: the socket scheme, and `encrypted_macaroon` set to `"no"` and `"yes"`. I swapped the predicate in `_resolve_node` from `_pick_rest_socket(n["sockets"])` to `n.get("encrypted_macaroon") == "yes"` and all 19 tests still pass.

So the pair pins that the scan continues past a rejected node, which was the actual gap and is genuinely covered, but it does not pin that the rejection is the socket check. A battery run against this would report 12 of 12 while that mutation survives. Giving both fixture nodes the same `encrypted_macaroon` value closes it. The https policy itself stays pinned by the `_pick_rest_socket` unit tests either way, so the exposure is narrow.

## Status

Nothing left here touches a funds path. Still zero check runs on the head commit with `mergeStateStatus` at BLOCKED, so CI remains the one thing that has not verified any of this, and that is not yours to unblock.

bufo24:
Both closed in `01d90b23`.

**The masking, finished.** `asset_type` now reads outside the connection handler too. The rule in `status()` is now uniform: only the request itself may report a connection problem, and anything read out of the response reports itself — connection, asset shape, wrong asset, balance shape, in that order. You were right that leaving it one line above the fix I'd just made was the odd part.

**The blind spot — this is the better of the two findings, and it was a real hole in my method.** Confirmed your substitution: on `7a5b2465`, swapping the predicate to `n.get("encrypted_macaroon") == "yes"` passed all 19 tests. The fixtures differed on two axes at once, so the pair pinned *that the scan continues past a rejected node* without pinning *what made it rejectable*. Both nodes now carry the same `encrypted_macaroon`, so only the socket can decide, and node identity is asserted by object identity rather than by a payload field.

I added your mutation to the battery permanently, plus dropping the predicate entirely:

| new mutation | verdict |
|---|---|
| predicate → `encrypted_macaroon == "yes"` | caught (was surviving) |
| predicate dropped, first node wins | caught, by both `_resolve_node` tests |
| `status()` asset shape → "Unable to connect" | caught |

15 guards, 15 caught. Worth stating the general lesson, since it's the second time my own test design has been the weak link: a fixture that varies more than one axis can pin the *consequence* of a rule while leaving the *rule* unpinned, and a green battery reports that as full coverage. Same failure class as mutation-testing the tests I added but not the one whose behaviour I changed underneath — both are the harness looking healthier than it is.

**On `ok=False`:** thanks for mapping all seven returns and spelling out the node-call boundary. That table is a better statement of the file's rule than the comment I wrote, so I've left the code as is rather than paraphrasing it worse.

Local: ruff, black, mypy clean; `pytest tests/wallets/` 314 passed / 20 skipped; `make checkbundle` OK. Still zero check runs on the head commit.

dsbaars (commented):
Both closed, verified rather than read.

`status()` is uniform now: four layers, each with its own message, every `except` returning so nothing can be read unbound, and the catch sets right (`ValueError` only where `int()` is).

The blind spot is genuinely closed. The mutation that survived last round is now caught by two tests:

| Mutation | Last round | Now |
|---|---|---|
| predicate to `encrypted_macaroon` | survived | 2 tests fail |
| predicate dropped, first node wins | caught | 2 tests fail |
| asset shape folded back into "Unable to connect" | n/a | 1 test fails |
| balance shape folded back | caught | 1 test fails |

Full suite 314 passed and 20 skipped, ruff, black and mypy clean.

## Two leftovers, both P3

**The same failure class, one level deeper.** The fixtures still differ on two axes at once: socket kind (lnd vs litd) and scheme (http vs https). A predicate that ignores the scheme entirely still passes everything:

```python
next((n for n in perms["nodes"] if "litd" in n["sockets"]), None)   # 20 passed
```

That predicate would accept a plaintext litd socket, which is the exact property the socket policy exists to enforce. Giving both fixtures the same socket kind, so only the scheme differs, closes it. Worth noting that this is the same shape as the finding it was fixing: two axes pin the consequence, not the rule.

**`status()` still masks one layer up.** `_gql` raises `ValueError` for a GraphQL `errors` array, and that lands in the connection handler. Against a mock server:

```
GraphQL error "wallet not found"          -> "Unable to connect to .../graphql."
throttle "Too Many Requests"              -> "Unable to connect to .../graphql."
```

The endpoint answered in both cases. This is the most likely version of the problem in practice: a mistyped `amboss_wallet_id`, an expired API key, or rate limiting. I hit throttling myself against the sandbox during these reviews. It predates this commit, so it is not a regression, but it is the last gap in the rule your own comment states, that only the request itself may report a connection problem. A dedicated exception type out of `_gql` would let `status()` report it as itself.

Neither touches a funds path.

## Status

Nothing left that I would hold this on. Still zero check runs on the head commit with `mergeStateStatus` at BLOCKED, so CI remains the only thing that has not verified any of this.

bufo24:
Both closed in `2af57844`.

**The two-axis problem, third time.** I reproduced your mutation before touching anything — `next((n for n in perms["nodes"] if "litd" in n["sockets"]), None)` gave 20 passed on `01d90b23`. The fixtures now differ on exactly one axis: both nodes are litd, one `http` and one `https`, same macaroon. litd-over-lnd preference stays pinned by the `_pick_rest_socket` unit tests, where it belongs, so nothing is lost by making the composition test single-purpose.

Three rounds, same shape: macaroon+scheme, then kind+scheme. The rule I've taken from it is that a discriminating pair must differ in exactly the property under test, and any second difference is a place a mutation can hide. That now sits in a comment on the fixtures so the next person doesn't reintroduce it.

**The masking, one level up.** Fixed properly rather than patched: `AmbossApiError` is now raised whenever rails *answered* but with something unusable — a GraphQL `errors` array, an error status code, an unreadable body, or a missing `data` key. `httpx.RequestError` propagates untouched, since those genuinely never reached the endpoint. `status()` reports the two apart:

```
GraphQL error "wallet not found"   ->  "Amboss API error: wallet not found"
HTTP 429                           ->  "Amboss API error: HTTP 429"
connection refused / DNS / timeout ->  "Unable to connect to <endpoint>."
```

Your throttling case is now one of the tests. I also folded in `data is None`, which previously surfaced as a `KeyError` and would have been masked the same way.

Worth noting one deliberate side effect: `_gql` no longer raises `ValueError`, so `test_gql_raises_the_graphql_error_message` expects `AmbossApiError` now. Every production caller uses `except Exception`, so behaviour elsewhere is unchanged — the pre-dispatch classification in `pay_invoice` still treats an API error as terminal, which is right, since nothing has been sent.

**Battery now 18 guards, 18 caught**, with three additions:

| new mutation | verdict |
|---|---|
| predicate → `"litd" in n["sockets"]` | caught (was surviving on `01d90b23`) |
| API errors reported as unreachable | caught by both new `status()` tests |
| error status code not raised | caught by the throttle test |

Local: ruff, black, mypy clean; `pytest tests/wallets/` 316 passed / 20 skipped; `make checkbundle` OK. Still zero check runs on the head commit, so CI remains the one thing that hasn't seen any of this.

dsbaars (commented):
The exception change is the riskiest edit in this PR so far, so I audited that first. `AmbossApiError` subclasses `Exception`, not `ValueError`, and all eight `_gql` call sites catch `Exception`, including `_send_context` and `_resolve_node` which have no local handler and land correctly in the pre-dispatch block. No `ok` outcome changes and nothing escapes into core. Your claim holds.

Both of last round's findings are closed, verified by mutation: the `"litd" in n["sockets"]` predicate is now caught by two tests, and folding an API error back into "Unable to connect" is caught by two.

The split works in both directions too. Closed port gives "Unable to connect", `data: null` gives "Amboss API error: response contained no data", an HTML 502 gives "HTTP 502 with an unreadable body".

## The battery is not 18 of 18

Four mutations that leave all 22 tests green:

| Mutation | Result |
|---|---|
| wrap `httpx.RequestError` as `AmbossApiError` too | 22 passed |
| swap the `errors` and status-code checks | 22 passed |
| remove the `data is None` guard | 22 passed |
| predicate keyed on host (`"b" in ...`) rather than scheme | 22 passed |

The first is the one worth acting on. The whole point of this commit is the split between "the endpoint answered" and "the endpoint did not", and the transport half is pinned nowhere. Someone can fold it back tomorrow and CI stays green.

The second is the exact invariant your own comment exists for: `errors` before the status code, so a 400 carrying a useful message is not flattened to "HTTP 400". No test uses that combination, since every `errors` test runs on 200 and the only non-2xx test carries no `errors` array.

The fourth is the same shape for the third round running. The fixtures still differ on host, `http://a` against `https://b`. Making both `a` leaves the scheme as the only axis.

## One new robustness point

`body.get("errors")` assumes the body is an object. Valid JSON that is not one falls through to an `AttributeError`, which is not an `AmbossApiError`:

```
JSON array body   -> "Unable to connect to .../graphql."
JSON scalar body  -> "Unable to connect to .../graphql."
HTML 502          -> "Amboss API error: HTTP 502 with an unreadable body"
```

That is the exact misclassification this commit removes, for the one body type still slipping through. A proxy or WAF in front of the API returning a JSON array is the realistic source. HTML is fine because `r.json()` already fails on it.

## The pattern

Three rounds running the battery reports more coverage than exists, and the reason is the same each time: it only contains mutations you thought of. The transport half is the cleanest example. You wrote the split, tested the side that was new, and left the side that already worked unpinned. That is the mirror image of the lesson you drew two rounds ago about mutation-testing the tests you add but not the one whose behaviour you changed underneath.

None of this touches a funds path. Suite 316 passed and 20 skipped here, lint clean. Still zero check runs on the head commit with `mergeStateStatus` at BLOCKED.

arcbtc:
Cool. Ruff and linting errors, run `make`

lnbits/wallets/amboss.py
```diff
@@ -0,0 +1,702 @@
...
+            # billed separately). LND's fee_msat isn't always a multiple of
+            # 1000, so this can be a fractional-sat string — int() would raise.
+            fee_msat = round(abs(float(fee)) * 1000) if fee is not None else None
+            return PaymentSuccessStatus(fee_msat=fee_msat, preimage=tx.get("preimage"))
+        if status in ("FAILED", "EXPIRED"):
+            return PaymentFailedStatus()
+        return PaymentPendingStatus()
+
+
+def _self_check() -> None:
```
motorina0, line 647: this has no place in the wallet file

lnbits/settings.py
```diff
@@ -629,6 +629,17 @@ class BlinkFundingSource(LNbitsSettings):
     blink_send_without_probe: bool = Field(default=True)
 
 
+class AmbossFundingSource(LNbitsSettings):
+    amboss_api_endpoint: str | None = Field(default="https://app.amboss.tech/graphql")
+    amboss_service_api_key: str | None = Field(default=None)
+    amboss_wallet_id: str | None = Field(default=None)
+    amboss_team_password: str | None = Field(default=None)
+    # Sandbox wallets can be told to auto-settle received invoices via the
```
motorina0, line 637: shouldn't the endpoint/api_key/wallet_id determine on the funding source side if this is a sandbox?

motorina0:
@bufo24 please do not post AI generated comments. They are too verbose and nobody can read them.

I had a quick look over the code and added some commetns.

Is there a sandbox environment that we can use to test this funding source?

lnbits/templates/components/admin/funding_sources.vue
```diff
@@ -50,7 +50,7 @@
               v-model="formData[key]"
               filled
               class="q-mt-sm"
-              :type="hideInput ? 'password' : 'text'"
+              :type="prop.text ? 'text' : hideInput ? 'password' : 'text'"
```
motorina0, line 53: Please revert. We consider all funding source details secrets (even the endpoints)

lnbits/wallets/amboss.py
```diff
@@ -0,0 +1,702 @@
...
+    PaymentStatus,
+    PaymentSuccessStatus,
+    StatusResponse,
+    Wallet,
+)
+
+# --- crypto: ports of the SDK's argon2 + nip44 (must match byte-for-byte) ------
+
+
+def _argon2id(password: bytes, salt: bytes) -> bytes:
```
motorina0, line 49: All private methods should go at the bottom of the file/class

lnbits/wallets/amboss.py
```diff
@@ -0,0 +1,702 @@
...
+                "wallet_id": self.wallet_id,
+                "request": {"bolt11": bolt11},
+                # payment_hash doubles as a free, deterministic dedup key: a
+                # retried pay_invoice() call for the same invoice lands on the
+                # same rails transaction instead of creating a second send.
+                "idempotency_key": decoded.payment_hash,
+            }
+            if self.sandbox and self.sandbox_auto_complete:
+                # rails only acts on this for SANDBOX wallets; harmless otherwise.
+                send_input["metadata"] = json.dumps(
```
motorina0, line 393: why does it have to mix prod with sandbox logic?

Isn't there a dedicated sandbox environment on Amobss side?

PR #4113, opened 2026-07-29
feat: allow a webhook on outgoing payments

### LINKED ISSUE #4112: [Feature request] Allow a webhook on outgoing payments (`POST /api/v1/payments` with `out: true`)
## Summary

The webhook dispatch path already handles outgoing payments, but there is no way to attach a webhook URL to one, so that path is unreachable through the public API. A client can send `"out": true` together with `"webhook": "https://..."`, gets a `200`, and the field is silently dropped: no error, no webhook, `webhook_status` stays `null` forever.

The database column, the model field and the dispatch call all exist. Only the wiring from the request to the payment row is missing.

## Current behaviour on `dev`

`POST /api/v1/payments` takes a `CreateInvoice` body for both directions, and `CreateInvoice` has a `webhook` field.

1. On the outgoing branch, `api_payments_create` calls `pay_invoice()` without the webhook (`lnbits/core/views/payment_api.py:272-284`):

```python
if invoice_data.out is True and key_info.key_type == KeyType.admin:
    ...
    payment = await pay_invoice(
        wallet_id=wallet_id,
        payment_request=invoice_data.bolt11,
        extra=invoice_data.extra,
        labels=invoice_data.labels,
        external_id=invoice_data.external_id,
    )
```

2. `pay_invoice()` has no `webhook` parameter (`lnbits/core/services/payments.py:58`), and the `CreatePayment` it builds leaves `webhook` unset (`payments.py:93-102`), even though `CreatePayment.webhook` exists (`lnbits/core/models/payments.py:60`).

3. Dispatch itself is direction agnostic and would fire as soon as the field is populated. `_pay_external_invoice` (`payments.py:867`) and `_pay_internal_invoice` (`payments.py:789`, `:792`) both call `_send_payment_notification_in_background`, and `send_payment_notification` ends with (`lnbits/core/services/notifications.py:301-302`):

```python
if payment.webhook and not payment.webhook_status:
    await dispatch_webhook(payment)
```

So the only reason an outgoing payment never triggers a webhook is that `payment.webhook` is always `None`.

The storage side is already in place. An outgoing payment fetched from a running 1.5.4 instance shows both fields, permanently null:

```json
{
  "checking_id": "...",
  "amount": -5236000,
  "fee": -18339,
  "status": "success",
  "webhook": null,
  "webhook_status": null
}
```

## Steps to reproduce

1. `POST /api/v1/payments` with an admin key:

```json
{
  "out": true,
  "bolt11": "lnbc...",
  "webhook": "https://example.com/webhook"
}
```

2. The payment succeeds and returns `200`.
3. `GET /api/v1/payments/{payment_hash}` shows `"webhook": null` and `"webhook_status": null`.
4. No request ever arrives at the webhook URL.

## Expected behaviour

The `webhook` field is honoured for outgoing payments the same way it is for incoming invoices: it is stored on the payment row, and `dispatch_webhook` is called once the payment reaches `success`.

## Why this is useful

A backend that pays invoices (payouts, withdrawals, refunds) wants a push notification when its outgoing payment settles, instead of polling `GET /api/v1/payments/{payment_hash}`. Websockets can do this, but they need a long lived connection and a process to hold it, which does not fit a stateless HTTP backend that already exposes a webhook endpoint for incoming invoices.

Right now the only options are polling on a timer or a websocket listener, even though the payment row and the dispatch code are both ready for a webhook.

A PR follows that threads the value through from the request to the payment row.

## Notes

- Opt in, so nothing changes for existing callers: no `webhook` in the body means no dispatch.
- `check_callback_url` runs inside `dispatch_webhook`, so `LNBITS_CALLBACK_URL_RULES` keeps applying unchanged.
- This composes with the outgoing webhook signing in #4016, because signing happens inside `dispatch_webhook`.
- If honouring `webhook` on outgoing payments is not wanted, then returning a `400` for `out: true` combined with `webhook` would at least make the current behaviour explicit instead of silently discarding the field.

## Related

- #2472, paying an invoice does not send a webhook. The dispatch side is wired up now, but it cannot be reached because the field cannot be set.
- #1192 and #4016, signing for outgoing webhook bodies.

## Version

Observed on 1.5.4 and on current `dev`.

dsbaars:
Closes #4112.

`POST /api/v1/payments` accepts a `webhook` field, but on the outgoing branch (`out: true`) the value was silently discarded: the route did not forward it, `pay_invoice()` had no parameter for it, and the `CreatePayment` it builds left the field unset. Callers got a `200`, `webhook_status` stayed `null`, and no request was ever sent.

Everything else was already in place. `CreatePayment.webhook` exists, `apipayments` already stores `webhook` and `webhook_status` for outgoing rows, and the dispatch path is direction agnostic: `_pay_external_invoice` and `_pay_internal_invoice` both call `_send_payment_notification_in_background`, and `send_payment_notification` ends with `if payment.webhook and not payment.webhook_status: await dispatch_webhook(payment)`. The only reason an outgoing payment never fired a webhook is that `payment.webhook` was always `None`.

This threads the value through so a service that pays invoices (payouts, withdrawals, refunds) can get a push notification when its outgoing payment settles, instead of polling `GET /api/v1/payments/{payment_hash}` or holding a websocket open.

### Changes

- `webhook` parameter on `pay_invoice()`, passed into the `CreatePayment` it builds
- `api_payments_create` forwards `invoice_data.webhook` on the `out: true` branch
- One sentence in the endpoint description noting that `webhook` is honoured in both directions

### Notes for reviewers

- Opt in, so nothing changes for existing callers: no `webhook` in the body means no dispatch, and `webhook_status` stays `null`.
- `check_callback_url` runs inside `dispatch_webhook`, so `LNBITS_CALLBACK_URL_RULES` keeps applying unchanged. No new network egress path is introduced.
- No migration needed, the column is already there.
- This composes with the outgoing webhook signing in #4016, because signing happens inside `dispatch_webhook`.

### Testing

- `tests/unit/test_pay_invoice.py::test_pay_external_invoice_with_webhook` pays an external invoice with a webhook set, asserts the field is persisted on the outgoing payment row, and asserts `dispatch_webhook` is called once with that same outgoing payment (negative amount, matching payment hash).
- `tests/unit/test_pay_invoice.py::test_pay_external_invoice_without_webhook_does_not_dispatch` is the negative case: no `webhook` in the call means the row keeps `webhook` as `None` and `dispatch_webhook` is never called.
- `tests/api/test_api.py::test_pay_invoice_with_webhook` covers the route wiring end to end: it creates an invoice, pays it with `out: true` plus `webhook`, and asserts the returned payment carries the webhook.

### DIFF
--- a/lnbits/core/services/payments.py
+++ b/lnbits/core/services/payments.py
@@ -65,6 +65,7 @@ async def pay_invoice(
     tag: str = "",
     labels: list[str] | None = None,
     external_id: str | None = None,
+    webhook: str | None = None,
     conn: Connection | None = None,
 ) -> Payment:
     if settings.lnbits_only_allow_incoming_payments:
@@ -99,6 +100,7 @@ async def pay_invoice(
             extra=extra,
             labels=labels,
             external_id=external_id,
+            webhook=webhook,
         )
 
     async with db.reuse_conn(conn) if conn else db.connect() as new_conn:
--- a/lnbits/core/views/payment_api.py
+++ b/lnbits/core/views/payment_api.py
@@ -256,6 +256,8 @@ async def api_all_payments_paginated(
         `amount`, `unit`, and `memo`. To pay an arbitrary invoice from the funds
         already in the authorized account, specify `out: true` and use the `bolt11`
         field to supply the BOLT11 invoice to be paid.
+        The optional `webhook` field is honoured in both directions: LNbits POSTs
+        the payment to that URL once the payment reaches `success`.
     """,
     status_code=HTTPStatus.CREATED,
     responses={
@@ -281,6 +283,7 @@ async def api_payments_create(
             extra=invoice_data.extra,
             labels=invoice_data.labels,
             external_id=invoice_data.external_id,
+            webhook=invoice_data.webhook,
         )
         return payment
 
--- a/tests/api/test_api.py
+++ b/tests/api/test_api.py
@@ -307,6 +307,38 @@ async def test_pay_invoice(
     # assert payment.payment_hash == invoice["payment_hash"]
 
 
+# check POST /api/v1/payments: the webhook is kept on an outgoing payment
+@pytest.mark.anyio
+async def test_pay_invoice_with_webhook(
+    client,
+    inkey_headers_to,
+    adminkey_headers_from,
+    mocker: MockerFixture,
+):
+    mocker.patch(
+        "lnbits.core.services.notifications.dispatch_webhook",
+        AsyncMock(return_value=None),
+    )
+
+    invoice_data = await get_random_invoice_data()
+    create_response = await client.post(
+        "/api/v1/payments", json=invoice_data, headers=inkey_headers_to
+    )
+    assert create_response.status_code < 300
+
+    webhook_url = "http://test.404.lnbits.com"
+    data = {
+        "out": True,
+        "bolt11": create_response.json()["bolt11"],
+        "webhook": webhook_url,
+    }
+    response = await client.post(
+        "/api/v1/payments", json=data, headers=adminkey_headers_from
+    )
+    assert response.status_code < 300
+    assert response.json()["webhook"] == webhook_url
+
+
 # check GET /api/v1/payments/<hash>: payment status
 @pytest.mark.anyio
 async def test_check_payment_without_key(client, invoice: Payment):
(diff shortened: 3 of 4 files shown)


motorina0 (approved):
LGTM

@talvasconcelos can you please test

PR #4115, opened 2026-07-30
feat: add core notification support for extensions using nip17 nip17b

arcbtc:
Adds nip17 and nip17b support for extensions.

(in two minds whether its a good idea, as comms would ideally be bi-directional, so steps on nostrclient features)

Example use here https://github.com/lnbits/chat/pull/6

image


lnbits/core/services/nostr.py
```diff
@@ -47,6 +88,354 @@ async def send_nostr_dm(
...
+) -> dict:
+    keys = Keys.parse(from_private_key_hex)
+    receiver = PublicKey.parse(to_pubkey_hex)
+    rumor = EventBuilder.private_msg_rumor(receiver, message).build(keys.public_key())
+    gift = _gift_wrap_with_keys(keys, receiver, rumor)
+    await _send_event_to_relays(gift, relays)
+    return json.loads(gift.as_json())
+
+
+async def send_nostr_nip17b_dm(
```
motorina0, line 105: is this for nostr groups (`nip17b`)?

arcbtc: Yes, I'll change to 171

PR #4118, opened 2026-07-31
Delete cliche

motorina0:
Not used


arcbtc:
idk, is risky, someone could still be using in their stack.
Lets not do for 1.6, there is no rush

dni (approved):

PR #4122, opened 2026-08-05
fix(wallet): handle ?lightning= urls and restore LNURL-pay payer details

akovalenko:
Opening a wallet with a payment request in the url — `/wallet?lightning=<bolt11|lnurl|lightning address>` — never got as far as a usable payment dialog. This is the entry point an OS-level handler for `lightning:` links naturally targets, so it is worth having working. Two separate bugs were in the way; the first one is the blocker, the second is what the dialog looks like once it opens.

### 1. The wallet page crashed before it could decode the request

`created()` read the query parameter and called `decodeRequest()` before the active wallet was assigned. `decodeRequest()` dereferences `this.g.wallet`, which is still `null` at that point: `lnurlScan()` reads `g.wallet.adminkey` for LNURL and lightning addresses, and the bolt11 path reads `g.wallet.currency`. Either way a `TypeError` aborted `created()`, so the wallet was never assigned at all — the dialog could not decode anything, its "Read" button did nothing on every retry, and the wallet pane stayed empty.

Assigning the wallet first is necessary but not sufficient: the dialog is then opened while the `$router.replace()` that strips the query from the url is still in flight, and that navigation tears the dialog down again — so the page looks like it simply ignored the parameter. The request is now handled from the navigation's promise, once the url rewrite has settled.

### 2. The LNURL-pay dialog lost the payer details

The dialog reads `targetUser`, `domain`, `description`, `image` and `fixed` off `parse.lnurlpay`, but `/api/v1/lnurlscan` stopped providing them. It used to build an enriched dict from the LNURL metadata; since the lnurl-library refactor it returns the model verbatim (`tag`, `callback`, `min/maxSendable`, `metadata`, `commentAllowed`).

The visible result: the confirmation says it is requesting between x and y without naming anybody, the description and image are never rendered, and because `fixed` is `undefined` a fixed-amount link is shown as a range with an editable amount.

These fields are now derived on the client, the way the withdraw branch already derives its domain from the callback. They live in a computed property rather than being merged into `parse.lnurlpay`, because that object is posted back to `/api/v1/payments/lnurl` as-is and `LnurlPayResponse` is configured with `extra="forbid"` — mixing them in would fail validation on pay.

### Testing

Both fixes were developed against a live 1.5.5 instance and confirmed working there with a lightning address in the url; the bolt11 and LNURL branches reach the dialog through the same `created()` path. They were then cherry-picked onto `dev`, where they apply unchanged. Also checked: a wallet url with no query parameter opens no dialog, and an unknown wallet id still lands on the error page. The metadata parser was exercised against a real `payRequest` response plus fixed-amount, image and malformed-metadata inputs; malformed metadata degrades to showing the domain rather than breaking the dialog.

`make bundle` output is included as a separate commit, since `make checkbundle` is what runs for fork PRs.

---

Prepared with Claude Code; see the `Co-Authored-By` trailers on the commits.


talvasconcelos (approved):
tested

arcbtc (approved):

PR #4127, opened 2026-08-09
fix(clnrest): harden Core Lightning REST wallet handling

santyr:
This update improves LNbits’ Core Lightning REST wallet support. It makes invoice creation, outgoing payments, payment status checks, and invoice notifications more reliable with newer Core Lightning versions.
It also prevents old invoices from being processed again after LNbits restarts, handles payment errors more safely, and adds tests for the corrected behavior.

### DIFF
--- a/lnbits/wallets/clnrest.py
+++ b/lnbits/wallets/clnrest.py
@@ -5,6 +5,8 @@ import os
 import ssl
 import uuid
 from collections.abc import AsyncGenerator
+from hashlib import sha256
+from typing import Any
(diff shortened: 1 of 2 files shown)


motorina0:
Please add a simple human-written and human-readable description.

santyr:
Description updated.

motorina0:
Most of these issues have been handled in different PRs.
If you still find something that is not cover, please open smaller PRs that are easy to review.

They can be AI created, but must be human reviewable.

PR #4141, opened 2026-09-03
Better async pay tool for wasm

arcbtc:
Rather than pay the invoice, returns so async payment methods can be used to stop wasm runtime having to exceed default (currently the payment in a function may take longer than default 5 sec wasm runtime rule).


motorina0:
- @motorina0 need to add JMeter tests for one or two exts
- @arcbtc please add a short description

arcbtc:
> @arcbtc please add a short description"
done

PR #4143, opened 2026-09-03
feat: add comment column to payments CSV export

AxelHamburch:
## What / Why

The wallet CSV export used a fixed column list (`paymentsCSV.columns` in
`lnbits-payment-list.js`) that did not include the payment `extra.comment`
value. As a result, comments left by payers via LNURLp (and TPoS) are visible
in the payment detail dialog but cannot be exported from the UI at all — the
plain export omits them, and the "with details" export only expands
`extra.details`, not `extra.comment`.

## Change

Add a `Comment` column sourced from `row.extra.comment`, mirroring the existing
`Fiat Currency` / `Fiat Amount` columns. Applies to both the plain and the
detailed CSV export. `bundle-components.min.js` regenerated via `npm run bundle`.


motorina0:
@talvasconcelos can you please test this PR

arcbtc (approved):

PR #4144, opened 2026-09-04
fix: settings ui refector + blockexplorer tweaks

arcbtc:
HUGE and well needed refactor for settings pages, just ui
Block explorer eye-candy + clearer ui + fixed public page loading bug (few backend views needed)
image
image
image
image


talvasconcelos (approved):
Tested.

Small UI nitpicks, not blockers:

- on mobile, the collapsed "left tabs" become a dropdown select. it should maybe be more obvious, or have a primary bg color
- "restart server", changing color to signal the need is good, but it should've preserved the small red badge dot, to make it clear that the changed setting needs it

lnbits/static/i18n/en.js
talvasconcelos, line 1: these need to get into the other lang files

talvasconcelos:
added a small UI sugar, to make the tabs navigation more obvious on mobile:

image

PR #4150, opened 2026-09-06
Remove a user wallet on the second delete instead of returning 404

### LINKED ISSUE #4149: [BUG] Deleting a user wallet the second time returns 404 instead of removing it
**Describe the bug**
On the admin Users page, deleting a wallet the second time should delete it for good. The endpoint says so: the first call is a soft delete, the second call removes the row. But the second call returns 404 "Wallet does not exist." and the wallet stays in the database with `deleted = true`.

**To Reproduce**

1. Log in as admin and open the Users page.
2. Open the wallets of a user and delete one wallet. This works, the wallet is marked as deleted.
3. Delete the same wallet again. The dialog says "Wallet is already deleted, are you sure you want to permanently delete this user wallet?" Confirm.
4. The API answers 404 and shows an error. The wallet is still in the database.

Same with the API: `DELETE /users/api/v1/user/{user_id}/wallet/{wallet_id}` two times.

**Expected behavior**
The second delete removes the wallet, like the endpoint summary says.

- LNbits version: dev, commit d941f0a3 (1.6.0)
- Database: sqlite

**Additional context**
In `lnbits/core/views/user_api.py`, `api_users_delete_user_wallet` calls `get_wallet(wallet)`. The default of `get_wallet` is `deleted=False`, so a soft deleted wallet is not found and the `force_delete_wallet` branch below is never reached. The undelete endpoint next to it passes `deleted=True`.

I have a fix and will open a pull request.

monasco:
Deleting a user wallet the second time from the admin Users page returned 404 "Wallet does not exist." and the wallet stayed in the database as soft deleted. The lookup used the default `deleted=False`, so it never found the soft deleted wallet. Now the second delete removes the wallet, like the endpoint summary says.

The first delete (soft delete) does not change.

Fixes #4149

### DIFF
--- a/lnbits/core/views/user_api.py
+++ b/lnbits/core/views/user_api.py
@@ -363,7 +363,7 @@ async def api_users_delete_all_user_wallet(user_id: str) -> SimpleStatus:
 async def api_users_delete_user_wallet(
     user_id: str, wallet: str, account: Account = Depends(check_admin)
 ) -> SimpleStatus:
-    wal = await get_wallet(wallet)
+    wal = await get_wallet(wallet, deleted=None)
     if not wal:
         raise HTTPException(
             status_code=HTTPStatus.NOT_FOUND,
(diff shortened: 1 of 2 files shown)


arcbtc (approved):
Nice spot

motorina0 (approved):

PR #4156, opened 2026-09-07
fix: temp disable boltz

arcbtc:
Its a sad day seeing Boltz disabled, but hopefully in some form it will return.
Thank you Boltz for all the excellent things you did for bitcoin. 
Out of all the exploits in the space from these AI wars, it was a remarkable thing you never lost 1 sat of user funds 🫡


lnbits/wallets/boltz.py
```diff
@@ -24,6 +24,9 @@
     Wallet,
 )
 
+# Re-enable when the public Boltz swap service resumes.
```
motorina0: we would need to make a new release for this
Can we have a temp env variable that defaults to `True`?

lnbits/static/js/components/lnbits-home-logos.js
```diff
@@ -78,11 +78,7 @@ window.app.component('lnbits-home-logos', {
           lightSrc: '/static/images/phoenixdl.png',
           darkSrc: '/static/images/phoenixd.png'
         },
-        {
-          href: 'https://boltz.exchange/',
-          lightSrc: '/static/images/boltzl.svg',
-          darkSrc: '/static/images/boltz.svg'
-        },
+        // Boltz is temporarily hidden while its swap service is suspended.
```
motorina0, line 81: Users that already have bolts should still see this.

lnbits/settings.py
```diff
@@ -1241,6 +1240,12 @@ class SuperUserSettings(LNbitsSettings):
         ]
     )
 
+    @validator("lnbits_allowed_funding_sources")
+    @classmethod
+    def exclude_suspended_funding_sources(cls, sources: list[str]) -> list[str]:
```
motorina0: similar to `LNBITS_ALLOWED_FUNDING_SOURCES` we should have an `LNBITS_BLOCKED_FUNDING_SOURCES`

lnbits/wallets/boltz.py
arcbtc: good idea

lnbits/static/js/components/lnbits-home-logos.js
arcbtc: why, its ded?

motorina0: Maybe they want to clear connection info, or back it up if they haven't done so.

arcbtc:
redundent, as boltz has some forks in the wild people could use

PR #4162, opened 2026-09-08
Custom memo for Credit/Debit flows.

### LINKED ISSUE #4155: [Feature request] Allow custom memo for wallet balance updates
### Description

The /users/api/v1/balance endpoint currently allows an administrator to credit or debit a wallet by providing a wallet ID and an amount.

However, the resulting transaction always uses a generic memo:

Admin credit for positive balance updates
Admin debit for negative balance updates

It would be useful if the API allowed the caller to optionally provide a custom memo for the balance operation.

### Proposed API

For example:

{
  "id": "wallet_id",
  "amount": 5000,
  "memo": "Manual refund"
}

For a debit:

{
  "id": "wallet_id",
  "amount": -2500,
  "memo": "Balance correction"
}

The supplied memo should then be used as the transaction memo.

For superusers, the debit/credit modal is extended so a memo can also be set in UI 

Image

### Backwards compatibility

The memo parameter should be optional. If no memo is provided, the current behaviour should remain unchanged:

Positive amount → Admin credit
Negative amount → Admin debit

### Benefit

This would make balance adjustments much more useful for applications integrating with LNbits, as transactions could contain meaningful context instead of a generic Admin credit or Admin debit description.

Thanks!

DoktorShift:
Closes lnbits/lnbits#4155

## Custom memo for superuser credit/debit

Superusers could not set a custom memo when crediting or debiting a wallet —
the memo was hardcoded to "Admin credit" / "Admin debit", and users had no
way of knowing why their balance changed.

### API

`PUT /users/api/v1/balance` now accepts an optional `memo`
(max 640 chars, same cap as invoice memos):

```json
{ "id": "wallet_id", "amount": 1000, "memo": "Beer for the Horde" }
```

Omitting `memo` keeps a sensible default, so existing calls work unchanged.

### Recognition in the user's wallet

Admin credit/debit payments are now tagged (`extra.tag = "admin"`), so the
payment list shows the familiar yellow `#admin` badge — the same mechanism
extensions use. Since the badge carries the provenance, the default memos are
now simply "Credit" / "Debit" instead of repeating it; with a custom memo the
row reads `#admin Beer for the Horde`. The tag is searchable and included in
CSV exports and `extra.tag` for API consumers.

Bildschirmfoto 2026-09-08 um 12 29 06

### UI

The credit/debit popup (users area + wallet page) now has a memo field, plus
some small UX fixes:

- opens below its button with a soft transition instead of covering it
- visible action button that shows **Credit** and flips to a red **Debit**
  when the amount is negative — Enter still submits
- amount field starts empty instead of a prefilled `0` that had to be
  deleted before typing a negative value; empty/zero submits are blocked
- removed the decorative pencil icon, numeric input on both popup variants
- the wallets table in the users area refreshes after a credit/debit

Bildschirmfoto 2026-09-08 um 12 28 40

The admin notification for balance updates now includes the memo as well.

Tests cover default and custom memos on both credit and debit, the `admin`
tag, and memo length validation.


motorina0:
LGTM, but unit test fails.

motorina0 (approved):

PR #4168, opened 2026-09-12
fix(extensions): preserve zip on download failure, soft-fail restore

### LINKED ISSUE #4070: Startup zip check never matches absolute paths → mass deactivate on Docker recreate
## Problem
`check_installed_extension_files()` in `lnbits/app.py` does:

```python
zip_files = glob.glob(os.path.join(settings.lnbits_data_folder, "zips", "*.zip"))
if f\"./{ext.zip_path!s}\" not in zip_files:
    await ext.download_archive()
```

`ext.zip_path` is already absolute (e.g. `/data/zips/lnurlp.zip`). Prefixed with `./` it becomes `.//data/zips/lnurlp.zip`, which is **never** in the glob list (`/data/zips/lnurlp.zip`). So a present zip is treated as missing.

`download_archive()` then **deletes** the local zip before re-download. If the extensions dir is ephemeral (default `LNBITS_EXTENSIONS_PATH=lnbits` → `/app/lnbits/extensions`) and network/DNS is not ready yet (common after power loss / container recreate), restore fails and `check_installed_extensions()` calls `deactivate_extension()` for every installed extension.

Seen on Umbrel (only `/data` is volume-mounted). Related packaging fix: https://github.com/getumbrel/umbrel-apps/pull/5895 (`LNBITS_EXTENSIONS_PATH=/data`).

## Suggested fix
```python
if not ext.zip_path.is_file():
    await ext.download_archive()
```

Optionally also: do not `os.remove` the zip until the new download succeeds (download to `.tmp` then replace), and/or soft-fail restore instead of permanent deactivate.

Happy to open a PR once our PAT has full `repo`/`workflow` scopes (previous PR attempt from a stale fork was closed as #4069).

Bartok9:
Fixes #4070: download_archive() deleted the extension zip before re-downloading, so a failed fetch (DNS down on Docker recreate) destroyed the good zip and permanently deactivated all extensions. Now downloads to a temp file, verifies hash, atomically replaces via os.replace, and soft-fails with a warning instead of deactivating.

Closes #4070

Agent-Owner: sera · Claim: sera


motorina0:
CI Jobs fail.

lnbits/core/models/extensions.py
```diff
@@ -592,35 +592,40 @@ def is_wasm(self) -> bool:
...
+                and self.meta.installed_release.hash
+                and self.meta.installed_release.hash != archive_hash
+            ):
+                tmp_zip_file.unlink(missing_ok=True)
+                raise AssertionError("File hash mismatch. Will not install.")
+
+            os.replace(tmp_zip_file, ext_zip_file)
             self._remember_payment_info()
 
+        except AssertionError:
```
motorina0, line 622: what is the point of this?

lnbits/app.py
```diff
@@ -291,10 +291,13 @@ async def check_installed_extensions(app: FastAPI):
                     f"{ext.id} ({ext.installed_version})"
                 )
         except Exception as e:
-            logger.warning(e)
-            await deactivate_extension(ext.id)
```
motorina0, line 295: the extension (re)install can fail for multiple reasons, for example a corrupt migration file

motorina0:
Is this PR for a side effect of https://github.com/lnbits/lnbits/pull/4091 ?

milah-247:
## Review: fix(extensions): preserve zip on download failure, soft-fail restore

Verified locally against `d119f16` (current HEAD).

**Checks performed:**
- 3 new tests in `test_app_check_installed_extension_files.py` pass
- `ruff check` clean on all changed files
- `mypy` clean on changed files
- Full `tests/unit/` suite run with `LNBITS_BACKEND_WALLET_CLASS=FakeWallet` (matches `tests.yml`)

**Confirmed behavior:**
- Simulated `ConnectionError` during download leaves the original zip byte-for-byte intact, no `.tmp` residue
- Successful download atomically replaces the zip only after hash verification passes
- `check_installed_extensions` no longer calls `deactivate_extension()` on restore failure (soft-fail confirmed via mock assertion)

**Full suite failures are pre-existing and unrelated to this diff** : none touch `app.py`, `extensions.py`, or the new test file:
- 3 Breez wallet test failures (`AttributeError` on `bolt11_decode`/`PaymentType`) , pre-existing SDK mismatch on `dev`
- Remaining failures are event-loop/lock and DB-uniqueness errors from serial local test execution without per-test isolation , not present in CI's fresh-container setup

**Open questions:**
1. `except AssertionError: raise` / re-raising generic exceptions as `AssertionError("Cannot fetch extension archive file")` , does collapsing distinct failure types (network vs. hash mismatch vs. other) lose useful context for callers upstream?
2. With soft-fail, a broken extension can now stay in a failed-but-not-deactivated state indefinitely with only a log line. Is there any user-facing signal (admin UI, health check) for this, or is it log-only?
3. The "CI Jobs fail" comment predates the Sep 16 force-push to `d119f16`. Worth re-running CI on current HEAD : my local repro suggests it should be green now.

Code changes look correct and well-tested. Approving pending a fresh CI run on current HEAD.

motorina0:
@milah-247 no AI comments please

lnbits/core/models/extensions.py
Bartok9: Keeps hash-mismatch as AssertionError (same as before) so callers that already special-case it still work. Other failures get wrapped into the same AssertionError message the old path used, after cleaning the .tmp so we never leave a half-written zip next to a good one.

lnbits/app.py
Bartok9: Agreed — corrupt migrations and other install failures should still deactivate. The change is only for restore/download failure when the on-disk zip was never replaced (Docker recreate / DNS not ready case in #4070). Happy to narrow the soft-fail to download/network errors and keep deactivate for everything else if you prefer that.

motorina0:
moving this to draft
@Bartok9:
 - please see my comments
 - please do not add verbose comments in code

Bartok9:
@motorina0 Related but not the same path as #4091. #4070 is specifically `download_archive` deleting the existing zip before fetch, so a failed re-download on Docker recreate wipes every extension. This PR only does temp download + `os.replace`, and soft-fails restore when the old zip is still intact. Happy to tighten soft-fail as noted on the app.py thread.

Bartok9:
@motorina0 Addressed:
- removed the verbose code comments
- soft-fail only when archive fetch fails (`Cannot fetch extension archive file`) and the existing zip is still on disk; other restore failures still call `deactivate_extension`
- CI was green on lint/unit/api after the last push; pushed a follow-up commit with the above

Ready for another look when convenient.

Bartok9:
Feedback addressed in replies above: hash-mismatch still raises AssertionError (unchanged), and soft-fail is scoped to download/restore failures only — corrupt migrations still deactivate via the app.py path. CI all green. Ready for review.

Bartok9:
Rebased onto current `upstream/dev` (958e4da), no conflicts. Diff unchanged: 3 files, the soft-fail preserve fix. Previous review comments from @motorina0 above are addressed in the replies — the hash-mismatch stays an AssertionError for existing callers; other failures are wrapped into the same AssertionError the old path used, only after the on-disk zip is left intact.

PR #4173, opened 2026-09-15
[Feature Request]: add optional currency to Fiat wallets

talvasconcelos:
Adds an optional field to Fiat wallets to specify the currency on creation!

Although fiat wallets are still an RC LNbits release, the PR has some guards againts duplicate fiat wallets. Only 1 fiat wallet per user per currency!

PR #4176, opened 2026-09-16
[Feature Request] Allow setting currency on fiat wallet

talvasconcelos:
Allows sending an optional currency parameter to create wallet of type fiat.

Replaces #4173


lnbits/core/crud/wallets.py
```diff
@@ -17,9 +18,16 @@ async def create_wallet(
     user_id: str,
     wallet_name: str | None = None,
     wallet_type: WalletType = WalletType.LIGHTNING,
+    currency: str | None = None,
     shared_wallet_id: str | None = None,
     conn: Connection | None = None,
 ) -> Wallet:
+    if wallet_type == WalletType.FIAT and currency is not None:
```
motorina0: why not allow for all wallet types?

PR #4187, opened 2026-09-18
fix: posgres export

arcbtc:
hardening in a previous fix created a regression bug exporting the db from LNbits


motorina0 (approved):
LGTM

PR #4193, opened 2026-09-18
fix: only credit the service fee once a payment succeeds

### LINKED ISSUE #4192: Service fee is credited for payments that later fail, and never reversed
## Description

When an outgoing payment does not resolve within `lnbits_funding_source_pay_invoice_wait_seconds`, `_pay_external_invoice()` returns the still-`pending` payment instead of raising. `pay_invoice()` then credits the service fee unconditionally. If the payment later resolves as `FAILED`, nothing reverses that fee credit — the service fee wallet permanently keeps sats that were never earned.

**Version:** v1.6.1-rc2 (`bcbfa7939`), also present on current `main`. Funding source: LndRest, PostgreSQL.

## Code path

1. **`pay_invoice()`** — `lnbits/core/services/payments.py:103-110` — fee is credited with no status guard:

```python
payment = await _pay_invoice(
    wallet.source_wallet_id, create_payment_model, conn=new_conn
)

await _credit_service_fee_wallet(wallet, payment, conn=new_conn)
```

2. **`_pay_external_invoice()`** — `payments.py:843-850` — returns a pending payment on timeout, no exception raised:

```python
except asyncio.TimeoutError:
    # return pending payment on timeout
    logger.debug(
        f"payment timeout after {wait_time}s, {checking_id} is still pending"
    )
    return payment
```

3. **`_credit_service_fee_wallet()`** — `payments.py:994-1017` — writes the record hardcoded as successful:

```python
await create_payment(
    checking_id=f"service_fee_{payment.payment_hash}",
    data=create_payment_model,
    status=PaymentState.SUCCESS,
    conn=conn,
)
```

4. **`update_pending_payment()`** — `payments.py:385-391` — marks the payment failed, but does not touch the fee record:

```python
status = await check_payment_status(payment)
if status.failed:
    payment.status = PaymentState.FAILED
    payment = await update_payment(payment, conn=conn)
```

`grep -rn "service_fee_" --include=*.py lnbits/` returns only the two creation sites (`core/services/payments.py`, `core/services/fiat_providers.py`). There is no code anywhere that deletes or reverses a `service_fee_*` record.

## When it triggers

| outcome | fee | correct |
|---|---|---|
| fails within the wait window | `PaymentError` raised, fee never credited | ✅ |
| succeeds within the wait window | credited | ✅ |
| times out → pending → later succeeds | credited early, ends up correct | ✅ |
| times out → pending → later fails | **credited, never reversed** | ❌ |

Only the last row is the bug. On a healthy node most payments resolve inside the 5s window, which is likely why this has gone unnoticed.

## Steps to reproduce

1. Set `lnbits_service_fee > 0` and configure `lnbits_service_fee_wallet`.
2. Pay an external invoice that neither settles nor fails within `lnbits_funding_source_pay_invoice_wait_seconds` (default 5s) — e.g. a destination with no viable route, so LND keeps retrying until `FAILURE_REASON_TIMEOUT`.
3. The payment goes `pending`, and a `service_fee_<hash>` record is immediately created in the fee wallet with `status = success`.
4. Wait for `check_pending_payments()` to pick up the funding source's `FAILED`.
5. The payment becomes `failed`; the `service_fee_<hash>` record is still `success`.

## Impact

Internal accounting drift: the sum of wallet balances exceeds the funding source balance by the total of the orphaned fees.

On our instance, 33 failed payments over ~7 months left **31,878 sats** credited to the service fee wallet — roughly 14% of that wallet's balance.

```sql
SELECT p.status, count(*), sum(f.amount)/1000 AS fee_sats
FROM apipayments f
JOIN apipayments p ON p.checking_id = replace(f.checking_id, 'service_fee_', '')
WHERE f.checking_id LIKE 'service\_fee\_%'
GROUP BY p.status;
```

```
 status  | count | fee_sats
---------+-------+----------
 failed  |    33 |    31878
 pending |     2 |     8372
 success |  2913 |  1525565
```

Concrete case: an outgoing payment of 940,818 sats (hash `f79d0a87…`) timed out at the funding source, stayed pending for ~35h, and was then marked failed by the node with `LNDRest Payment failed: FAILURE_REASON_TIMEOUT`. Its 9,408 sat service fee record remains `success`.

## Suggested fix

Either:

- reverse or delete the `service_fee_<payment_hash>` record in `update_pending_payment()` when the payment transitions to `FAILED`; or
- defer `_credit_service_fee_wallet()` until the payment actually reaches `SUCCESS`. `update_payment_success_status()` looks like the natural place — it already recomputes `service_fee()` for the payment's own `fee` field.

The second option also removes the window where the fee wallet shows income for a payment that has not settled yet.

Happy to open a PR if you have a preference between the two.

gorrdy:
Fixes #4192.

`pay_invoice` credited the service fee wallet unconditionally, right after `_pay_invoice` returned:

```python
payment = await _pay_invoice(
    wallet.source_wallet_id, create_payment_model, conn=new_conn
)

await _credit_service_fee_wallet(wallet, payment, conn=new_conn)
```

When an outgoing payment does not resolve within `lnbits_funding_source_pay_invoice_wait_seconds`, `_pay_external_invoice` returns the still-`pending` payment instead of raising, so the fee was booked for a payment that had not settled. If that payment later failed, nothing reversed the credit — there is no code anywhere that deletes or reverses a `service_fee_*` record — and the fee wallet kept sats that were never earned.

## Change

Credit the fee where a payment actually reaches `SUCCESS`:

- **`update_payment_success_status`** for external payments. This is the single point all three external success paths go through: the fast settle in `_pay_external_invoice`, the pending re-check in `update_pending_payment`, and `_verify_external_payment`.
- **`_pay_internal_invoice`** for internal payments, which are created as `SUCCESS` directly and never pass through `update_payment_success_status`.

Two things that the obvious one-line version gets wrong, and that this PR guards against:

- `update_payment_success_status` is also reached by **incoming** payments via `update_pending_payment`, so it carries an `payment.is_out` check. Without it, incoming payments would start being charged a service fee, which has never been the behaviour.
- A payment can be confirmed successful more than once (the pending check racing with an explicit status lookup), so `_credit_service_fee_wallet` now skips when the `service_fee_<hash>` record already exists, rather than letting the insert hit the primary key.

## Behaviour

| payment outcome | before | after |
|---|---|---|
| fails within the wait window | no fee (`PaymentError` raised) | unchanged |
| succeeds within the wait window | fee credited | unchanged |
| times out → pending → succeeds | fee credited while pending | fee credited on settle |
| times out → pending → **fails** | **fee credited, never reversed** | **no fee** |
| internal payment | fee credited | unchanged |
| incoming payment | no fee | unchanged |

The third row also closes the window where the fee wallet showed income for a payment that had not settled yet.

## Testing

`tests/unit` — 643 passed, 3 failed; `tests/api` — 272 passed.

The 3 failures are in `tests/unit/test_wallet_payment_ambiguity.py` (Breez) and reproduce identically on unmodified `main` at `9349da1`, so they are unrelated to this change.

`test_service_fee` in `tests/unit/test_pay_invoice.py` covers the fast-settle path and still passes. Happy to add a regression test for the timeout → pending → failed path if you'd like one — it needs a `pay_invoice` mock that neither settles nor fails inside the wait window, followed by a `update_pending_payment` with a failed status.

`mypy`, `ruff` and `black` are clean on the changed file.

## Not covered

`lnbits/core/services/fiat_providers.py` creates `service_fee_*` records through its own path. I have left it alone since I cannot test the fiat providers, but it is worth a look — the same question applies there.


motorina0:
- added tests
- small refactorings
- rebased from dev
- check bolt12 `pay_offer()` service fee

motorina0 (approved):

PR #4194, opened 2026-09-20
feat: onchain wallet type

arcbtc:
image
image


blackcoffeexbt:
Performed an end-to-end test of the on-chain wallet. Everything worked really well overall. Love it.

The UI is excellent, and features such as coin selection and custom fee selection are very nicely done.

### Issues

**Replace by fee**

RBF was off on my test tx. Should this be on by default? Can a setting be added to server / user wallet / tx settings?

**Typo**

The message:

> "Amount to small (below dust limit)"

should be:

> "Amount too small (below dust limit)"

**Fee slider maximum**

I couldn't increase the fee rate above the recommended maximum using the slider. In my test this was capped at around **2 sat/vB**.

I think users should still be able to select a higher fee rate if they want to. For example, the LNbits calculated "high priority" fee is lower than current fee recommendations from mempool.space which has a very good fee calc algo.

### Suggestions

**Rename "Server wallet" to "Hot wallet"**

Could "Server wallet" be renamed to **"Hot wallet"**? This makes the security implications clearer.

It would also be useful to display a warning explaining that the keys to this wallet are held on the server and could be vulnerable. In a way that doesn't put off a user.

**Disable Hot wallet option when unavailable**

If the server wallet / hot wallet feature has not been enabled by the admin, could the option be greyed out rather than simply unavailable? An info/warning icon could explain something along the lines of:

> "Hot wallets must be enabled by the server administrator."

**Admin: Onchain encryption key requirement**

In the server admin UI, it isn't immediately clear that an encryption key must be generated before on-chain wallets can be enabled.

A tooltip, warning, or short piece of explanatory text here would make the setup process clearer.

**Seed phrase warning**

On the wallet seed phrase screen, I suggest moving:

> **Anyone with this phrase can spend your bitcoin. Keep your backup somewhere private.**

above the seed words.

It should also be made considerably more prominent — for example, bold text, a larger font, or a warning-style component — so that users see it before reading or copying the seed phrase.

PR #4021, opened 2026-06-23
feat: Extensions for all

motorina0:
## Summary

  Adds a WASM extension framework for LNbits, allowing third-party extensions to run in a constrained runtime with explicit install-time permissions, sandboxed UI rendering, scoped storage, and host-mediated access to
  LNbits functionality.

## Backend request flow
```
• Client request
       |
       v
  /api/v1/ext/{ext_id}/{route}
       |
       v
  WASM API route handler
       |
       | 1. match route from extension config
       | 2. check auth if route is private
       | 3. read path/query/body payload
       v
  invoke_wasm_extension_export(ext_id, export_name, payload, user)
       |
       v
  Load WASM extension metadata
       |
       v
  Load cached WASM component
       |
       v
  Create ExtensionHostAPI(ext_id, permissions, user_id, access_token)
       |
       v
  Wire host functions into WASM linker
       |
       v
  Call WASM export function
       |
       v
  +-----------------------------+
  |       WASM extension        |
  |                             |
  | export handles request      |
  | and may call host imports   |
  +-------------+---------------+
                |
                | host function call
                v
  +-----------------------------+
  |      ExtensionHostAPI       |
  |                             |
  | checks permission/auth      |
  | scopes storage by owner     |
  | calls LNbits core services  |
  +-------------+---------------+
                |
                v
  LNbits core
  (wallets / payments / storage / utils / extension API / HTTP client)
                |
                v
  ExtensionHostAPI response
                |
                v
  WASM export returns JSON
                |
                v
  HTTP response to client
```

## Frontend
```
• +-------------------------------------------------------------+
  |                         LNbits UI                           |
  |                                                             |
  |  Main layout                                                |
  |  - left menu                                                |
  |  - top bar                                                  |
  |  - authenticated user context                               |
  |                                                             |
  |        route: /ext/{ext_id}/...                             |
  |                           |                                 |
  |                           v                                 |
  |              +--------------------------+                   |
  |              | WasmExtensionComponent   |                   |
  |              |                          |                   |
  |              | - asks backend for frame |                   |
  |              | - owns bridge            |                   |
  |              | - checks allowed routes  |                   |
  |              | - handles QR / notify    |                   |
  |              +-------------+------------+                   |
  |                            |                                |
  |                            | creates sandboxed iframe       |
  |                            v                                |
  |              +--------------------------+                   |
  |              | sandboxed iframe         |                   |
  |              | /ext-frame/{ext_id}/{n}  |                   |
  |              |                          |                   |
[...]

PR #4195, opened 2026-09-21
fix: paid install QR code never shown for WASM extensions with permissions

bitkarrot:
## Bug

For a WASM extension that both **declares permissions** and has a **`pay_link`** (paid install), clicking **Show QR** in the releases dialog never displays the invoice QR code.

### Root cause

`showInstallQRCode` gates on `resolveExtensionPermissionGrant` (added in #4021). When the user clicks "Grant & Install", `resolveExtensionPermissionDialog` sets `showManageExtensionDialog = false`, closing the manage dialog entirely. The method then continues, fetches the invoice (`PUT /{ext_id}/invoice/install` succeeds, `payment_request` is set on `selectedRelease`), but the QR card is rendered *inside* that same dialog — so it never becomes visible.

`payAndInstall` got matching handling in #4021 (it closes the dialog deliberately for inline wallet payment); the QR path needed the opposite — the dialog reopened after the grant resolves.

### Reproduce

1. Add a manifest source listing a WASM extension with `permissions` in `config.json` and a `pay_link` on the release entry
2. Extensions → extension → Manage → release → **Show QR**
3. Grant permissions in the dialog → invoice is fetched (200 OK) but the dialog closes; QR never shown

All currently paid extensions are non-WASM and skip the permission gate, which is why this has not been hit before.

## Fix

One line: reopen the manage dialog after the permission grant resolves, before fetching the invoice. `bundle-components.min.js` rebuilt via `npm run bundle`.

## Test plan

- [x] WASM extension with `pay_link`: Show QR → grant → QR + copy-invoice render
- [ ] Free WASM extension install flow unchanged (permission grant still works)
- [ ] Non-WASM paid extension: Show QR works as before (no permission gate)


motorina0 (approved):
LGTM

Issue #4006, opened 2026-06-13
[BUG] NWC as funding source is spamming wallets


dadofsambonzuki:
**Describe the bug**

Connect funding source to any NWC provider like Rizful and Alby.

**To Reproduce**
Add NWC funding source

**Expected behavior**
Polling mechanism to be more conservative.

Potentially related: https://github.com/lnbits/lnbits/issues/3973

Related PR: https://github.com/lnbits/lnbits/pull/4008

dadofsambonzuki:
https://njump.to/nevent1qqsdvacjcyacshnpjyhtfex4mr6sarhd0qvauklam5f2vtntrkt0ufgpz9mhxue69uhkummnw3ezuamfdejj7q3qxv8mzscll8vvy5rsdw7dcqtd2j268a6yupr6gzqh86f2ulhy9kkqxpqqqqqqzcg3qdm

MegalithicBTC:
Here is a sketch of one way that rate limiting could work for NWC actions -- example is in Ruby but it could give you a general idea for any language: https://github.com/MegalithicBTC/nwc-ruby#polling-as-a-backup

MegalithicBTC:
Another sketch, in node.js: https://docs.megalithic.me/rizful-tips-and-tricks/accept-bitcoin-payments-on-your-website-or-app-with-rizful#poll-until-paid

MegalithicBTC:
Please be aware -- external NWC providers are currently not safe to use in LNBits. You will hit rate limits and any well-run NWC service will quickly blacklist your connection.

dadofsambonzuki:
OK, I'm now running https://github.com/lnbits/lnbits/pull/4008 patched to my instance.

@MegalithicBTC - pls advise if you are still seeing excessive polling issues.

cc @blackcoffeexbt

MegalithicBTC:
@dadofsambonzuki thanks, ideally we would find an LNBits user to test it directly and we could watch it in our logs..  if you find yourself reading this issue and you are an LNBits user who would like to test this, please contact us directly: https://megalithic.me/contact

dadofsambonzuki:
> @dadofsambonzuki thanks, ideally we would find an LNBits user to test it directly and we could watch it in our logs..  if you find yourself reading this issue and you are an LNBits user who would like to test this, please contact us directly: https://megalithic.me/contact

Yhm

dadofsambonzuki:
[...]

PR #4197, opened 2026-09-21
fix: fail expired incoming invoices in a bulk pass, independent of the funding source

dadofsambonzuki:
## Problem

`check_pending_payments()` is documented as also deleting expired invoices, but that cannot happen today:

1. `get_payments(...)` uses the default page size (10), so one run inspects at most 10 pending rows — a backlog of expired invoices takes many hours to clear, and until then every run keeps offering those invoices to the funding source.
2. The function returns early for `VoidWallet`, so with a placeholder backend nothing is cleared at all.

In production (self-hosted instance, NWC funding source) a single client created **309 unpaid 1-sat invoices** over two days. Because nothing drained them, the pending-payment machinery kept asking the provider about them: demand reached ~450 requests/hour, the provider began returning `RATE_LIMITED`, and then stopped answering for hours. The demand-side analysis is in #4006.

## Change

Fail expired **incoming** invoices in one unbounded, DB-only pass, before the funding source is consulted:

- an invoice past its expiry can never be paid, so no backend call is needed (and none is made);
- the pass runs before the `VoidWallet` early return, so it also works with a placeholder backend;
- pending **outgoing** payments are untouched — those still have to be resolved against the backend.

## Testing

`tests/unit/test_services_payments.py`, two new tests:

- `test_fail_expired_incoming_invoices_drains_backlog_without_wallet_calls` — expired incoming rows become `failed` + `expired`, live and outgoing rows are left alone, and `check_payment_status` is never awaited;
- `test_check_pending_payments_drains_expired_invoices_on_voidwallet` — expired rows drain even though the sweep returns early.

Both pass, in isolation and as part of the full file.

Caveat: in my local environment three unrelated tests in that same file also fail (`test_lnurl_withdraw_success_keeps_invoice_pending`, `test_webhook_is_validated_before_invoice_creation`, `test_update_pending_payment_and_bulk_pending_updates`). They fail identically on unmodified `v1.6.2`, so they are unrelated to this change.

Refs #4006


motorina0:
I have fixed `check_pending_payments()` to do batches of 100 (max 10 batches, max total 1000 pending checks).
The job runs every 30 minutes.

motorina0 (approved):

